PR #3, opened 2016-02-10
Convert docs to gitbook

doug-wade:
This isn't ready to merge yet -- I wanna do at least one publish first, but I wanted to discuss gitbooks as a distribution method for our docs first.  This was recommended by [the gulp guys](https://github.com/gulpjs/gulp/pull/1465); seems like a reasonable approach to me.


roblg:
I'm not sure I fully understand the implications of going with this format. The [output looks pretty though](http://shasta.tools/palisade/). Would we host these on github pages?

gigabo:
This seems cool.

Looks like we could use [our own domain](https://help.gitbook.com/platform/domains.html), but it will still be under `domain.com/react-server-docs`?

Might get more flexibility with a simple static site generator for a github pages site.

gigabo:
Let's revisit this when we've got the _content_ of our docs under control.

I'd actually like to see us set up a `react-server` instance for a docs site. 🚀

PR #5, opened 2016-02-11
Attempt at travis integration

gigabo:
Also pulled eslint config up to the top-level and added some README badges.


gigabo:
Success!

gigabo:
Note that this actually runs the _tests_ twice. Filed https://github.com/redfin/react-server/issues/6 for that.

Only runs `eslint` once.

.travis.yml
```diff
@@ -0,0 +1,11 @@
+---
+language: node_js
+
+node_js:
+ - "stable"
```
roblg: Super-weird that we can use `4.3.2`, especially since that's the LTS version? 

Weird. https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Provided-Node.js-Versions

roblg:
LGTM

PR #9, opened 2016-02-11
Expose FramebackController and emit events

gigabo:
- Request instance available via `FramebackController.getCurrent()`
- Events emitted:
    - `showFrame`
    - `hideFrame`


roblg:
LGTM

PR #10, opened 2016-02-11
Add react-server-gulp-route-loader

roblg:
`triton-route-loader` --> `react-server-gulp-route-loader`

Comments on name are obviously welcome...

We could also probably scrap this PR and start from scratch w/o losing much...


gigabo:
> All committers of the pull request should sign our Contributor License Agreement

Yikes.  These folks were all Redfin employees at the time the commits were made... @davidalber can we ignore the CLA for this one?

roblg:
My plan was totally to do that. :)

On Thu, Feb 11, 2016 at 3:52 PM, Bo Borgerson <notifications@github.com>
wrote:

> All committers of the pull request should sign our Contributor License
> Agreement
>
> Yikes. These folks were all Redfin employees at the time the commits were
> made... @davidalber <https://github.com/davidalber> can we ignore the CLA
> for this one?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/redfin/react-server/pull/10#issuecomment-183115457>.
>



-- 

*ROBERT GAY | REDFIN | Technical Leadrobert.gay@redfin.com
<robert.gay@redfin.com> | Tel:206-576-5824*

davidalber:
@gigabo: Definitely.

gigabo:
LGTM.  Name seems as good as any.

packages/react-server-gulp-route-loader/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "react-server-gulp-route-loader",
+  "version": "0.0.1",
+  "description": "A gulp plugin to handle lazy-loading of react-server Page objects to support code-splitting with webpack",
+  "main": "index.js",
```
gigabo: Wait... it looks like it's called `main.js` above.  Does this work?

FWIW - I think `index.js` is the more common convention.

aickin:
Dunno if it's interesting or not, but I ended up taking a slightly different approach in the CLI project. Instead of putting a token like __LAZY_REQUIRE__ and replacing it at build time, I just had the page entry be a string, like:

```javascript
routes: {
		Simple: {
			path: ['/simple'],
			method: 'get',
			page: "./SimplePage"
		},
}
```

and then during the build step output a server and a client version of the routes. The former just directly calls `require`, while the latter calls `require.ensure`. I figured it'd probably be easier to document that page is the path to the page module, rather than `__LAZY_REQUIRE__(pathToPageModule)`. Thoughts?

packages/react-server-gulp-route-loader/package.json
roblg: Prooobably not. I knew there was something I hadn't tested. I'll fix it.

roblg:
> during the build step output a server and a client version of the routes.

That definitely does not sound worse than existing weirdness.

Are you doing a rewrite at the place the routes file is required? Or do you output the two different routes files, and then replace the existing routes file with something like:

```javascript
module.exports = if (SERVER_SIDE ) ? require("./routes-server") : require("routes-client");
```

aickin:
I output two different route files, one called `routes_server.js` and one called `routes_client.js`. Then when starting up the server I have a line that looks something like:

```javascript
triton.middleware(server, require("./routes_server"));
```

(yes, it's still triton... ;)

Then in the webpack build for the client I include `routes_client.js` in the entrypoints portion; something like:

```javascript
entry: {Simple: ["./routes_client", "./pages/SimplePage"]}
```

Does that make sense? If you want the full ugliness, check out [https://github.com/aickin/react-server-cli/blob/master/src/compileClient.js].

roblg:
@aickin I'm finally getting around to thinking about this PR again, and I think there are a couple of things I hadn't considered before.

The big thing is that the routes file format in the `react-server-cli` examples is _almost_ JSON -- there are no `require`s. This is pretty different than what we're doing internally at the moment, for better or worse. :grin: I think it'd be doable, but a PITA to read our routes files, and modify only the `routes` property of the exported object.

So I'm not sure where to go from here. I'm not super-excited about the `__LAZY_REQUIRE__` thing, but the use-a-string-for-page-property thing is going to take more thought on our end.

I *think* I'm inclined to close/abandon this PR in favor of adding a route loading plugin more similar to your approach, and we'll just publish our `__LAZY_REQUIRE__` route loading plugin internally and go on with life for now, and we can revisit internal adoption of the string-for-page-property approach at a later date. (Or maybe we don't need a route loading plugin if `react-server-cli` is going to be the primary build tool for new arrivals?)

Thoughts?

(cc: @gigabo @doug-wade)

gigabo:
@roblg - I agree with your last comment here.  Any tickets to shake out of that before closing this?

roblg:
Erm... I think not, unless we want to extract @aickin's route loading thing from `react-server-cli` into a separate plugin, which seems probably unnecessary?

PR #11, opened 2016-02-12
Pull logger-loader out into `react-server-gulp-module-tagger`

gigabo:
TODO: Pass in the token to be replaced (e.g. `.pipe(tagger('__LOGGER__'))`)


packages/react-server-gulp-module-tagger/index.js
```diff
@@ -6,10 +6,10 @@ var replace = require("gulp-replace")
 //   - "__LOGGER__({ /* options */ })"
 var isWindows = ('win32' === process.platform)
 ,   REPLACE_TOKEN = /(?:__LOGGER__|__CHANNEL__)(?:\(\s*(\{[\s\S]*?\})\s*\))?/g
-,   THIS_MODULE   = isWindows
-						? /(?:[^\\]+\\node_modules\\)?react-server\\buildutils\\logger-loader\.js$/
-						: /(?:[^\/]+\/node_modules\/)?react-server\/buildutils\/logger-loader\.js$/
 ,   BASE_PATH     = module.filename.replace(THIS_MODULE,'')
+,   THIS_MODULE   = isWindows
+	? /(?:[^\\]+\\node_modules\\)?react-server-gulp-module-tagger\.js$/
+	: /(?:[^\/]+\/node_modules\/)?react-server-gulp-module-tagger\.js$/
```
gigabo: Oh, this is wrong.

gigabo:
Okay, at long last.  Working.

doug-wade:
:+1:

.travis.yml
```diff
@@ -14,7 +14,7 @@ addons:
       - g++-4.8
 
 before_script:
-  - npm install -g gulp
-  - for d in `ls packages/`; do cd packages/$d; npm install; cd -; done
+  - npm install -g gulp lerna
+  - lerna bootstrap
```
roblg: YES. I actually meant to leave this comment on one of the PRs from yesterday. `bootstrap` seems like the way it should be.

roblg:
The names of these build plugins are outta control, but I think the diff looks reasonable. :+1:

.travis.yml
gigabo: Yeah, so much cleaner!

Incidentally I couldn't get lerna to link the two packages together without requiring `react-server-gulp-module-tagger` with a caret.  It [_looks_](https://github.com/kittens/lerna/blob/master/lib/commands/bootstrap/linkDependenciesForPackage.js#L48) like an exact version match should work.  I haven't traced it through to figure out why it didn't.

Issue #13, opened 2016-02-12
triton-root elements should not add extra dom nesting/nodes


kcjonson:
Currently react server adds extra dom nodes for each root element that it creates.  This makes full page or "app like" layouts quite tricky, especially since they don't have unique classes on them.  Preferably we would not add an extra node if its not needed.

If this proves impossible, a fallback would be to allow developers to add unique classes to the root elements so that they can be styled.

Example below:
```
<body class='route-'>
   <div id='content' data-triton-content>
      <div data-triton-root-id="0" data-triton-timing-offset="25">
         <div class="Header" data-reactid=".wcmxgbtvk0" data-react-checksum="2028020681">...</div>
      </div>
   </div>
</body>
```

gigabo:
Thanks @kcjonson.  It _would_ be nice to get rid of the extra level of nesting.  I'll see if I can make that work.

gigabo:
So... I've tried:
- Cloning each root element with a `data-triton-root-id` prop.  Doesn't work, prop not propagated down to actual dom element.
- Walking down through children to find first actual dom element to clone with a `data-triton-root-id` prop.  Doesn't work, some elements don't have the dom node in their children until render time, which is too late.
- Doing string-replacement to add a `data-triton-root-id` attribute to the HTML on the way out.  Doesn't work, React freaks out and blows away the DOM branch.
- Doing string-replacement, and then calling `removeAtribute(...)` before rendering into the root.  Doesn't work, React _still_ somehow freaks out and blows away the DOM branch.

Not sure we're going to be able to _eliminate_ the wrapping div.

What about being able to set properties on it?  Like `className`, `style`, `id`, etc?  Would that get you what you need for layout purposes?

egid:
From my perspective, yes - allowing us to add attributes to root elements solves most of the issue. It still winds up being messier than needed, and I'm sure has some cases where it's not a full solution, but it solves the immediate problem.

gigabo:
Okay, cool.  That should actually be pretty easy and fairly clean.

I'll just add support for those props to `<RootElement />`.

kcjonson:
Yea, that solves the technical problem.  Still feels a bit dirty to have all the extra nodes lying around, but if its going to open a can of worms we can move on.

gigabo:
Available in `v0.0.9`.

PR #16, opened 2016-02-13
Support html attributes on RootElement

gigabo:
Couldn't get the divs out of the document.

At least now we can set attributes on them.

https://github.com/redfin/react-server/issues/13


roblg:
LGTM :+1:

PR #2, opened 2016-02-10
Add Yeoman generator and example; fix bugs from npm link, npm start

doug-wade:
This adds a yeoman generator to help new devs quickly scaffold a react-server instance.  Yeoman generators have to start with `generator-`, and `generator-react-server` is already taken -- we could do generator-reactserver or generator-rs, but I thought generator-corvair might be more memorable.

This also resolves a couple of issues I encountered along the way.

    dougwade packages/example-corvair ‹add-yeoman-generator*› » npm start

    > example-corvair@0.0.1 start /Users/doug.wade/react-server/packages/example-corvair
    > node server.js

    Start server...
    Listening on port 3000
    Started!
    (node) OutgoingMessage.flush is deprecated. Use flushHeaders instead.
    Warning: React.renderToString is deprecated. Please use ReactDOMServer.renderToString from require('react-dom/server') instead.
    2016-02-10T08:43:07.681Z - error: [react-server.core.renderMiddleware] Error rendering element 0 message=TRITON_CONFIGS environment variable required to start server., stack=Error: TRITON_CONFIGS environment variable required to start server.
        at module.exports (/Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/target/server/config.js:17:11)
        at getNonInternalConfigs (/Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/target/server/renderMiddleware.js:904:19)
        at bootstrapClient (/Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/target/server/renderMiddleware.js:809:13)
        at writeElements (/Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/target/server/renderMiddleware.js:773:4)
        at doElement (/Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/target/server/renderMiddleware.js:601:3)
        at /Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/target/server/renderMiddleware.js:609:11
        at /Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/node_modules/continuation-local-storage/context.js:74:17
        at _fulfilled (/Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/node_modules/q/q.js:834:54)
        at self.promiseDispatch.done (/Users/doug.wade/react-server/packages/example-corvair/node_modules/react-server/node_modules/q/q.js:863:30)
[...]

PR #17, opened 2016-02-14
Add package react-server-cli, a startup script that compiles client code and starts up a server.

aickin:
Here's a preliminary PR that adds a separate package that contains a startup script for compiling client code and running a server. The README.md describes its goals and what it currently does, so I won't repeat that here.

If you want to see what a minimal project using `react-server-cli` looks like, check out [the hello world example](https://github.com/aickin/react-server-examples/tree/master/helloworld).

A few things to note:

* I decided to simplify the `routes` format a little bit, getting rid of `__LAZY_REQUIRE__` and making it more like a JSON format (although technically it's still JavaScript).
* I think there's a good debate to be had about whether starting up a server is something that a library should do, or if every project should have its own `server.js` that instantiates Express etc. To start, I think it's probably good to run this way to ease developer pain and solve 95% of cases, but I totally understand arguments against.
* It could be cleaner to have compiling client code and running a server as two separate functions that are in different packages. They do have some interdependencies, though, so it's not 100% clear to me that they can be easily split up.

There's potentially a lot of bikeshedding to do here, and I welcome it. A few topics off the top of my head:

* Is `react-server-cli` even a good name? Maybe it implies (incorrectly) that it is intended to be a globally installed tool.
* Should routes just be a JSON file rather than a JavaScript file? Constraining what it can do makes it much easier to compile into code splitting on the client.
* Is it right to have a bin directory with an executable file?
* Currently `react-server` is a peer dependency, but should it be a runtime dependency of `react-server-cli` instead?
* How important is allowing customizing of webpack loaders and config? What about custom middleware for Express?

I'm sure that there's more that's escaping me at the moment, and I'm sure there's lots of stuff I messed up, but rather than try and anticipate every response, let's just start the discussion. Cheers.

### DIFF
--- a/packages/react-server-cli/README.md
+++ b/packages/react-server-cli/README.md
@@ -61,6 +61,16 @@ We are also considering completely getting rid of server-side rendering in devel
 
 Production mode's priority is optimization at the expense of startup time. A separate code bundle is generated for every entry point into your app so that there is at most just one JS and one CSS file loaded by the framework. All code is minified, and hot reloading is turned off.
 
+####Building static files for production use
+
+In many production configurations, you may not want `react-server-cli` to serve up your static JavaScript and CSS files. Typically, this is because you have a more performant static file server already set up or because you upload all your static files to a CDN server.
+
+To use `react-server-cli` in this sort of production setup, follow these steps:
+
+1. `react-server-cli --production --compile-only` compiles the JavaScript and CSS files into the directory `__clientTemp/build`.
+1. Upload the contents of `__clientTemp/build` to your static file server.
+1. `react-server-cli --production --js-url="http://mystaticfileserver.com/somedirectory/"` to start your HTML server depending on JavaScript and CSS files from your static file server.
+
 ###Setting Options Manually
 
 While development and production mode are good starting points, you can of course choose to override any of the setup by passing in options at the command line:
@@ -93,12 +103,10 @@ Defaults to **false** in development mode and **true** in production.
 #### --compile-only
 Compile the client JavaScript only, and don't start any servers. This is what you want to do if you are building the client JavaScript to be hosted on a CDN or separate server. Unless you have a very specific reason, it's almost always a good idea to only do this in production mode.
 
-For maximum compatibility between servers and compiled JavaScript, this option implies --bundleperroute.
-
 Defaults to **false**.
 
 #### --js-url
-A URL base for the pre-compiled client JavaScript; usually this is a base URL on a CDN or separate server. Setting a value for jsurl means that react-server-cli will not compile the client JavaScript at all, and it will not serve up any of the client JavaScript. Obviously, this means that --jsurl overrides all of the options related to JavaScript compilation: --hot, --minify, and --bundleperroute.
+A URL base for the pre-compiled client JavaScript; usually this is a base URL on a CDN or separate server. Setting a value for js-url means that react-server-cli will not compile the client JavaScript at all, and it will not serve up any of the client JavaScript. Obviously, this means that --js-url overrides and ignores all of the options related to JavaScript compilation and serving: --hot, --js-port, and --minify.
 
 Defaults to **null**.
 
@@ -120,4 +128,6 @@ Here are a few of the things on the unordered wishlist to add to `react-server-c
 * A programmatic API.
 * Automatic compilation of SASS and LESS.
 * Ability to opt out of Babel compilation.
+* Inclusion of CSS Modules
+* Best practices for static file HTTP caching (last-mod, filename hashes, etags, etc.)
 * Help with proxying API endpoints.
--- a/packages/react-server-cli/README.md
+++ b/packages/react-server-cli/README.md
@@ -75,7 +75,7 @@ The port to start up the main server, which will serve the pre-rendered HTML fil
 
 Defaults to **3000**.
 
-#### --jsPort
+#### --js-port
 The port to use when `react-server-cli` is serving up the client JavaScript.
 
 Defaults to **3001**.
@@ -90,19 +90,19 @@ Minify client JavaScript and CSS.
 
 Defaults to **false** in development mode and **true** in production.
 
-#### --compileonly
+#### --compile-only
 Compile the client JavaScript only, and don't start any servers. This is what you want to do if you are building the client JavaScript to be hosted on a CDN or separate server. Unless you have a very specific reason, it's almost always a good idea to only do this in production mode.
 
 For maximum compatibility between servers and compiled JavaScript, this option implies --bundleperroute.
 
 Defaults to **false**.
 
-#### --jsurl
+#### --js-url
 A URL base for the pre-compiled client JavaScript; usually this is a base URL on a CDN or separate server. Setting a value for jsurl means that react-server-cli will not compile the client JavaScript at all, and it will not serve up any of the client JavaScript. Obviously, this means that --jsurl overrides all of the options related to JavaScript compilation: --hot, --minify, and --bundleperroute.
 
 Defaults to **null**.
 
-#### --loglevel
+#### --log-level
 Sets the severity level for the logs being reported. Values are, in ascending order of severity: 'debug', 'info', 'notice', 'warning', 'error', 'critical', 'alert', 'emergency'.
 
 Default is **'debug'** in development mode and **'notice'** in production.
--- a/packages/react-server-cli/README.md
+++ b/packages/react-server-cli/README.md
@@ -1,0 +1,123 @@
+# react-server-cli
+
+A simple command line app that will compile a routes file for the client and start up express. To use:
+
+1. `npm install --save react-server-cli react-server`
+2. Add `./node_modules/react-server-cli/bin/react-server-cli` as the value for `scripts.start` in package.json.
+3. `npm start` from the directory that has your routes.js file.
+
+## Routes Format
+
+Note that the routes file needs to be in a bit different format than what we have used in the past in `react-server`. Rather than `routes.route.page` being a function, it needs to be a path to a file that exports a page class. Middleware also needs to be an array of file paths. For example:
+
+```javascript
+module.exports = {
+	// these will be applied to every page in the site.
+	middleware: ["./FooMiddleware", "./BarMiddleware"],
+
+	// this maps URLs to modules that export a Page class.
+	routes: {
+		BazRoute: {
+			path: ["/"],
+			method: "get",
+			page: "./BazPage"
+		},
+		BakRoute: {
+			path: ["/bak"],
+			method: "get",
+			page: "./BakPage"
+		}
+	}
+};
+```
+
+##What It Does
+
+The CLI builds and runs a `react-server` project, using Express. It compiles JS(X) and CSS into efficiently loadable bundles with code splitting using webpack, and it supports hot reloading of React components on the client-side during development.
+
+##Built-in Features
+
+###Babel Compilation
+It's rare to see a project these days in the JavaScript world that isn't at least experimenting with ES2015 and ES7. To make this easier, all code in your project will be run through Babel, and source maps will be generated back to the original file.
+
+To take advantage of the Babel compilation, you need to install the Babel plugins and presets you want and reference them in a `.babelrc` file in your code directory. For more on the `.babelrc` format, see [its documentation here](https://babeljs.io/docs/usage/babelrc/).
+
+##Options
+Smart defaults are the goal, and `react-server-cli` has two base modes: **development** and **production**. `react-server-cli` will determine which base mode it's in by looking at (in order):
+
+1. If the `--production` flag was sent in, it's **production**.
+1. If `process.env.NODE_ENV` is `'production'`, it's **production**.
+1. Otherwise, the base mode is **development**.
+
+###Development mode: making a great DX
+
+Development mode is the default, and its goals are rapid startup and code-test loops. Hot mode is enabled for all code, although at this time, editing the routes file or modules that export a Page class still requires a browser reload to see changes. Modules that export a React component should reload without a browser refresh.
+
+In development mode, code is not minified in order to speed up startup time, so please do not think that the sizes of bundles in development mode is indicative of how big they will be in production. In fact, it's really best not to do any kind of perf testing in development mode; it will just lead you astray.
+
+We are also considering completely getting rid of server-side rendering in development mode by default to speed startup.
+
+###Production mode: optimizing delivery
+
+Production mode's priority is optimization at the expense of startup time. A separate code bundle is generated for every entry point into your app so that there is at most just one JS and one CSS file loaded by the framework. All code is minified, and hot reloading is turned off.
+
+###Setting Options Manually
+
+While development and production mode are good starting points, you can of course choose to override any of the setup by passing in options at the command line:
+
+#### --routes
+The routes file to load.
+
+Defaults to **"./routes.js"**.
+
+#### --port, -p
+The port to start up the main server, which will serve the pre-rendered HTML files.
+
+Defaults to **3000**.
+
+#### --jsPort
+The port to use when `react-server-cli` is serving up the client JavaScript.
+
+Defaults to **3001**.
+
+#### --hot, -h
+Use hot reloading of client JavaScript. Modules that export React components will reload without refreshing the browser.
+
+Defaults to **true** in development mode and **false** in production mode.
+
+#### --minify, -m
+Minify client JavaScript and CSS.
+
+Defaults to **false** in development mode and **true** in production.
+
+#### --compileonly
+Compile the client JavaScript only, and don't start any servers. This is what you want to do if you are building the client JavaScript to be hosted on a CDN or separate server. Unless you have a very specific reason, it's almost always a good idea to only do this in production mode.
+
+For maximum compatibility between servers and compiled JavaScript, this option implies --bundleperroute.
+
+Defaults to **false**.
+
+#### --jsurl
+A URL base for the pre-compiled client JavaScript; usually this is a base URL on a CDN or separate server. Setting a value for jsurl means that react-server-cli will not compile the client JavaScript at all, and it will not serve up any of the client JavaScript. Obviously, this means that --jsurl overrides all of the options related to JavaScript compilation: --hot, --minify, and --bundleperroute.
+
+Defaults to **null**.
+
+#### --loglevel
+Sets the severity level for the logs being reported. Values are, in ascending order of severity: 'debug', 'info', 'notice', 'warning', 'error', 'critical', 'alert', 'emergency'.
+
+Default is **'debug'** in development mode and **'notice'** in production.
+
+#### --help, -?
+Shows command line options.
+
+##Future Directions
+
+Here are a few of the things on the unordered wishlist to add to `react-server-cli`:
+
+* Default to https, with http as a special case. If no cert & key specified, make a self-signed cert/key combo.
+* Ability to output a completely static site, if the user wants to forgo server-side rendering.
+* Ability to forgo server rendering in development mode.
+* A programmatic API.
+* Automatic compilation of SASS and LESS.
+* Ability to opt out of Babel compilation.
+* Help with proxying API endpoints.
--- a/packages/react-server-cli/gulpfile.babel.js
+++ b/packages/react-server-cli/gulpfile.babel.js
@@ -10,7 +10,7 @@ gulp.task("default", () => {
 		.pipe(gulp.dest("target"));
 });
 
-gulp.task("eslint", [], function() {
+gulp.task("eslint", [], () => {
 	return gulp.src("src/*.js")
         // eslint() attaches the lint output to the eslint property
         // of the file object so it can be used by other modules.
--- a/packages/react-server-cli/gulpfile.babel.js
+++ b/packages/react-server-cli/gulpfile.babel.js
@@ -1,4 +1,5 @@
-var gulp = require("gulp"),
+var eslint = require("gulp-eslint"),
+	gulp = require("gulp"),
 	babel = require("gulp-babel");
 
 gulp.task("default", () => {
@@ -6,3 +7,19 @@ gulp.task("default", () => {
 		.pipe(babel({presets: ["es2015", "react"], plugins: ["transform-runtime"]}))
 		.pipe(gulp.dest("target"));
 });
+
+gulp.task("eslint", [], function() {
+	return gulp.src("src/*.js")
+        // eslint() attaches the lint output to the eslint property
+        // of the file object so it can be used by other modules.
+        .pipe(eslint())
+        // eslint.format() outputs the lint results to the console.
+        // Alternatively use eslint.formatEach() (see Docs).
+        .pipe(eslint.format())
+        // To have the process exit with an error code (1) on
+        // lint error, return the stream and pipe to failOnError last.
+        .pipe(eslint.failAfterError());
+});
+
+// there are no tests for this project :(
+gulp.task("test", ["default"]);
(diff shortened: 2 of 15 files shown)


packages/react-server-cli/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+A simple command line app that will compile a routes file for the client and start up express. To use:
+
+1. `npm install --save react-server-cli react-server`
+2. Add `./node_modules/react-server-cli/bin/react-server-cli` as the value for `scripts.start` in package.json.
+3. `npm start` from the directory that has your routes.js file.
+
+## Routes Format
+
+Note that the routes file needs to be in a bit different format than what we have used in the past in `react-server`. Rather than `routes.route.page` being a function, it needs to be a path to a file that exports a page class. Middleware also needs to be an array of file paths. For example:
```
gigabo: This is a much nicer interface!

It seems like a good direction for `react-server` in general.  Could the transformation machinery be pulled out into a separate package so that it's generally accessible?

gigabo:
I've only just started looking through this, but I have a few immediate rambling thoughts.

1. This is awesome!  It provides a lot of what `react-server` is missing.
2. You do good READMEs.
3. There's a lot in here besides the CLI.  In my mind the CLI package should just provide a tool to help get things set up and running.  I'd like general `react-server` improvements to be generally available.  Seems like some of this should move into `packages/react-server/` or into new packages that can be used independently.
4. Why not global?  I'd like a tool that helps me bootstrap a project from nothing.  Or is that just a different thing?  I guess that's what #2 is about.  How do these two play together?
5. I really like the idea of providing a batteries-included setup.  `react-server` works as an express middleware, but for users who are just getting started it's nice to set the server up for them so they can focus on setting up _pages_ as soon as possible instead of worrying about just getting things running.

I'll dig into this more.  Just wanted to get those first impressions out there.

Thanks for this!

packages/react-server-cli/src/compileClient.js
```diff
@@ -0,0 +1,197 @@
...
+		coreJsMiddleware(${JSON.stringify(staticUrl)}),
+		coreCssMiddleware(${JSON.stringify(staticUrl)}),
+		${existingMiddleware.join(",")}
+	],
+	routes:{`);
+
+	for (let routeName in routes.routes) {
+		if (routes.routes.hasOwnProperty(routeName)) {
+			let route = routes.routes[routeName];
+			var relativePathToPage = path.relative(workingDirAbsolute, path.resolve(routesDir, route.page));
```
gigabo: Heads up: The navigator now supports [_objects_ at `route.page`](https://github.com/redfin/react-server/blob/master/packages/react-server/core/context/Navigator.js#L79-L103) for adaptive routing based on device classification.

packages/react-server-cli/README.md
aickin: Totally; it's almost all in compileClient.js. 

In a general sense, I feel very unsure about the question of how many different packages with public APIs this project should be. It feels too monolithic to me right now, especially munging code compilation and server setup. But I'll admit that I don't know right now what the right divisions are.

packages/react-server-cli/src/compileClient.js
aickin: Oooooh; I'll need to fix that. Are there any documentation or examples outside of the code? Thanks!

aickin:
1. Glad you think so, and thanks so much for taking the time to look at it!
1. Aw shucks, thanks.
1. Yeah, I think proper scope of this project is the thing I have the least clarity on, and I'd love feedback. I wrote it as a separate project this summer partially just because it felt like I couldn't really merge changes back to `react-server`; it wouldn't surprise me if some parts of it belong in `react-server` itself. I feel like there are at least five semi-separate tasks that `react-server-cli` performs right now:
 - Transpile the routes.js file into a code-split version for client and a non-code-split version for server.
 - Add CSS and JavaScript middleware to the routes config so that they will load properly on the client side.
 - Bundle JavaScript and CSS for the client with webpack.
 - Start an HTML server that will render the pages.
 - Start a static server that will serve up bundled JavaScript and CSS for the pages.
These tasks are somewhat separable, but they have interdependencies, especially when you do things like start in hot-reload mode. Do you have any opinions on what the proper project boundaries are?
1. I think my issue with it being global is that it doesn't do project bootstrapping/scaffolding (which would make a lot of sense as a global install), and I fear that `react-server` will change rapidly enough that it will be hard to make one globally installed version of `react-server-cli` understand how to compile code and run servers for multiple different locally installed versions of `react-server` on the same machine. I do agree that bootstrapping is important, and I'm totally agnostic about whether that's a feature of `react-server-cli` or is a yeoman generator a la #2. I could even imagine renaming the yeoman generator `react-server-cli` and this project something like `react-server-compile-start`.
1. Could not agree more. I know from experience that it's actually really complicated to get `react-server` running if you only have the `react-server` middleware; it took me two weeks or so to get this CLI project to where it is now, and I knew a lot about the internals of `react-server`. If we want devs to make sites with `react-server`, the time between `npm install` and getting in to the code-run loop has to be as small as humanly possible.

aickin:
Answering my own question:

> Currently `react-server` is a peer dependency, but should it be a runtime dependency of `react-server-cli` instead?

I remember now why I made it a peer dependency. It's because the app code needs to access `react-server` sometimes, specifically for things like the logger and the `Link` component.

packages/react-server-cli/src/compileClient.js
aickin: OK, so I __think__ I understand it, although for some reason `reduce` always scrambles my already addled brain a bit :cry: . 

My plan: allow `page` in the routes file to be a string or an object of strings. If it's a single string, I will compile it to:

```
page: {
 default: {
  done: {
   function(cb) {
    cb(require(THE_MODULE_STRING)); // or require.ensure for client-side
   }
  }
}
```
and if it's an object of strings, it'll look just like that but with multiple entries named appropriately instead of a single entry named `default`. I'll work on this now.

aickin:
Commit 11a8fa1 should address the issue of route files having page values that are objects. I tested it locally with a project that served up one page to phones and another to desktops, and React connected correctly in both cases. Thanks for letting me know about it, @gigabo!

aickin:
Hey there! I know that this is a big, rough PR, and it's hard to review this much code at once, but I'm wondering if there's anything I could be doing to help it along. Would it help, for example, if I removed some features (like, for example, hot loading) or spent some more time commenting the code? Let me know what would be helpful, and I'll see what I can do.

On the flip side, if you don't want this to be the direction of `react-server` (or if it's too much to maintain or whatever reason really), that's totally OK, and you won't hurt my feelings if you don't want to pull the code. Cheers!

packages/react-server-cli/src/compileClient.js
```diff
@@ -0,0 +1,216 @@
...
+							});`);
+			} else {
+				routesOutput.push(`
+							cb(unwrapEs6Module(require("${relativePathToPage}")));`);
+			}
+			routesOutput.push(`
+						}
+					};
+				},`);
+		}
```
gigabo: Wow... That's... wow.  Braces.  I bet the output of this is nicely indented, huh?

packages/react-server-cli/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+if (isProduction && !process.env.NODE_ENV) { //eslint-disable-line no-process-env
+	process.env.NODE_ENV = "production"; //eslint-disable-line no-process-env
+}
+const argv = parseCliArgs(isProduction);
+
+// TODO: do we need a post-processor for logger here?
+// these require calls are after the argument parsing because we want to set NODE_ENV
+// before they get loaded.
+const logging = require("react-server").logging,
+	logger = logging.getLogger({name: "react-server-cli/index.js", color: {server: 9}}),
```
gigabo: Can this be made to use `react-server-gulp-module-tagger` so we don't have a hard-coded `__LOGGER__` object here?

gigabo:
> it's hard to review this much code at once

Yep.  But worth it.  Definitely want this!  Just need to digest. :meat_on_bone:

packages/react-server-cli/src/compileClient.js
aickin: Yeah, it's... not ideal, although, yes, the output is nicely indented. Do you have any better ideas on code that outputs code? I'm not particularly proud of this solution.

packages/react-server-cli/src/index.js
aickin: I certainly could; I didn't have access to that code when I wrote this.

aickin:
Totally fair; I know it's a lot. Thanks for your time!

packages/react-server-cli/src/compileClient.js
gigabo: > Do you have any better ideas

No!  Code generation is just inherently awkward.

The fact that you're _producing_ readable code outweighs the eye-crossing-ness of this function, IMO.

packages/react-server-cli/gulpfile.babel.js
```diff
@@ -0,0 +1,8 @@
+var gulp = require("gulp"),
+	babel = require("gulp-babel");
+
+gulp.task("default", () => {
+	return gulp.src("src/**/*.js")
+		.pipe(babel({presets: ["es2015", "react"], plugins: ["transform-runtime"]}))
+		.pipe(gulp.dest("target"));
+});
```
gigabo: Please add an "eslint" task so this will get linted by travis.

If you can think of a clean way to get the one in `packages/react-server/gulpfile.js` factored out, even better. :)

You'll also need to patch the top-level gulpfile to pick up `gulpfile.babel.js` files (it currently looks only for `gulpfile.js`).

packages/react-server-cli/src/coreJsMiddleware.js
```diff
@@ -0,0 +1,20 @@
...
+			return [
+				`${baseUrl}${routeName}.bundle.js`,
+				{
+					type: "text/javascript",
+					text: baseUrl ? `
+						if (typeof window !== "undefined" && window.__setReactServerBase) {
+							window.__setReactServerBase(${JSON.stringify(baseUrl)});
+						}` : "",
+				},
+				...next(),
```
gigabo: > `...next(),`

Ooh, that's slick.

packages/react-server-cli/src/coreCssMiddleware.js
```diff
@@ -0,0 +1,13 @@
+
+module.exports = function(pathToStatic) {
```
gigabo: It bothers me a little that these modules export functions that _produce_ middleware instead of just the middleware classes themselves.  I wonder if there's another way we could get `pathToStatic` in here.

Would it be terrible to just make it a global server-side also?  Already doing that browser-side...

packages/react-server-cli/gulpfile.babel.js
aickin: Thanks; should be fixed in b48525e. (I did not try to factor out the eslint task, though...)

packages/react-server-cli/src/coreCssMiddleware.js
aickin: > Already doing that browser-side...

Yeah, although I _think_ we could remove it as a browser-side global now (i.e. get rid of `window.__reactServerBase`; it shouldn't be used any more.) We still have to leave the method that assigns it to `__webpack_public_path__`, though, or else the webpacked code freaks out that it's on a different server than the base HTML.

I hear what you are saying, but these two middlewares are not intended to be externally available, so the only client should be the transpiled routes file, which calls the constructor with the path to the static files. That seems better to me than adding a global that others might start depending on; globals have a way of becoming impossible to modify/remove because it's so hard to find their dependents. I also worry about having a global that's defined in `react-server-cli`; if app code depends on the global, the app won't be portable between `react-server-cli` and raw `react-server`.

So: I'd prefer this not depend on a global, but if it's what's needed to get the code pulled, I can totally do so. ;)

packages/react-server-cli/src/compileClient.js
```diff
@@ -0,0 +1,216 @@
...
+const writeClientBootstrapFile = (outputDir) => {
+	var outputFile = outputDir + "/entry.js";
+	fs.writeFileSync(outputFile, `
+		if (typeof window !== "undefined") {
+			window.__setReactServerBase = function(path) {
+				// according to http://webpack.github.io/docs/configuration.html#output-publicpath
+				// we should never set __webpack_public_path__ when hot module replacement is on.
+				if (!module.hot) {
+					__webpack_public_path__ = path;
+					window.__reactServerBase = path;
```
aickin: I don't think that `window.__reactServerBase` is used any more. I should test taking it out in a few different scenarios.

packages/react-server-cli/src/coreCssMiddleware.js
gigabo: No, you're right.  A global would be far worse.  I wonder about a static method that sets the prefix?

Anyway, not something that needs to hold this PR up!

packages/react-server-cli/src/parseCliArgs.js
```diff
@@ -0,0 +1,56 @@
...
+		.option("loglevel", {
+			default: isProduction ? "notice" : "debug",
+			describe: "Set the severity level for the logs being reported. Values are, in ascending order of severity: 'debug', 'info', 'notice', 'warning', 'error', 'critical', 'alert', 'emergency'. Default is 'notice' in production mode, 'debug' otherwise.",
+			type: "string",
+		})
+		.option("compileonly", {
+			default: false,
+			describe: "Compile the client JavaScript only, and don't start any servers. This is what you want to do if you are building the client JavaScript to be hosted on a CDN. Unless you have a very specific reason, it's almost alway a good idea to only do this in production mode. Defaults to false.",
+			type: "boolean",
+		})
```
gigabo: `jsPort` is camel.  `compileonly` is not.  Would match code if it were.

Would be _extra_ slick if it were `--dash-separated` and auto-camel'd à la [meow](https://www.npmjs.com/package/meow).

packages/react-server-cli/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+startServer(argv.routes, {
+	port: argv.port,
+	jsPort: argv.jsPort,
+	hot: argv.hot,
+	minify: argv.minify,
+	compileOnly: argv.compileonly,
+	jsUrl: argv.jsurl,
+});
```
gigabo: Why the big explicit copy?

Seems like less friction:

```javascript
startServer(argv.routes, argv)
```

(Assuming the `compileonly` option gets camel'd somehow)

packages/react-server-cli/src/compileClient.js
gigabo: Sweet!

packages/react-server-cli/src/coreCssMiddleware.js
aickin: Ugh, I think I was wrong about being able to get rid of `window.__reactServerBase`. It's used when the JS is compiled using `compileOnly` and the server is launched with `jsUrl` to tell the server code where the compiled client code lives.

In that case, the compiled client JS has `null` for `pathToStatic`, whereas the server code has the real runtime `pathToStatic`. `window.__reactServerBase` is a sneaky way to pass the path from the server to the client without having to recompile the client routes file.

packages/react-server-cli/src/compileClient.js
aickin: See above note; I think I was wrong in the case where `compileOnly` is used to make a compiled client bundle that can be deployed at any URL. Sigh.

gigabo:
I've added some comments for minor things.  None of those is a blocker.

I'd like to get some of the build stuff pulled out into a separate package, but that doesn't have to happen immediately, either.

@roblg, @doug-wade: This looks good to me.  Have you guys had a chance to look closely?

---
Regarding @aickin's bikeshedding topics:

> Is react-server-cli even a good name?

Seems appropriate to me.

> Should routes just be a JSON file rather than a JavaScript file?

Ugh.  There are definitely benefits to that, but actual JSON is a pain.  No comments, quotes all over the place, commas breaking things.   I vote no.

> Is it right to have a bin directory with an executable file?

Uh.... yes?  @doug-wade?

> Currently react-server is a peer dependency, but should it be a runtime dependency of react-server-cli instead?

Peer dependency seems right to me.

> How important is allowing customizing of webpack loaders and config? What about custom middleware for Express?

Probably eventually important, but near term not?

---
I'd like to get this merged in so we can move on to @aickin's next awesome thing.

packages/react-server-cli/src/index.js
aickin: Good point; fixed in aa5411c. Thanks!

packages/react-server-cli/src/parseCliArgs.js
aickin: Good point. Standardized the args in aa5411c so that the command line args are all hyphenated and the JavaScript representations are all camelCased. The changed args are `log-level`, `js-url`, `js-port`, and `compile-only`. Thanks!

packages/react-server-cli/src/index.js
aickin: Fixed in 01a40ab; thanks!

roblg:
@gigabo @aickin I've now attempted to look at this numerous times and also seem to get distracted. I doubt I'd be spotting many bugs w/o pulling it down and trying to break it.

I like the direction, and it fills a need that we have, so I say let's get it in, and we'll squash any bugs that pop up.

:+1:

packages/react-server-cli/src/compileClient.js
```diff
@@ -0,0 +1,222 @@
+import webpack from "webpack"
+import path from "path"
+import mkdirp from "mkdirp"
+import fs from "fs"
+import ExtractTextPlugin from "extract-text-webpack-plugin"
+
+// commented out to please eslint, but re-add if logging is needed in this file.
```
doug-wade: You can avoid the "uncomment to use" by suppressing the eslint no-unused-vars rule for just these two lines with an eslint enable/disable block, if you prefer: http://eslint.org/docs/user-guide/configuring#configuring-rules

packages/react-server-cli/src/compileClient.js
```diff
@@ -0,0 +1,222 @@
...
+
+	// for each route, let's create an entrypoint file that includes the page file and the routes file
+	let bootstrapFile = writeClientBootstrapFile(workingDirAbsolute);
+	const entrypointBase = hot ? [`webpack-dev-server/client?${outputUrl}`,"webpack/hot/only-dev-server"] : [];
+	let entrypoints = {};
+	for (let routeName of Object.keys(routes.routes)) {
+		let route = routes.routes[routeName];
+		let formats = normalizeRoutesPage(route.page);
+		for (let format of Object.keys(formats)) {
+			const absolutePathToPage = path.resolve(routesDirAbsolute, formats[format]);
```
doug-wade: why do you use let for route and formats but const for absolutePathToPage?  Iiuc, you never change the reference for route or formats, so you could use const?  (I'm not really recommending a change, mostly just trying to catch up)

packages/react-server-cli/src/compileClient.js
```diff
@@ -0,0 +1,222 @@
...
+		plugins: [
+			new ExtractTextPlugin("[name].css"),
+		],
+	};
+
+	if (minify) {
+		webpackConfig.plugins = [
+			...webpackConfig.plugins,
+			new webpack.DefinePlugin({
+				'process.env': {NODE_ENV: '"production"'},
```
doug-wade: this seems unintuitive -- I feel like I should be able to minify my js for local testing without using all the rest of my production environment config.

packages/react-server-cli/src/compileClient.js
```diff
@@ -0,0 +1,222 @@
...
+			}),
+			// TODO: should this be done as babel plugin?
```
doug-wade: Probably not?  you don't want to uglify server side, and I'd like to try to keep the babel config as similar as possible on the server and client (I _think_ we should be transpiling them both pretty aggressively, since V8 features often start out pretty slow, and we want to support the oldest browsers reasonably possible)

packages/react-server-cli/src/compileClient.js
doug-wade: probably out-of-scope, but I generally prefer using a templating engine for code generation, personally

doug-wade:
Finally made it through.  re: Is it right to have a bin directory with an executable file? -- iiuc, yes.

I had some thoughts, but my feeling is let's merge this and sort them out later.

gigabo:
:rocket:

packages/react-server-cli/src/compileClient.js
aickin: Oh huh, it looks like you're right. 

For some reason I thought that because the loop variable is reassigned each iteration that it had to be `let`. Looks like that's not the case, so it should totally be `const`. Thanks!

packages/react-server-cli/src/compileClient.js
aickin: Hmmmm... interesting point. The issue is that I believe React still doesn't minify fully unless you set NODE_ENV to production, so that the env checks for dev mode can be minified out. Without `production` in the minification step, it won't truly be minified. I see your point, though, and I don't know how to reconcile it.

aickin:
:tada: 

w0000000t. Thanks so much for your time and work, @doug-wade, @roblg, and @gigabo. Yeehaw.

Two things still on my todo list for this (I'm in an all-day meeting and shouldn't actually be looking at this today :smile: ):

1. Responding to the rest of @doug-wade's comments.
1. Adding someone at Redfin as an npm owner for `react-server-cli`. (Who should that be? @gigabo?)

gigabo:
> Adding someone at Redfin as an npm owner for `react-server-cli`. (Who should that be? @gigabo?)

Oh, hmm... yes, please.  Additionally would you mind if I add others here as necessary?

aickin:
Of course not. In fact:

```
Dear lawyers,
 I hereby donate the npm name `react-server-cli` and `react-server` to this project.
``` 

I would love to be left on as an owner, but it's your call.

doug-wade:
iiuc, owner can be an array (https://docs.npmjs.com/cli/owner).  We may instead want to create a team and grant permissions to team members, rather than making everyone an owner (https://docs.npmjs.com/cli/access and https://docs.npmjs.com/cli/team).

aickin:
Yeah, I already added @gigabo as an owner to react-server; I just need to do the same for react-server-cli. I happily leave the team/owner structure decisions to @gigabo. :smile:

doug-wade:
I swung by the npm irc channel, and they seem to recommend setting up an organization to manage our permissions.  @gigabo -- wanna be the super-admin (https://docs.npmjs.com/orgs/roles#super-admin) ?

gigabo:
Uhhh...  Step 2 of creating an organization:

image

kcjonson:
onceyouseeit

gigabo:
@kcjonson - I love that _that's_ what bothers you about that image. :grinning:

aickin:
@gigabo officially added as an owner to npm `react-server-cli`. Do your worst!

gigabo:
Thanks @aickin.  We should get another version out this week.

Excited to see what you drop next!

packages/react-server-cli/src/compileClient.js
aickin: Eventually, I would love to be able to send differently compiled versions of the code to different browsers based on their native ES2015 capabilities, but that would require a minifier that understands ES2015. Uglify is working on ES2015 support, but it's somewhat slow going. I've heard through the grapevine that Babel may be working on a minifier plugin that would be totally ES2015 compatible, though, which is tempting.

All of this is theoretical for now, though, so no real reason to fight over it. :smile_cat:

packages/react-server-cli/src/compileClient.js
aickin: What's the project feeling about suppressing linting rules? I sort of assumed that it would be frowned upon unless completely necessary.

gigabo: Maybe slightly less frowned upon than leaving commented-out code laying around? ;)

PR #18, opened 2016-02-14
Add a .gitignore for node_modules at the root

aickin:
When you `npm install` from the root directory, git thinks `node_modules` need to be checked in. This fixes that. Cheers.


gigabo:
Nice, thanks!

PR #19, opened 2016-02-15
Upgrade to react-dom; eliminated deprecated calls into react

aickin:
In 0.14.0, React deprecated `render`, `renderToString`, and `renderToStaticMarkup` in the `react` package and moved them to `react-dom` and `react-dom/server`.

This tiny PR upgrades all the calls I could find to the old functions, making the code forward-compatible, and it eliminates console deprecation warnings I was seeing both on server and client. Cheers!


roblg:
Yep. We've been meaning to do this for awhile.

PR #20, opened 2016-02-15
Bumping the react peer dependencies

aickin:
`react-server` currently has a peer dependency of a specific version of `react` and `react-dom` (0.14.2), which makes it incompatible with the most recent bug fix release of those packages. (Well, it makes npm claim it's incompatible, but the code works just fine if you install the latest `react(-dom)`.)

This PR allows `react-server` to be installed with any of the 0.14.x releases of `react(-dom)`, for x >= 2. It doesn't allow 0.15 (which may be overly conservative?).

I think this should also be done for `superagent`, but I wasn't as familiar with its versions and potential upgrade headaches, so I left it be for now. Thanks!

### DIFF
--- a/packages/react-server/package.json
+++ b/packages/react-server/package.json
@@ -35,8 +35,8 @@
     "winston": "0.8.3"
   },
   "peerDependencies": {
-    "react": "0.14.2",
-    "react-dom": "0.14.2",
+    "react": "~0.14.2",
+    "react-dom": "~0.14.2",
     "superagent": "1.2.0"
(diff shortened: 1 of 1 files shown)


roblg:
I think it makes sense to be conservative w/ React versions for now. They're pretty good about not breaking things unexpectedly between releases, but I think it still probably makes sense to know what's in the release before we claim to support it. But then again, I'm pretty conservative myself...

In any event, this patch makes sense.

PR #22, opened 2016-02-15
Move tests out of core

roblg:
This PR does a couple different test-related things in an attempt to make it more approachable to write new tests (because we need better coverage :grin:):

* Moves tests into `__tests__/` subdirectories that are siblings of the modules that they test (like jest).
* Makes page class paths passed to `startServerBeforeAll` relative to the spec file, rather than relative to a parent directory. (It requires using `__filename` in the spec, which I don't like, but I couldn't figure out how to get jasmine to tell me which spec was running, and I dislike this less than I dislike having paths be relative to non-obvious things.)
* Adds some better error handling around the webpack build part of the tests.
* Updates the integration test documentation


packages/react-server/core/__tests__/integration/redirectForward/RedirectForwardSpec.js
```diff
@@ -1,12 +1,12 @@
-var helper = require("../specRuntime/testHelper"),
+var helper = require("../../../test/specRuntime/testHelper"),
```
gigabo: Eesh.  That's a lot of dots.

gigabo: What if we moved the test helpers into a separate package?  Then they could be required without worrying about relative location.

packages/react-server/core/__tests__/integration/README.md
```diff
@@ -1,10 +1,10 @@
-# Testing Triton
```
gigabo: This README is a little buried now.  It's too good to be so hidden.

packages/react-server/core/__tests__/integration/containers/pages/SimpleContainerPage.js
```diff
@@ -0,0 +1,13 @@
+
```
gigabo: Yeah... once this lands I want to set up some tests for `<RootElement />` and `<RootContainer />`.

packages/react-server/core/__tests__/integration/redirectForward/RedirectForwardSpec.js
roblg: > What if we moved the test helpers into a separate package? Then they could be required without worrying about relative location.
I had that thought too, but it seems like a weird thing to publish a separate module for? I'm not opposed to it though.

gigabo: Well, not urgent.

I can't say how many times I'll be willing to type `../../../` before I snap and do _something_, though. ;)

roblg:
Going to merge to unblock more test-writing. Other feedback can be addressed in new PRs

packages/react-server/core/__tests__/integration/redirectForward/RedirectForwardSpec.js
aickin: How about:
- helpers in a separate directory under packages
- don't publish to npm
- include a dev dependency in `react-server` package.json of 
```javascript
"react-server-test-helpers": "file:../react-server-test-helpers"
```

Then you can require the package like normal, and I __think__ dev dependencies get wiped away when published to npm.

gigabo: Why don't we want `react-server-test-helpers` in npm?  Could they be useful for users to test their pages?

roblg: Seems like that would work, assuming `lerna` does sane things with local dependencies. I'll give it a try.

Issue #6, opened 2016-02-11
Don't run `gulp test` in npm prepublish


gigabo:
Prepublish runs on `npm install`.  This is apparently [by design](https://github.com/npm/npm/issues/3059#issuecomment-12282746) (though controversial).

The recommendation seems to be: Have a script that you run to publish that does stuff before publishing.

PR #24, opened 2016-02-16
No `gulp test` in prepublish

gigabo:
Fixes https://github.com/redfin/react-server/issues/6.

Currently redundantly testing during `npm install` in Travis.

Our actual publishing to npm also frequently times out and has to retry.

Can we just publish when travis is green or we've manually run `gulp test` ourselves?

Do we need a `publish` script that wraps testing and publishing as recommended
in https://github.com/npm/npm/issues/3059#issuecomment-12282746?

### DIFF
--- a/packages/react-server/package.json
+++ b/packages/react-server/package.json
@@ -8,7 +8,7 @@
   "author": "Navtej",
   "license": "Apache License 2.0",
   "scripts": {
-    "prepublish": "gulp build && gulp test",
+    "prepublish": "gulp build",
     "test": "gulp test"
   },
   "files": [


roblg:
I think lerna has a way to specify a per-project prepublish file, or something - maybe that could run `gulp test`. I have no idea if that's better than what's going on here, but it seems like it might be, because it wouldn't be running during npm's prepublish step.

gigabo:
Yes, lerna supports [per-project prepublish scripts](https://github.com/kittens/lerna/blob/master/lib/commands/publish/npmPublishAsTemp.js#L28-L30), but it doesn't look like they check the output at all.  So if our tests failed, we'd still proceed to publish.

Is running our tests (and linter) in CI enough?

doug-wade:
seems reasonable to me -- I don't think we want anyone other than the CI bot to publish.

gigabo:
> I don't think we want anyone other than the CI bot to publish.

Uhh... we've been publishing by hand so far. :grin:

doug-wade:
ed: ideally, no-one other than the CI bot would publish.

gigabo:
Merging this.  Publishing fails too often.

PR #27, opened 2016-02-18
Support page rendering with non-2xx responses

gigabo:
Default behavior is unchanged.

Page may elect to proceed with ordinary rendering via `setHaveDocument()`.

Response is still be set by `handleRoute()` return value.

Example:

```javascript
handleRoute() {
    this.setHaveDocument(true);
    return { code: 404 };
}

getTitle() {
    return "404 Not Found";
}

getElements() {
    ...
}
```

### DIFF
--- a/packages/react-server/core/__tests__/integration/internalServerError/InternalServerErrorSpec.js
+++ b/packages/react-server/core/__tests__/integration/internalServerError/InternalServerErrorSpec.js
@@ -1,0 +1,43 @@
+var helper = require("../../../test/specRuntime/testHelper");
+var Browser = require("zombie");
+
+describe("A 500 internal server error page", () => {
+
+	a500("has no body by default", '/internalServerErrorNoDocument', txt => {
+		// Yikes... not good default behavior.
+		expect(txt).toBe('[object Object]\n')
+	});
+
+	a500("has a body with `haveDocument(true)`", '/internalServerErrorWithDocument', txt => {
+		expect(txt).not.toMatch('Cannot GET /internalServerErrorNoDocument')
+		expect(txt).toMatch('foo</title>')
+		expect(txt).toMatch('foo</div>')
+	});
+
+	a500("can result from an exception during `handleRoute()`", '/internalServerErrorException', txt => {
+		expect(txt).toBe('[object Object]\n')
+	}, xit); // Pending... why doesn't this work?
+
+	a500("can result from a rejection from `handleRoute()`", '/internalServerErrorRejection', txt => {
+		expect(txt).toBe('[object Object]\n')
+	});
+
+	// Pass `xit` for `the500` to mark a test as pending.
+	function a500(spec, url, callback, the500=it) {
+		the500(spec, done => new Browser()
+			.fetch(`http://localhost:${helper.getPort()}${url}`)
+			.then(res => (expect(res.status).toBe(500), res.text()))
+			.then(callback)
+			.then(done)
+		);
+	}
+
+	helper.startServerBeforeAll(__filename, [
+		"./pages/InternalServerErrorNoDocument",
+		"./pages/InternalServerErrorWithDocument",
+		"./pages/InternalServerErrorException",
+		"./pages/InternalServerErrorRejection",
+	]);
+
+	helper.stopServerAfterAll();
+});
--- a/packages/react-server/core/__tests__/integration/internalServerError/InternalServerErrorSpec.js
+++ b/packages/react-server/core/__tests__/integration/internalServerError/InternalServerErrorSpec.js
@@ -8,7 +8,7 @@ describe("A 500 internal server error page", () => {
 		expect(txt).toBe('[object Object]\n')
 	});
 
-	a500("has a body with `haveDocument(true)`", '/internalServerErrorWithDocument', txt => {
+	a500("has a body with `hasDocument: true`", '/internalServerErrorWithDocument', txt => {
 		expect(txt).not.toMatch('Cannot GET /internalServerErrorNoDocument')
 		expect(txt).toMatch('foo</title>')
 		expect(txt).toMatch('foo</div>')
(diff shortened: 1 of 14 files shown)


gigabo:
I'm going to set up some tests for 500 pages, too.

Putting this out for feedback on the interface first.

gigabo:
Oops.  Accidentally deleted the branch, which closed the PR... uh...

gigabo:
Okay, I added basic tests for 500 pages with and without document render.

Tests uncovered some weird behavior WRT exceptions arising during `handleRoute()` which I'll patch with a separate PR.

packages/react-server/core/__tests__/integration/internalServerError/InternalServerErrorSpec.js
```diff
@@ -0,0 +1,43 @@
...
+	a500("can result from an exception during `handleRoute()`", '/internalServerErrorException', txt => {
+		expect(txt).toBe('[object Object]\n')
+	}, xit); // Pending... why doesn't this work?
+
+	a500("can result from a rejection from `handleRoute()`", '/internalServerErrorRejection', txt => {
+		expect(txt).toBe('[object Object]\n')
+	});
+
+	// Pass `xit` for `the500` to mark a test as pending.
+	function a500(spec, url, callback, the500=it) {
```
roblg: Clever :)

packages/react-server/core/renderMiddleware.js
```diff
@@ -105,16 +105,25 @@ module.exports = function(server, routes) {
...
+				// The page can elect to proceed to render
+				// even with a non-2xx response.  If it
+				// _doesn't_ do so then we're done.
+				var done = !(page && page.getHaveDocument());
+
+				if (err.status === 301 || err.status === 302 || err.status === 307) {
 					res.redirect(err.status, err.redirectUrl);
-				} else {
-					next(err);
+				} else if (done) {
```
roblg: It looks like the previous behavior was to issue `res.redirect`, and then `return;` since the return statement moved inside `if (done)` down below, it's not obvious that that's the case anymore. (In any event, I was under the impression 301, 302 aren't supposed to have additional content?)

roblg:
Love the tests!

And I know I'm a little late on this, but I'm a little uncertain about the `setHaveDocument(...)` API. If it's something that's going to be very rare to use, it's probably OK, but I wonder if it makes more sense as something like: `return {code: 500, haveDocument: true}`? That wouldn't allow us to render a document in the case where there's a `throw`, but if someone _really_ wanted that behavior, they could wrap their whole handleRoute in a promise or a try/catch?

gigabo:
>  I'm a little uncertain about the `setHaveDocument(...)` API.

Yeah, same.  Feels weird.

> something like: return `{code: 500, haveDocument: true}`?

Yeah, thought about that initially, but I was worried about introducing another magic key in the `handleRoute()` response object.  I guess maybe that's just a matter of documentation?  I like the _look_ of it more, and it ties the "... but I have a document" more closely to the "my response code is 500".

> wouldn't allow us to render a document in the case where there's a `throw`

That's explicitly disallowed [here](https://github.com/redfin/react-server/pull/27/files#diff-99c3399df939946ce08560fc1a6d4e81R185), but no special handling would be necessary with a new key in the `handleRoute()` response object...

Okay, you've convinced me!  I'll push an update that moves the trigger into the response object.

packages/react-server/core/renderMiddleware.js
gigabo: Ugh, this diff wound up looking weird...  Moved some things around in here, but the behavior should still be appropriate:

- If the page _hasn't_ advertised `hasDocument` then we're done _no matter what `err` is_.  301, 500, whatever.
- If the page _has_ advertised `hasDocument` then we'll proceed to render.

> 301, 302 aren't supposed to have additional content?

It's valid (and recommended) to provide a small body for redirects.

https://tools.ietf.org/html/rfc2616#section-10.3.2

> > Unless the request method was HEAD, the entity of the response SHOULD contain a short hypertext note with a hyperlink to the new URI(s).

roblg: Sorry, yeah, I was unclear in what I meant. I think `res.redirect` adds the appropriate body -- the target URL for the redirect. I don't think there's ever a case where we'd want to fall through _past_ the `if (done)` call below, and it's hard to tell whether or not that's the case just by looking at this code. (Though it probably is the case? Unless somebody says `setHaveDocument` and then sends a redirect?).

I wonder if just saying `done = true` right under the `res.redirect` call would be clearer.

gigabo: >  I think `res.redirect` adds the appropriate body

Oh, I see.  Thanks, I didn't realize that.  I'll push a patch to make `hasDocument` work for redirects.

> I don't think there's ever a case where we'd want to fall through

The case is: I want to customize the content of my redirect body.  It's cool that express provides a helper that populates a default, but I want the ability to render my own.

That means I'm responsible for actually putting the link in, but presumably if I'm claiming `hasDocument` on a redirect I know what I'm doing?

roblg: > That means I'm responsible for actually putting the link in, but presumably if I'm claiming hasDocument on a redirect I know what I'm doing?

w/ the updated impl that moves `hasDocument` into the return object, I think I agree. It's weird (I have literally no idea how you'd implement that behavior w/ spring or raw java servlets), but as long as supporting it doesn't make the code more complicated, I think it seems reasonable. I think w/ the `setHaveDocument` approach, it'd be possible to intend for setHaveDocument to work for 500s or something, and accidentally have it apply to 503s as well. (Weird/strange edge case either way, I know. :) 
)

gigabo:
@roblg: :+1:?

gigabo:
Ugh.  Just realized I pushed my changes to `redfin/react-server` last night instead of `gigabo/react-server`. :disappointed:

packages/react-server/core/context/Navigator.js
```diff
@@ -165,10 +165,12 @@ class Navigator extends EventEmitter {
 
 			page.setStatus(handleRouteResult.code);
 
+			page.setHasDocument(handleRouteResult.hasDocument);
```
roblg: :+1: I like this better.

roblg:
LGTM

PR #28, opened 2016-02-19
Handleroute exceptions

gigabo:
Pretty simple.  Funnel exceptions arising during `handleRoute()` through the navigator's `catch()`.

### DIFF
--- a/packages/react-server/core/__tests__/integration/internalServerError/pages/InternalServerErrorException.js
+++ b/packages/react-server/core/__tests__/integration/internalServerError/pages/InternalServerErrorException.js
@@ -3,8 +3,9 @@ import React from "react"
 export default class InternalServerErrorException {
 	handleRoute() {
 
-		// Not throwing an `Error` since this is easier to match.
-		throw "died";
+		// Need to throw a real error here, since CLS is going to
+		// assign a property to the error to track its source context.
+		throw new Error("died");
 
 		return {code: 200};
 	}
--- a/packages/react-server/core/context/Navigator.js
+++ b/packages/react-server/core/context/Navigator.js
@@ -161,7 +161,10 @@ class Navigator extends EventEmitter {
 
 		// call page.handleRoute(), and use the resulting code to decide how to
 		// respond.
-		page.handleRoute().then(handleRouteResult => {
+		// We call it in a promise handler so any exception that
+		// arises will get converted to a rejection that we can handle
+		// below.
+		Q().then(page.handleRoute).then(handleRouteResult => {
 
 			page.setStatus(handleRouteResult.code);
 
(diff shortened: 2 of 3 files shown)


packages/react-server/core/__tests__/integration/internalServerError/pages/InternalServerErrorException.js
```diff
@@ -3,8 +3,9 @@ import React from "react"
 export default class InternalServerErrorException {
 	handleRoute() {
 
-		// Not throwing an `Error` since this is easier to match.
-		throw "died";
+		// Need to throw a real error here, since CLS is going to
+		// assign a property to the error to track its source context.
+		throw new Error("died");
```
roblg: If somebody in app code were to throw a string, it just gets no context, right (it doesn't blow up or anything?). I'm pretty sure we've done this before accidentally...

packages/react-server/core/context/Navigator.js
```diff
@@ -161,7 +161,10 @@ class Navigator extends EventEmitter {
 
 		// call page.handleRoute(), and use the resulting code to decide how to
 		// respond.
-		page.handleRoute().then(handleRouteResult => {
+		// We call it in a promise handler so any exception that
+		// arises will get converted to a rejection that we can handle
+		// below.
+		Q().then(page.handleRoute).then(handleRouteResult => {
```
roblg: Since `handleRoute` can return a constant, I think we want `Q().then(Q(page.handleRoute))...`?

gigabo: I _think_ that's done implicitly to the return-values of `.then()` callbacks.

packages/react-server/core/__tests__/integration/internalServerError/pages/InternalServerErrorException.js
gigabo: Yeah, it gets no context and the `logger.error()` message is about failing to bind the context instead of about failing to navigate.  The message includes the thrown text either way.

This diff is mostly just about not leaving examples of anti-patterns laying around in tests. :sweat_smile:

packages/react-server/core/context/Navigator.js
gigabo: @roblg Any lingering concern here?

AFAIK `Q().then(foo)` handles arbitrary return values from `foo` appropriately here.

roblg: Ah, um, hmm, weird. I _swear_ that I tried this exact thing before I wrote that message earlier, and it didn't resolve the promise.
```
> var foo = () => { return {foo: "bar"} };
undefined
> Q().then(foo).then(x => console.log(x))
```
But I just did it again, and it resolved so... :shrug:

LGTM :+1:

PR #31, opened 2016-02-23
Fix a syntax error in the generated routes file with multiple routes

aickin:
I made a mistake in the code generation when I refactored for allowing multiple entries for `page`. Missed a comma, and any site with multiple routes breaks. Didn't see it because I was only testing with one route. Sorry!


gigabo:
Thanks.

PR #35, opened 2016-02-24
Kill the requestLocalStorageNamespaces gauge

gigabo:
This isn't something that needs to be logged per request.

The module is exported, so users can track this themselves if necessary.


roblg:
LGTM. :+1:

doug-wade:
:+1:

PR #38, opened 2016-02-25
Upgrade to latest winston

gigabo:
This is preparation for upgrading to the _next_ winston which introduces
support for logging react elements (which contain symbols) without choking.

https://github.com/winstonjs/winston/issues/805


doug-wade:
:+1:

roblg:
Sweeeeeet. (Where's the dancing corgi emoji Github? Come on.)

PR #43, opened 2016-03-02
Add guards in server-side logger setters

gigabo:
These were somehow failing during corvair client tests.  The failure is at
`require` time when `setColorize` and `setTimestamp` are both set.  Maybe
loggers leaking between client/server tests?  Not sure why this just cropped
up now, but it's easy to guard against.


roblg:
LGTM :+1:

Issue #44, opened 2016-03-03
Should "clean" before publishing


roblg:
To make sure that we're not publishing old crufty files, we should add a "clean" step that we run before building to publish.

PR #45, opened 2016-03-03
Exclude tests and test code from published package

roblg:
Fixing up our package.json and .npmignore so that we're not publishing tests + test helpers anymore.

I published a gist with the [diff between the old and new packages](https://gist.github.com/roblg/ac37b61e212483c60f85).

I also noticed while doing this that we were still accidentally publishing a bunch of test files. I filed #44 to address cleaning before publishing.

### DIFF
--- a/packages/react-server/.npmignore
+++ b/packages/react-server/.npmignore
@@ -1,3 +1,16 @@
+# Goal: include everything in target/, buildutils/, _except_ for __tests__ dirs
 
-**/__tests__/**
+# test-specific dirs
+**/__tests__/
+/target/test-temp
+/target/*/test/
 
+# source/docs dirs
+/core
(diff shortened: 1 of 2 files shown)


gigabo:
Nice!  Leaner and meaner! :+1:

roblg:
I probably should have mentioned that this brings the package.tgz file size down from 1.2MB to 350K. :joy_cat:

roblg:
I also also should have mentioned that I published a pre-release version of 0.0.24 w/ only this change, and wired up our internal app and all of its dependencies to use it. It was a PITA to plumb that change through everywhere, but once I did, everything started, so I don't think I'm missing anything important.

doug-wade:
:+1:

Issue #34, opened 2016-02-24
react-server-cli does not work with npm 2.x


aickin:
As reported in #2 at comment https://github.com/redfin/react-server/pull/2#issuecomment-188311662

To repro:
1. Use npm 2.x (I used 2.14.20)
1. `git clone https://github.com/doug-wade/example-reactserver.git`
1. `cd example-reactserver/`
1. `npm install`
1. `./node_modules/react-server-cli/bin/react-server-cli`

Gives this result:

```
2016-02-24T15:40:09.989Z - warning: [react-server-cli/index.js] PRODUCTION WARNING: the following current settings are discouraged in production environments. (If you are developing, carry on!):
2016-02-24T15:40:09.991Z - warning: [react-server-cli/index.js] -- Hot reload is enabled. Pass --hot=false, pass --production, or set NODE_ENV=production to turn off.
2016-02-24T15:40:09.991Z - warning: [react-server-cli/index.js] -- Minification is disabled. Pass --minify, pass --production, or set NODE_ENV=production to turn on.
2016-02-24T15:40:09.991Z - warning: [react-server-cli/index.js] -- NODE_ENV is not set to "production".
2016-02-24T15:40:13.584Z - notice: [react-server-cli/startServer.js] Starting servers...
2016-02-24T15:40:13.584Z - info: [react-server-cli/startServer.js] Starting hot reload JavaScript server...
2016-02-24T15:40:13.628Z - info: [react-server-cli/startServer.js] Starting HTML server...
2016-02-24T15:40:14.047Z - info: [react-server-cli/startServer.js] Started HTML server on port 3000
2016-02-24T15:40:14.174Z - info: [react-server-cli/startServer.js] Started hot reload JavaScript server on port 3001
2016-02-24T15:40:14.174Z - notice: [react-server-cli/startServer.js] Ready for requests on port 3000.
Hash: 22a0d3906fbc6214957a
Version: webpack 1.12.14
Time: 488ms
               Asset     Size  Chunks       Chunk Names
HelloWorld.bundle.js  22.3 kB       0       HelloWorld
chunk    {0} HelloWorld.bundle.js (HelloWorld) 64 bytes [rendered]
    [0] multi HelloWorld 64 bytes {0} [built] [4 errors]

ERROR in multi HelloWorld
Module not found: Error: Cannot resolve module 'webpack-dev-server/client' in /Users/sashaaickin/code/example-reactserver
 @ multi HelloWorld

ERROR in multi HelloWorld
[...]

PR #46, opened 2016-03-04
Partial fix for #34. npm v2 now works as long as hot mode is disabled.

aickin:
This is a simple partial fix for #34, adding two libraries as peer dependencies for `react-server-cli`. In both npm 2 and npm 3, I was able to do the following:

1. clone `example-reactserver`
1. `npm install`
1. (for npm 3 only) `npm install` any uninstalled peer dependencies.
1. `./node_modules/react-server-cli/bin/react-server-cli --hot=false`
1. open `localhost:3000`

And things worked. Huzzah.

I still have dependency work to do to make hot reload mode work with npm v2, so #34 should stay open.

~~(Also, note that in order to get `example-reactserver` to serve up its page correctly, [this bug fix PR](https://github.com/doug-wade/example-reactserver/pull/1) needs to be pulled.)~~ (PR was merged.)

### DIFF
--- a/packages/react-server-cli/package.json
+++ b/packages/react-server-cli/package.json
@@ -39,5 +39,9 @@
     "gulp-babel": "~6.1.2",
     "gulp-eslint": "0.15.0",
     "react-server-gulp-module-tagger": "0.0.8"
+  },
+  "peerDependencies": {
+    "babel-loader": "~6.2",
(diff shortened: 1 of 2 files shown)


aickin:
I found a better way to do this that fixes both hot and non-hot mode, so I'm going to close this PR to make it clearer.

Issue #12, opened 2016-02-12
Pass in the token to be replaced in the module tagger


gigabo:
Something like:

```javascript
var tagger = require("react-server-gulp-module-tagger");
...
gulp.src(...).pipe(tagger("__LOGGER__"));
```

doug-wade:
To allow using different tokens in client code, rather than __LOGGER__?  And we should def keep the current values as defaults so that you can still `.pipe(tagger())`

doug-wade:
This is now possible with the babel plugin, but I think we should leave the issue until we support it in the gulp plugin as well.

PR #47, opened 2016-03-05
Add __CACHE__ as a supported module tag

gigabo:
Just 'cause, you know... it might be useful.

Wouldn't be necessary if #12 were done, but this is way faster.


roblg:
I want :shrug: back!

LGTM :+1:

PR #49, opened 2016-03-07
Add HTTPS support to react-server-cli

aickin:
This adds the ability to start up `react-server-cli` with the most common configurations for HTTPS. The API of `startServer` can take in any of the https options that can be sent to `https.createServer`, although only `key`, `cert`, `ca`, `pfx`, and `passphrase` are supported at the command line.

I also added (maybe foolishly?) the ability to start up with a self-signed certificate for quick and dirty testing. This is especially useful when combined with Chrome's ability to turn off cert checking for localhost: chrome://flags/#allow-insecure-localhost .

I've tested this in both self-signed and key/cert file mode (by mapping 127.0.0.1 to a domain name I own and have keys for in /etc/hosts), both in hot and non-hot modes.


gigabo:
This is starting to get into the realm of stuff that seems nicer to have in a config file.

Come to think of it most of the options to the CLI seem that way.

This seems good to me as far as https support goes.

aickin:
Added config file for `react-server-cli` in #87. If we merge that one, I'll update this PR to include https in the config file options.

aickin:
This branch got out of sync with master because of the programmatic API changes. Fixed in fee2e26.

gigabo:
@aickin - Looks like this has some conflicts with master again.

aickin:
Thanks. Should be fixed now!

gigabo:
Thanks!

PR #51, opened 2016-03-07
Expose `getRequest()` on the page chain

gigabo:
I want to be able to get at query string parameters without re-parsing.

This seems like the shortest path:

```javascript
import {getCurrentRequestContext} from "react-server";
...
const foo = getCurrentRequestContext().page.getRequest().getQuery().foo;
```


roblg:
:+1: we should have done this a long time ago ;)

PR #52, opened 2016-03-07
Expose some helpers for external cache management

gigabo:
- Request.noCacheReponse()
    - Bypass the cache for the current request.
- Request.getCacheEntry()
    - Get the CacheEntry for the current request.


packages/react-server/core/TritonAgent/Cache.js
```diff
@@ -1,320 +1,10 @@
 
 var logger = require('../logging').getLogger(__LOGGER__)
 ,	Q = require('q')
-,	{ mixin } = require("./util")
-,	isEqual = require("lodash/lang/isEqual")
 ,	isArray = require("lodash/lang/isArray")
+,	CacheEntry = require("./CacheEntry")
 ;
```
gigabo: Below here is all just moved to a dedicated module so `CacheEntry` can be imported into `Request`.

packages/react-server/core/TritonAgent/Request.js
```diff
@@ -101,6 +102,27 @@ Request.prototype.type = function (type) {
 	return this;
 }
```
gigabo: Here are the two new methods.

packages/react-server/core/TritonAgent/CacheEntry.js
```diff
@@ -0,0 +1,317 @@
...
+	whenDataReadyInternal () {
+		return this.dfd.promise;
+	}
+
+	decrementRequesters () {
+		logger.debug("Decrementing: " + this.url);
+		this.requesters -= 1;
+
+		if (this.requesters === 0) {
+			this.cache._removeEntry(this);
```
roblg: for the uncached responses, `this.cache` could be null here on the client-side. (Not that I'd expect anybody to want to do that, since things on the client-side aren't cached anyway...)

packages/react-server/core/TritonAgent/Request.js
```diff
@@ -101,6 +102,27 @@ Request.prototype.type = function (type) {
...
 
+Request.prototype.noCacheResponse = function () {
+	this._noCacheResponse = true;
+	return this;
+}
+
+Request.prototype.getCacheEntry = function() {
+	if(this._noCacheResponse) {
+		return new CacheEntry(
+			null,
```
roblg: why not just pass `this._cache` in here?

gigabo: Because we don't want it to go into the cache!

This is to support out-of-band requests:  I make a request whose response I'm going to stash away and re-use a bunch of times.

When I make the actual request I'll do:

```javascript
TritonAgent.get(url).noCacheResponse().then(stuffIntoCache);
```

Then, later, when I'm _using_ the response I'll do:

```javascript
TritonAgent.get(url).getCacheEntry().setResponse(getFromCache()).requesters++;
```

packages/react-server/core/TritonAgent/Request.js
```diff
@@ -101,6 +102,27 @@ Request.prototype.type = function (type) {
...
+		)
+	} else {
+		return this._cache.entry(
+			this._getCacheAffectingData(),
+			SERVER_SIDE, // createIfMissing
+			this._cacheWhitelist
+		)
+	}
+}
+
```
roblg: It feels a little weird that calling a public `getCacheEntry` will create an entry in the cache. I feel like maybe `getCacheEntry` should always pass `false` for createIfMissing, and we can leave the call to `this._cache.entry(..., SERVER_SIDE, ...)` in `Response.prototype.end`?

packages/react-server/core/TritonAgent/CacheEntry.js
gigabo: Yes, for sure.

If you're caching responses yourself you're responsible for getting them into the `TritonAgent` cache when they need to be.

packages/react-server/core/TritonAgent/Request.js
gigabo: Boy, this PR would be a lot easier to grok if it had the _use_ of these new methods along side of it. :-/

I made [this comment](https://github.com/redfin/react-server/pull/52/files#r55273860) while you were making this one.  Does that clear things up?

roblg:
After some thinking about this, I think I'm uncomfortable with making the TritonAgent Cache's CacheEntry part of its *public* API. Currently, the CacheEntry is basically "package-private". There's a little usage in renderMiddleware, and we depend on the cache serialization format in FragementDataCache.jsx, but everything is internal to react-server, and for the most part it's all in Request.js and Cache.js. I feel like client apps shouldn't be manipulating that cache directly.

I do think supporting external caching would be awesome though. Would it make sense to extend the plugRequest API to be able to affect the response? (We'd probably also want to extend it to allow plugins that aren't globally registered...) I could also imagine pulling some (or part) of `executeRequest` out of `Request.prototype.end`, and providing an extension hook so that a request plugin could monkey-patch a particular request to change the way the data is fetched when `end` is called? Or something along those lines?

Thoughts?

gigabo:
> I'm uncomfortable with making the TritonAgent Cache's CacheEntry part of its public API

You mean the class itself?  It's still not exported in `common.js` and there's no path to it via anything that _is_.

roblg:
> It's still not exported in common.js and there's no path to it via anything that is.
It's accessible via `TritonAgent.get(...).getCacheEntry()`, which I'd consider part of the public API. (or the transitive closure of the public API, or something fancy like that?)

gigabo:
_Instances_ are available via `getCacheEntry()`.

I guess I don't see the harm...  Why make ourselves jump through hoops?

roblg:
> _Instances_ are available via getCacheEntry().
> 
> I guess I don't see the harm... Why make ourselves jump through hoops?

In this case, I feel like _we_ should be the ones jumping through hoops, instead of clients. In order for a client to use this object, they have to understand deeply the lifecycle and internal state of a CacheEntry, which seems not good.

Once we give clients access to that object, we're also tying ourselves to this particular implementation, since now changing basically any of the cache entry could potentially require a semver major version bump (including renaming the `requesters` field, with the proposed usage). Previously it was private, and anyone who is in there mucking with it, and calling methods on it should expect to be broken by any little thing we wanted to do, but if we went with this approach we'd have to expect people to use it. I'd much rather TritonAgent itself was the boundary of things we had to support. (And the fact that TritonAgent exposes its cache publicly via `cache()` is also something that we should probably clean up -- we only have one usage for it internally, and it's a hack. :) )

gigabo:
Okay.

PR #53, opened 2016-03-09
Handle other users of the history nav API

gigabo:
Mark our frames and ignore others.

Sometimes need to clobber other frames, but if users are behaving
correctly this is safe.  Added commentary to this effect.

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -34,6 +34,14 @@ Q.onerror = (err) => {
 	logger.error("Unhandled exception in Q promise", err);
 }
 
+var CURRENT_STATE_FRAME = 0;
+var NEXT_STATE_FRAME = 0;
+
+function pushFrame() {
+	CURRENT_STATE_FRAME = ++NEXT_STATE_FRAME;
+	return CURRENT_STATE_FRAME;
+}
+
 class ClientController extends EventEmitter {
 
 	constructor ({routes}) {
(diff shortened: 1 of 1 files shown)


packages/react-server/core/ClientController.js
```diff
@@ -88,7 +96,23 @@ class ClientController extends EventEmitter {
 			var url = request.getUrl();
 
 			if (type === History.events.PUSHSTATE) {
-				this._history.pushState({frameback:true}, null, url);
+				// Sorry folks.  If we need to do a client
+				// transition, then we're going to clobber
+				// your state.  You must be able to render
+				// from URL, anyway, so if you're set up right
+				// it won't affect user experience.  It means,
```
doug-wade: can't this effect perf, which effects ux?

gigabo: Yes, it affects performance.  If client code behaves correctly, though, it won't affect content or behavior.

roblg:
Code seems sound. Not 100% sure I understand the issue we're resolving though.

gigabo:
> the issue we're resolving

A page wants to use the history navigation API itself.  For example a list page updates its URL when sorting options change.  It can re-sort without re-rendering the whole page.  In pushes onto the history stack when the user changes sort options.  In order to handle re-sorting when the back-button is pressed it needs to listen to the browser's `popstate` event.

We're _also_ listening to `popstate`.  This PR just lets the page handle _its_ `popstate` events (when possible).

Importantly in this scenario the page must _also_ be able to render with a given sort order just given the URL.  This is generally true of all history navigation API usage that modifies the URL.  That means it's safe for us to clobber any of its history frames (which we need to do at client-transition boundaries).  We pay the perf penalty of a re-render when the page _could_ have updated in-place, but the display and behavior remain correct.

roblg:
SGTM :+1:

gigabo:
Also should be noted that the page is responsible for ignoring any `popstate` event with `reactServerFrame` in its state.

Issue #50, opened 2016-03-07
DISCUSSION: Feature ideas for `react-server` & `react-server-cli`


aickin:
(Dunno if this is the right place to put this; please advise if it's not.)

I was thinking about how I could best contribute to making `react-server` a public success, and I wanted to open a little bit of a discussion about what features might be added to (or subtracted from!) `react-server-cli` and `react-server`. I brainstormed for a little while on my own, and the list is pretty scattered and wildly biased towards my own priorities. It's really only meant to be a starting point, and to prevent me from coding on something for a week or two that has no chance whatsoever of being merged because it's not in line with y'all's vision. Caveat: I'm not even sure that these are good ideas if there were infinite time to implement, and they betray my two main priorities: making dev rampup as quick & painless as possible, and having react-server(-cli) be a good example of web performance best practices.

Without any further ado, here's my list, with guesses at T-shirt size:

- **Replace superagent with fetch (L/XL)**: seems like [fetch is rapidly becoming the standard](http://caniuse.com/#feat=fetch), with pretty good polyfills for both client and node. Swapping out superagent for fetch would reduce the amount of documentation required for data loading: rather than sending in a `loader` argument to the page API and then explaining superagent, we'd just say "use fetch like you already do".
- **Port to webpack 2 with tree shaking (M/L)**: This could in theory reduce the size of the initial bundle. Webpack 2 is not out of beta, yet, though, so it's somewhat moot until then.
- **Add support for HTTP/2 with server push (XL/XXL)**: I think one of the most powerful things about `react-server` is that the server-side knows, in a structured way, about all of the associated resources for a page (JS, CSS, JSON XHR) and could push them pro-actively. Not only would this theoretically be faster than the current code, it would also obviate the need for the TritonAgent cache, which would simplify the logic a lot in the HTTP/2 case.
- ~~**Commons chunk (S)**: Tweak the webpack build in react-server-cli to pull out a common chunk of JavaScript, which would include the core of `react-server`. This would allow that chunk to be cached and shared in between pages.~~ (**Edited**: Forgot I already did this, and it was merged in e30947a6dee. Too much context switching!)
- **Long-term caching of static assets (M/L)**: Implement [the webpack config best practices](https://medium.com/@okonetchnikov/long-term-caching-of-static-assets-with-webpack-1ecb139adb95#.m53ry0enh) for setting a far future expiration on static assets.
[...]

Issue #33, opened 2016-02-23
Proposal: Package reshuffle for user friendliness


gigabo:
Right now `react-server` isn't easy to get started with.

It's missing big pieces that are provided by `react-server-cli`.

Proposal:

- Move `react-server` to `react-server-core`.
- Pull the bulk of `react-server-cli` into `react-server`
    - `react-server` exports `startServer.js` (as `.start()` or `.run()`)
    - `react-server-cli` just becomes CLI for calling `require('react-server').start()`.

doug-wade:
I'm struggling to understand the scope of `react-server`.  As far as I can tell, we seem to believe that we are going to manage all of the complexity of the build for every client of react-server, which I think is unrealistic, as it requires new clients that want to do something we haven't considered.  Consider a client that wants to set up a react-server instance with browserify, clojurescript, and support a dev, test, staging, perf testing, and multiple prod environments (say, for instance, they would like to have different build for China vs the US).  As I understand the proposal, that client would need to open a pr to what is you are proposing would be named `react-server` to add browserify support, clojurescript support, jshint for linting, and to change the existing way of handling environments to allow them to use `react-server`.  As I understand it, you'd either fork `react-server` and use that (which would be frustrating when you want to get new features, since you'd have to rebase onto/merge `redfin/react-server`), or you'd use `react-server-core` and lose any other configuration that is exported by the proposed `react-server` solution.  

There's also a question of support -- in our use of react-server, we've had to spend a lot of time debugging webpack errors.  If we hide all the complexity of webpack in the `react-server-cli`, when there _is_ an error (which although I would love to guarantee would never happen, I think realistically will), then rather than using the existing webpack documentation, clients would have to rely solely on the code from `react-server-cli` to figure out the source of their failure, fork the cli, build and deploy with the fork, and then submit the fork upstream.

Also, since we Babelify the code for clients, how would clients opt in or out of language features?  Do they simply have to accept the variant of es6 exported by `react-server`?  Or what if they want a different set of webpack plugins?  Or a different set of eslint settings?  Do we plan on consuming the config files of every possible bundler, transpiler, linter, etc as part of our api?

aickin:
@doug-wade thanks for the thoughtful post, and sorry for my delay in jumping in. Life intervenes sometimes. First a small point:

> Also, since we Babelify the code for clients, how would clients opt in or out of language features? 

[...]

PR #54, opened 2016-03-09
Expose an API to react-server-cli's startServer.

aickin:
Implements a simple export of the startServer API, as proposed in #33 and #50, so that servers can be started up programmatically.

I think it's probably also a good idea to expose the client compilation as an API to allow more flexibility, but I believe that's orthogonal to this PR.


packages/react-server-cli/src/cli.js
```diff
@@ -0,0 +1,44 @@
...
+if (isProduction && !process.env.NODE_ENV) { //eslint-disable-line no-process-env
+	process.env.NODE_ENV = "production"; //eslint-disable-line no-process-env
+}
+const argv = parseCliArgs(isProduction);
+
+// these require calls are after the argument parsing because we want to set NODE_ENV
+// before they get loaded.
+const logging = require("react-server").logging,
+	logger = logging.getLogger(__LOGGER__),
+	start = require(".").start;
```
aickin: Annoying that renaming makes it look like the entire file is new. :(

So you don't have to look at your self: this is one of the two lines changed in this file. I used "." so that it would be sure use the exposed API function.

packages/react-server-cli/src/cli.js
```diff
@@ -0,0 +1,44 @@
...
+	if (!argv.minify) {
+		logger.warning("-- Minification is disabled. Pass --minify, pass --production, or set NODE_ENV=production to turn on.");
+	}
+
+	if (process.env.NODE_ENV !== "production") { //eslint-disable-line no-process-env
+		logger.warning("-- NODE_ENV is not set to \"production\".");
+	}
+}
+
+start(argv.routes, argv);
```
aickin: And this is the other changed line, just changed `startServer` to `start`.

packages/react-server-cli/src/cli.js
roblg: Tricky `.` business.

roblg:
LGTM

packages/react-server-cli/src/cli.js
aickin: Happy to switch to `./index` if you want. Felt slightly DRYer this way but slightly more confusing.

gigabo:
This is cool.

PR #56, opened 2016-03-11
Route logger output to appropriate browser console methods

gigabo:
Helps with filtration in developer tools.

### DIFF
--- a/packages/react-server/core/logging/client.js
+++ b/packages/react-server/core/logging/client.js
@@ -10,23 +10,14 @@
 
 var common = require('./common')
 
-var _console = (typeof console === 'undefined')
-	?{
-		// IE9 doesn't even _define_ console unless the dev tools are open.
-		log   : () => {},
-		error : () => {},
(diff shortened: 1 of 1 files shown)


packages/react-server/core/logging/client.js
```diff
@@ -10,13 +10,35 @@
...
+		debug : Function.prototype.bind.call(console.debug, console),
+		info  : Function.prototype.bind.call(console.info,  console),
+	};
+
+var lvl_map = {
+	emergency : 'error',
+	alert     : 'error',
+	critical  : 'error',
+	error     : 'error',
+	warning   : 'warn',
```
emecell: Does this still only allow `logger.warning` vs. being able to do `logger.warn`?

gigabo: Yep.  This only addresses routing of existing `logger` methods.  It doesn't add any new ones.

roblg:
LGTM

PR #57, opened 2016-03-11
Disable frameback on phones

gigabo:
Having some layout issues that we need to figure out before we re-enable.

This is really too bad.  The ux improvement is most pronounced on phones. :(


gigabo:
Sounds like we're going to be able to disable in a more targeted way on our end.  Closing this.  _phew_

PR #59, opened 2016-03-12
Don't omit roots for empty elements

gigabo:
Some error cases put empty strings into the elements array in
`renderMiddleware`.  Those originally turned into empty roots, but started
erroring out when root element attributes came online.  This restores original
behavior.


roblg:
WHEEE!

PR #62, opened 2016-03-17
Convert to bluebird

gigabo:
:bomb: 

TODO:
- [x] Better port for `Q.allSettled()`

### DIFF
--- a/packages/react-server/core/__tests__/integration/title/AsyncServerTimeoutTitlePage.js
+++ b/packages/react-server/core/__tests__/integration/title/AsyncServerTimeoutTitlePage.js
@@ -1,9 +1,9 @@
-var Q = require("q");
+var Promise = require("bluebird");
 
 class AsyncServerTimeoutTitlePage {
 	getTitle() {
 		// this is a title that should NOT display on the server, as it's more than the server timeout.
-		return Q("An asynchonous timeout title").delay(500);
+		return Promise.resolve("An asynchonous timeout title").delay(500);
 	}
 }
 
--- a/packages/react-server/core/renderMiddleware.js
+++ b/packages/react-server/core/renderMiddleware.js
@@ -7,7 +7,7 @@ var logger = require('./logging').getLogger(__LOGGER__),
 	RequestLocalStorage = require('./util/RequestLocalStorage'),
 	RLS = RequestLocalStorage.getNamespace(),
 	LABString = require('./util/LABString'),
-	Q = require('q'),
+	Promise = require('bluebird'),
 	config = require('./config'),
 	ExpressServerRequest = require("./ExpressServerRequest"),
 	expressState = require('express-state'),
@@ -58,6 +58,12 @@ module.exports = function(server, routes) {
 
 		var start = RLS().startTime = new Date();
 		var startHR = process.hrtime();
+		var resolve, reject;
+
+		var navigatePromise = new Promise((res, rej) => {
+			resolve = res;
+			reject  = rej;
+		});
 
 		logger.debug(`Incoming request for ${req.path}`);
 
@@ -90,7 +96,7 @@ module.exports = function(server, routes) {
 		// setup navigation handler (TODO: should we have a 'once' version?)
 		context.onNavigate( (err, page) => {
 
-			if (!navigateDfd.promise.isPending()) {
+			if (!navigatePromise.isPending()) {
 				logger.error("Finished navigation after FAILSAFE_ROUTER_TIMEOUT", {
 					page: context.page,
 					path: req.path,
@@ -99,7 +105,7 @@ module.exports = function(server, routes) {
 			}
 
 			// Success.
-			navigateDfd.resolve();
+			resolve();
 
 			if (err) {
 				// The page can elect to proceed to render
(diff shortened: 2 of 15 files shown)


packages/react-server/core/util/RequestLocalStorage.js
```diff
@@ -1,5 +1,21 @@
 var RequestLocalStorage = require('request-local-storage');
+var Promise = require('bluebird');
 
-RequestLocalStorage.patch(require('cls-q'));
+RequestLocalStorage.patch(require('cls-bluebird'));
+
+// Unfortunately `cls-bluebird` doesn't go far enough, and the maintainer
+// doesn't seem to be receptive to patches.  There's been a PR open with a
+// similar patch since 2014.  Not going to wait on that.
```
doug-wade: I would include a link (on the off-chance it ever merges) and a quick explanation of why we need/example of where we use `invoke`

packages/react-server/core/util/PageUtil.js
```diff
@@ -183,11 +183,11 @@ function standardizeDebugComments(debugComments) {
 }
 
 function standardizeMetaTags(metaTags) {
-	return PageUtil.makeArray(metaTags).map(metaTag => Q(metaTag));
+	return PageUtil.makeArray(metaTags).map(metaTag => Promise.resolve(metaTag));
```
doug-wade: What's the difference between this and `.map(Promise.resolve)`?

packages/react-server/core/util/RequestLocalStorage.js
gigabo: Thanks.  Added a link in the comment.

I'll also link here, so it shows up in the PR over there: https://github.com/TimBeyer/cls-bluebird/pull/1

packages/react-server/core/renderMiddleware.js
```diff
@@ -894,8 +904,8 @@ function setupLateArrivals(res) {
 	});
 
 	// TODO: maximum-wait-time-exceeded-so-cancel-pending-requests code
-	var promises = notLoaded.map( result => result.entry.dfd.promise );
-	RLS().lateArrivals = Q.allSettled(promises)
+	var promises = notLoaded.map( result => result.entry.promise );
+	RLS().lateArrivals = Promise.all(promises).catch(() => {})
```
doug-wade: should we emit these into the logs or at the error beacon or something?

packages/react-server/core/util/PageUtil.js
gigabo: The binding of `this` within the `resolve` method.

In practice, I expect `Promise.resolve` doesn't use its context so `.map(Promise.resolve)` would be fine.  I'd have to read `Promise.resolve` to be sure, though.

Definitely looks way cleaner that way!

gigabo: Come to think of it I'm already using `Promise.resolve` unbound all over the place.

So I guess it works. :D

gigabo: Oh, there's also the question of what `Promise.resolve` does with extra arguments...  The index will be passed in with `.map(Promise.resolve)`.

packages/react-server/core/__tests__/integration/title/AsyncServerTimeoutTitlePage.js
```diff
@@ -1,10 +1,10 @@
-var Q = require("q");
+var Promise = require("bluebird");
 
 class AsyncServerTimeoutTitlePage {
 	getTitle() {
 		// this is a title that should NOT display on the server, as it's more than the server timeout.
-		return Q("An asynchonous timeout title").delay(500);
+		return Promise.resolve("An asynchonous timeout title").delay(500);
```
doug-wade: As I understand it, we require our clients to use the same promise library that we use internally (which makes sense), which means that they need to take their own dependency on bluebird.  A week ago, I'd have said that's not a problem, since we'll always be on the same promise library, yet here we are.  Would it make sense to export an austensibly custom promise implementation that is just bluebird stripped down to the parts of the api that we actually use and then guarantee that if you use that implementation (as stored in the monorepo), that you'll always be able to upgrade cleanly between react-server versions (even if we change promise implementations under the covers)?  Or to use a peerDependency to communicate that clients need to use bluebird (and which versions we support)?

packages/react-server/core/renderMiddleware.js
gigabo: Yeah, that's a good idea.

Now that I look at the documentation for [`Promise.all`](http://bluebirdjs.com/docs/api/promise.all.html) it looks like this actually isn't equivalent to `Q.allSettled()`, since `Promise.all` _short circuits_ on rejection, whereas `Q.allSettled()` waits for resolution _or rejection_ of all promises.

There were a few `Q.allSettled` calls that I ported this way.  I'll need to come up with a better equivalent.

packages/react-server/core/__tests__/integration/title/AsyncServerTimeoutTitlePage.js
gigabo: Bluebird's cool with other libraries' promises.  If it `.then()`s it's good.

packages/react-server/core/renderMiddleware.js
gigabo: Updated to [a correct port](https://github.com/redfin/react-server/pull/62/commits/c3f1756d9199cba2993e49c92766cea182c9bb65) of `Q.allSettled()` using [`promise.reflect`](http://bluebirdjs.com/docs/api/reflect.html).

Not going to change the error handling behavior here.  Trying to stick to a straight port.  Goal is no change in behavior (aside from performance increase :rocket:).

doug-wade:
lgtm :+1:

roblg:
I'm about to go on vacation, so I trust that y'all have spent more time thinking about this than me. :grin:

:+1:

aickin:
I think the runtime perf differences are great, but I'm wondering if you've looked at how this affects the initial bundle size. I think Bluebird is something like 20k while Q is just 4K (gzip + minify). For a minimal hello world project, react-server is currently about 84k, so this could be a pretty big increase. 

What are your thoughts?

gigabo:
Yeah, that's definitely a tradeoff.

In the context of a real-world site running `react-server` the extra ~16KB is generally going to be a less significant percentage of overall page weight, but as far as weight of `react-server` _itself_ it's pretty big.

As far as run-time benefit: We're seeing a ~10% improvement in server-side response time for one of our highest traffic pages, which is pretty hard to resist.  We also see measurable reductions in GC pause time and `setImmediate` delay.

Aside from the performance benefit we're also taking advantage of the [monitoring events](http://bluebirdjs.com/docs/features.html#promise-monitoring) emitted by `bluebird` to instrument our _usage_ of promises.  We had monkey-patched `Q` to get coarse stats on usage, but it turns out we hadn't done a very good job: We were under-counting our promise-creation by nearly a factor of 10!  `bluebird` gives us finer-grained events via a public API.

Here are some stats from a pair of load tests in isolation (first with `Q`, then `bluebird`):

image

I'm definitely sensitive to increases in library size.  Every byte matters, especially on mobile!  But in my opinion the benefits outweigh the cost in this case.

gigabo:
Not doing this now.

PR #63, opened 2016-03-22
Add babel preset to share babel settings

doug-wade:
Eventually intended to replace react-server/buildutils/gulp-common.js in a more reusable way (for example, with use with a command-line babel).  Should make migration to Babel 6 easier.


gigabo:
Whoa.  Cool.

PR #64, opened 2016-03-23
ClientController bug

chris-voss:
There was a derefencing of state inside of the historyListener when state could be null.  This fixes that.


gigabo:
Thanks @chris-voss!

PR #67, opened 2016-03-28
Only set href on base tag if it existed previously

TonyHYK:
This comes up when we do a client navigation, where it will set the href on the base tag to "undefined".


gigabo:
Nice.  Thanks @TonyHYK.

PR #69, opened 2016-03-28
Route framback pop through the navigator

gigabo:
Need to update the navigator's current request back to outer page so, for
example, `RequestContext.getCurrentPath()` works correctly.

This is a more natural flow through the back-nav.

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -119,6 +119,13 @@ class ClientController extends EventEmitter {
 				this.context.navigator.finishRoute();
 			});
 
+		} else if (this.framebackController.isActive()) {
+
+			// Tell the navigator we got this one, too.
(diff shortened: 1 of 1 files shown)


packages/react-server/core/ClientController.js
```diff
@@ -573,22 +580,18 @@ class ClientController extends EventEmitter {
...
+			if (context) {
+				var path = this._history.getPath();
+
+				// Pass in "popstate" because this is
+				// when a user clicks the forward/back
+				// button.
+				context.navigate(
+					new ClientRequest(path, {frameback}),
+					History.events.POPSTATE
+				);
```
gigabo: This diff chunk looks bigger than it is because of the indentation change.  It's just replacing the `if/else` with what was previously the body of the `else` branch (eliminating the `if` case).

doug-wade:
lgtm :+1:

PR #70, opened 2016-03-31
Dont blow away the dom on client transition

doug-wade:
When doing client transitions, we can reuse the dom by letting React diff the shadow dom as it would with component re-renders, which reduces client-side render time.  This adds an option to ClientRequest that reuses the dom on client transitions.  Example use:

```
require('react-server')
  .getCurrentRequestContext()
  .navigate(new ClientRequest('http://example.com/client-transition', { reuseDom: true }, History.events.PUSHSTATE);
```


gigabo:
Moved to #78.

PR #72, opened 2016-04-04
Update master title right before frameback navigation

TonyHYK:
Get the newest document title before frameback happens. This is to pick up any potential changes to the title made by the page. I kept the `masterTitle` initialization in the constructor as fallback, since there should never be a case where we expect `this.masterTitle` to return undefined.

### DIFF
--- a/packages/react-server/core/FramebackController.js
+++ b/packages/react-server/core/FramebackController.js
@@ -92,6 +92,8 @@ class FramebackController extends EventEmitter {
 		logger.debug(`Navigating to ${url}`);
 
 		this.active = true;
+		// update master title in case the title has changed since initialization
+		this.masterTitle = document.title;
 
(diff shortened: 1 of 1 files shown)


gigabo:
Hi @TonyHYK, what's a use case for this?

TonyHYK:
An example on the map page is...

1. Land on Berkeley map page: notice that the title is "Berkeley, CA Real Estate...etc"
2. Search for San Francisco and the document title will change to San Francisco Real Estate
3. Now go to the LDP and back to trigger frameback.

If we don't update `masterTitle`, the document title will revert to Berkeley while the map is pointing to San Francisco.

gigabo:
Oh, I see.  The `masterTitle` was only being set in the constructor.  nice fix!

PR #74, opened 2016-04-05
Preload data bundles for snappy client transitions

gigabo:
Adds a method to preload data for a URL that might be navigated to via client transition.

```javascript
ReactServerAgent.preloadDataForUrl(url);
```

The data is then cached for the lifetime of the client controller.  If a client transition is made to `url` then the `ReactServerAgent` cache is pre-populated.

A new page lifecycle was added to provide bundles of preloaded data.  It just runs `handleRoute`, waits for data to resolve, and then spits out the RSA cache as a JSON blob.  It's triggered by a query-string parameter.

TODO: Invalidation.  Probably need a method to blow away the preload cache.  Maybe also one to drop data for individual URLs.

### DIFF
--- a/packages/react-server/core/ReactServerAgent.js
+++ b/packages/react-server/core/ReactServerAgent.js
@@ -1,4 +1,5 @@
 var RLS = require('./util/RequestLocalStorage').getNamespace()
+,	Q = require("q")
 ,	Cache = require("./ReactServerAgent/Cache")
 ,	Request = require("./ReactServerAgent/Request")
 ,	Plugins = require("./ReactServerAgent/Plugins")
@@ -10,9 +11,13 @@ function makeRequest (method, url) {
 	return new Request(method, url, API.cache());
 }
 
-const PRELOADS = {};
+const BUNDLE_CACHE     = {};
+const BUNDLE_PARAMETER = '_react_server_preload_bundle';
+const BUNDLE_OPTS      = {[BUNDLE_PARAMETER]: 1};
 
 var API = {
+
+	BUNDLE_PARAMETER,
 
 	get (url, data) {
 		var req = makeRequest('GET', url);
@@ -100,15 +105,21 @@ var API = {
 	},
 
 	preloadDataForURL (url) {
-		if (PRELOADS[url]) return PRELOADS[url];
-		return this.get(url, {_react_server_preload_bundle: 1})
-			.then(data => PRELOADS[url] = data.body); // eslint-disable-line no-return-assign
+		if (SERVER_SIDE) throw new Error("Can't preload server-side");
+		if (!BUNDLE_CACHE[url]){
+			BUNDLE_CACHE[url] = this.get(url, BUNDLE_OPTS)
+				.then(data => data.body);
+
+		}
+		return BUNDLE_CACHE[url];
 	},
 
-	_checkForPreload (url) {
-		if (PRELOADS[url]) {
-			API.cache().rehydrate(PRELOADS[url]);
-		}
+	_rehydratePreloadData(url) {
+		// If we don't have any then we can't use it.
+		if (!BUNDLE_CACHE[url]) return Q();
+
+		return BUNDLE_CACHE[url]
+			.then(data => API.cache().rehydrate(data));
 	},
 
 }
--- a/packages/react-server/core/ReactServerAgent.js
+++ b/packages/react-server/core/ReactServerAgent.js
@@ -107,11 +107,13 @@ var API = {
 	preloadDataForURL (url) {
 		if (SERVER_SIDE) throw new Error("Can't preload server-side");
 		if (!BUNDLE_CACHE[url]){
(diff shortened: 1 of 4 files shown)


packages/react-server/core/ReactServerAgent.js
```diff
@@ -10,8 +11,14 @@ function makeRequest (method, url) {
 	return new Request(method, url, API.cache());
 }
 
+const BUNDLE_CACHE     = {};
```
aickin: Very nitpicky thought: we currently use "bundle" to talk about the static JS bundles, so maybe its worth using a different term here, or a more specific one like `DATA_BUNDLE_CACHE`. 

As always, feel free to ignore.

gigabo: Thanks @aickin, I've pushed a patch that uses `DATA_BUNDLE` to disambiguate.

aickin:
This is an absurdly elegant way to deal with bundling and preloading data. @roblg mentioned the bundling to me the other night and I was blown away. Bravo.

One thought, perhaps parallel to this PR: I wonder if it's also worthwhile to write some code that allows the pre-loading of the JavaScript and CSS bundles, which are the other potentially significant drivers of latency. It seems to that this solution will be awesome when going from one page of a particular type to another of the same type (like LDP --> LDP), but less so when switching types. I could imagine it just using the [built-in browser `prefetch`](http://caniuse.com/#feat=link-rel-prefetch), which currently works in about 2/3rds of browsers. Thoughts?

gigabo:
> pre-loading of the JavaScript and CSS bundles

Ooh, yeah! 🚀 

Let's see... For the JS bundle webpack will fetch it for us (and get it all prepped) if we call the loader in the routes object.

For CSS, unfortunately, I'm not sure if we can cheaply preload since the _URLs_ are returned from `getHeadStylesheets` and it's not safe to call page methods out of order.

Good idea!

aickin:
Hmmm... good point. You don't want to call `handleRoute` on the client; that kind of defeats the entire purpose.

Here's a wacky idea: could you call `getHeadStylesheets` and `getScripts` (or whatever it's called) on the server side and pass the JS/CSS URLs down in the JSON with the data? Then you could add `<link rel="prefetch">` for each resource immediately, with the hope that the browser will download them before the user navigates. This would also help in the situation where a needed JS file isn't part of the page bundle (e.g. third party scripts), which the webpack loader will not be able to prefetch.

One downside with this idea vs. your current solution is that you'd need to parse the JSON from the server when it's downloaded, rather than when it's needed, which I really like about the current solution. If that was really a problem in practice, you could conceivably make some easier-to-parse format.

gigabo:
@aickin - That's perfect!

I'm going to freeze the scope of _this_ PR to try to get some 👍 s and get it merged.  Filed #77 to capture JS and CSS asset prefetch as the next step for client transition performance improvement.

aickin:
Sounds great. Thanks! 

And fwiw, this PR lgtm (am I even allowed to say that?).

bartkusa:
As an aside, <link rel="prefetch"> is/was kinda bad in Chrome. Last time I looked (a year ago?), it took priority over _the current page's_ requests, like images.

Just check Chromium's bug tracker before forging ahead.

aickin:
Do you have a citation for when that was true? At least according to [this presentation by Ilya Grigorik](https://docs.google.com/presentation/d/18zlAdKAxnc51y_kj-6sWLmnjl6TLnaru_WH0LJTjP-o/present?slide=id.g120f70e9a_0140), at least, prefetch hints "have lowest possible priority".

packages/react-server/core/ReactServerAgent.js
```diff
@@ -10,8 +11,14 @@ function makeRequest (method, url) {
 	return new Request(method, url, API.cache());
 }
 
+const DATA_BUNDLE_CACHE     = {};
```
roblg: Still trying to read through this PR, but this looks like a module-level variable? Does it have request-specific stuff in it?

roblg: Oh, I see. Can't preload server-side. Still, this variable as a global gives me the heebie-jeebies. Can we add a comment here to that effect as well? I literally almost had a heart attack....

gigabo: Heh.  Good call.  Pushed a patch that both adds a comment and also just makes `DATA_BUNDLE_CACHE` undefined on the server.

bartkusa:
https://code.google.com/p/chromium/issues/detail?id=61476

roblg:
LGTM :+1: 

And FWIW, @aickin you have my vote to be allowed to say "lgtm" ;)

gigabo:
> And FWIW, @aickin you have my vote to be allowed to say "lgtm" ;)

Agreed.  Thanks for reviewing!

aickin:
@bartkusa Thanks for the warning and the cite!

@gigabo & @roblg Aw shucks, thanks!

PR #75, opened 2016-04-06
Added the ClientController instance to the window for an all access pass

RobKopp:
Added the ClientController instance to the window rather than cherry picking out the late arrival functions for easier access in applications.


gigabo:
This is awesome!  Thanks @RobKopp!

PR #76, opened 2016-04-06
Reloading a page causes the browser history to be modified, and the C…

jmhwang7:
…lientController cannot differentiate between frames of different sessions causing it to not navigate in some cases. This fix by Luke M. fixes that.

(cherry picked from commit c635c5539e02b86edb366d3b10adeadf446b6240)
Signed-off-by: Jennifer Hwang <jennifer.hwang@redfin.com>


packages/react-server/core/ClientController.js
```diff
@@ -34,12 +34,13 @@ Q.onerror = (err) => {
 	logger.error("Unhandled exception in Q promise", err);
 }
 
-var CURRENT_STATE_FRAME = 0;
+var SESSION_START_PREFIX = (new Date()).getTime() + '_';
 var NEXT_STATE_FRAME = 0;
+var CURRENT_STATE_FRAME = SESSION_START_PREFIX + NEXT_STATE_FRAME;
```
gigabo: I don't like having two places where frame identifiers are created (inline here and in `pushFrame`).

The _initial_ frame isn't used (new frames are always obtained via `pushFrame()`), so this could just be declared without assignment (left `undefined`) or made explicitly `null`:

```javascript
var CURRENT_STATE_FRAME; // No frame until `pushFrame()` is called.
```

Could also initialize with a valid (though unused) identifier while still avoiding duplication by just calling `pushFrame()` and burning the first frame:

```javascript
var CURRENT_STATE_FRAME = pushFrame();
```

`pushFrame` gets hoisted and the enclosed variables are all defined before `CURRENT_STATE_FRAME` so this wouldn't require re-ordering anything.  Kind of weird to always burn a frame identifier, though.

gigabo:
This is a nice fix.  Please get the indentation cleaned up for `eslint` and do something about the [duplicate construction of frame identifiers](https://github.com/redfin/react-server/pull/76#discussion_r58753121) and let's get this merged in!

gigabo:
Thanks @jmhwang7!

PR #78, opened 2016-04-06
Client transition dom

gigabo:
Taking over responsibility for getting @doug-wade's awesome DOM preservation changes shepherded into master, since he's getting pulled in three different directions in his day job.

This is #70 with a few patches to get things working in some cases that didn't come up during the hackathon that spawned it.

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -523,13 +523,14 @@ class ClientController extends EventEmitter {
 
 			[].slice.call(
 				document.querySelectorAll(`div[${PAGE_CONTAINER_NODE_ID}]`)
-			).forEach(root => {
-
-				// Gotta get rid of our containers, too.
-				// Need to do this _after_ killing the
-				// elements, since they might live within
-				// these containers.
-				root.parentNode.removeChild(root);
+			).forEach((root, i) => {
+				if (i >= index) {
+					// Gotta get rid of our containers,
+					// too.  Need to do this _after_
+					// killing the elements, since they
+					// might live within these containers.
+					root.parentNode.removeChild(root);
+				}
 			});
 		}
 	}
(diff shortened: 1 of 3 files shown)


gigabo:
BTW - As mentioned in https://github.com/doug-wade/react-server/pull/2: Still need to deal with root container/element attribute differences, but probably should fix redfin#73 first.

gigabo:
Added some commentary to explain the DOM re-use strategy and hopefully help ease review here.

packages/react-server/core/ClientController.js
```diff
@@ -391,31 +391,63 @@ class ClientController extends EventEmitter {
...
+				// we can re-use we'll render into them.
+				//
+				// DOM re-use is currently opt-in.
+				//
+				if (this._reuseDom) {
+					var oldRootElement = document.querySelector(
+						`div[${REACT_SERVER_DATA_ATTRIBUTE}="${index}"]`
+					);
+					var oldRootContainer = document.querySelector(
+						`div[${PAGE_CONTAINER_NODE_ID}="${index}"]`
```
roblg: Code style nit: defining these `vars` inside the if-statement with intent to use them outside. Works in JS, but is not what I would call expected. :)

packages/react-server/core/ClientController.js
```diff
@@ -391,31 +391,63 @@ class ClientController extends EventEmitter {
...
 
-					// If we're closing a container its
-					// parent is once again our mountNode.
+				// The current strategy for re-use is: So long
+				// as the _shape_ of the root structure is the
+				// same, we'll re-use.  Once the new page's
+				// shape diverges, we'll blow away the
+				// remaining elements left over from the
+				// previous page and create everything for the
+				// new page as we go.
```
roblg: This comment is super-helpful. :+1:

roblg:
LGTM :+1:

packages/react-server/core/ClientController.js
gigabo: Pulled the declarations out next to the one for `mountNode`.

PR #79, opened 2016-04-07
Test pages

gigabo:
This is a place for pages that let us play around with `react-server` features.  It's not necessarily for _automated_ tests (though we could build some around it), it's for _interactive_ testing during development.

I'm not sure if I'm using `react-server-cli` in the best way.  I'm doing an external gulp build and then running `start` on the _target_ JS.  Would need to have the server running _next to_ a gulp `watch` to get hot-reloading, I guess?

I'm running it thusly:

```bash
$ gulp build && node index.js
```

It's tiny, so that's fast, but it's not really taking full advantage of the framework I think.


gigabo:
Oh, just realized I need a `README`.  Will update probably in the morning.

gigabo:
Unless there are _objections_ (bad name? bad idea?) I'm just going to merge this later this morning, since I have actual test _pages_ I want to build to work on new features.

Still interested in ideas for better build/run workflow here.

roblg:
What's the difference between "test pages" and "examples of react-server features" in this case?

gigabo:
Not much, I guess?  The pages I'm thinking of for working on features aren't showcase-y things.  They're just simple pages that exercise certain `react-server` features.

roblg:
I don't feel that strongly either way, but since I expect these would actually be reasonable examples to link to from documentation (since they'll be simple, and exercise a fairly limited feature set), it might make sense to advertise them that way. We've talked about actually pulling out the page tests into a separate project, and having "page tests" and "test pages" seems likely to cause confusion. :)

gigabo:
`react-server-examples` is available.  Like that better?

aickin:
Is the idea that it's all in one big server project, or is it possible that there would be multiple projects?

Asking because I'm interested in porting my [react-server-examples](https://github.com/aickin/react-server-examples) over.

packages/react-server-test-pages/gulpfile.babel.js
```diff
@@ -0,0 +1,22 @@
...
+const babel = require("gulp-babel");
+const logging = require("react-server-gulp-module-tagger");
+
+gulp.task("travis-ci", ["build", "eslint"]);
+
+gulp.task("build", () => gulp.src("src/**/*.js")
+	.pipe(logging())
+	.pipe(babel())
+	.pipe(gulp.dest("lib/"))
+);
```
aickin: So, it's worth noting that IIRC that `react-server-cli` [will do babel transformation by itself](https://github.com/redfin/react-server/blob/master/packages/react-server-cli/bin/react-server-cli), so this build step is (I think?) superfluous. But maybe you want this for consistency with the other packages, so I guess it's ok.

The upside for allowing `react-server-cli` to do the babel compile is that I think you could just add an entry for `scripts.start` that was `react-server-cli`, and then the install instructions would just be `npm start`. Also, watching/hot loading works by default. Thoughts?

gigabo:
> Asking because I'm interested in porting my react-server-examples over.

Oh, sweet!  Those are definitely better showcase examples than what I'm imagining here.  I think they make sense to maintain as separate packages.

We've historically built test pages on our internal `react-server` consumer, but I'd like to start putting that sort of thing out where:

- It's useful to _other_ `react-server` users.
- It takes advantage of newer `react-server` features (cli).
- It builds faster (our internal build is... not thin).

But these aren't the sort of pages where I'd say "go here to learn how to do the right thing with `react-server`".  They're more: "go here if you want to test out how your changes to `react-server` render scheduling affect the feel of client transitions".

packages/react-server-test-pages/gulpfile.babel.js
gigabo: Okay, got it.  I had this thing totally inside out.  Patch inbound that uses `react-server-cli` as intended.

packages/react-server-cli/README.md
```diff
@@ -115,6 +115,12 @@ Sets the severity level for the logs being reported. Values are, in ascending or
 
 Default is **'debug'** in development mode and **'notice'** in production.
 
+#### --global-react
+
+Expose `React` as a global variable so modules that use JSX don't have to import it (without obviously using it).
+
+Defaults to **false**.
+
```
gigabo: @aickin - This is something that the old `index.js` in this package provided that `react-server-cli` didn't.

Does this new option look okay to you?

aickin: It's all right, though not my 100% favorite, for a few reasons:

- Just a general aversion to magic globals when a module system exists. Also, does this put a React global in both server and client global scope?
- Components that are written assuming a global React will not work in environments that use standard module import, which then forces more apps to add React to global scope if they want to share that code. In this way, it infects the ecosystem and causes compatibility problems.
- It maybe has the potential to confuse packagers and other new dev tools that depend on ES2015 imports to understand module dependency graphs.
- If the issue is that you don't want to have to type `import React from 'react'` at the top of component files with JSX and no reference to React, there is [a simple babel transform](https://www.npmjs.com/package/babel-plugin-react-require) that will do that for you. The output of your babel build will then be compatible with projects that don't use global React and will have parseable dependency trees.

All that said, would I really mind terribly? No, not really. Your call!

gigabo: > a simple babel transform

Thaaaaat's better.  Thanks @aickin!

gigabo:
Okay, gonna merge this.  I've got another branch with feature work that's based on it.

If you've got lingering objections please let me know!

PR #80, opened 2016-04-07
Log a couple of types of errors previously swallowed by the cli

gigabo:
- Errors starting a server were re-thrown in a native promise `.then`
		rejection handler, which just caused the `.then` to return a rejected
		promise that went into the void.  Now logging via `logger.error`.
- Errors loading a page were similarly lost.  Added a try/catch with a
		`console.error`.


aickin:
LGTM. Sorry about my sloppiness. :(

gigabo:
I'm actually going to hold off on this until #83 lands so you don't get stuck with merge conflicts again. ;)

PR #82, opened 2016-04-07
Root when

gigabo:
The `<RootElement listen={...}>` is really concise with gated emitters, but for general stores it's easier to have the gate and the emitter separate.  It's also nice to be able to have two gates even with gated emitters.

So, this adds `<RootElement when={...}>`.  The promise can resolve with a props object, but it doesn't _need_ to.  Render-time props are the union of `listen` emitter props and `when` promise props, with the promise winning collisions.

### DIFF
--- a/packages/react-server/core/components/RootElement.js
+++ b/packages/react-server/core/components/RootElement.js
@@ -73,6 +73,7 @@ module.exports = RootElement;
 
 RootElement.propTypes = {
 	listen: React.PropTypes.func,
+	when: React.PropTypes.object, // A promise.
 	childProps: React.PropTypes.object,
 	_isRootElement: React.PropTypes.bool,
 }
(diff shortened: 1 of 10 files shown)


gigabo:
This also fleshes out the test pages packages a little bit.  Might make the goal there a little clearer.

packages/react-server/core/components/RootElement.js
```diff
@@ -147,3 +148,26 @@ RootElement.scheduleRender = function(element) {
...
+	// It may be a gated emitter, but we want to make sure we squeeze
+	// props out of it from the beginning if it's not.
+	// Finally gate on the `when`.
+	return Q(listen && RootElement.installListener(element, listen))
+		.then(el => el && (rendered = el))
+		.then(() => when)
+		.then(childProps => childProps
+			?React.cloneElement(rendered, {childProps})
+			:rendered
+		)
```
roblg: I'm unsure of the complete flow here, but it seems like if `listen` isn't specified on props, `rendered` will never be set, which seems like it means the promise would resolve w/ `undefined`?

gigabo: `rendered` is initialized to `element` upon declaration.

roblg: :facepalm:

roblg:
LGTM. Not that that matters that much, because I'm obviously incapable of reading today.

gigabo:
Thanks for reviewing.  I've been working on a change set based on this, so I'll be glad to get it merged. :)

gigabo:
BTW - I'd also like to start using https://github.com/redfin/react-server/pull/78, so please have a look there if you get a chance.

PR #83, opened 2016-04-08
Update integration tests to use react-server-cli

aickin:
This PR pulls all of the integration tests out of `react-server` and into their own directory, and it makes all the integration tests use `react-server-cli` to start and run the servers. It allows us to delete a bunch of duplicative code, and it makes sure that `react-server-cli` gets exercised in the test suite.

There are three main pieces to this PR:

1. The moving of integration tests to `react-server-integration-tests`. This involved a lot of file/directory moving and changing of import paths and relative paths. (in a21ed27)
1. Adding the ability to stop a server started by `react-server-cli`. When I added the ability to programmatically start a server, it didn't really occur to me to add the ability to stop one. :facepalm: This meant changing the return value of startServer pretty significantly. (in e90cd1e)
1. Replacing the guts of `testHelper` with `react-server-cli`. (also in e90cd1e)


aickin:
I see the build failed; could be because I'm using node 4/npm2. I'll look at it when I get back to desktop.

packages/react-server-cli/src/startServer.js
```diff
@@ -31,69 +35,113 @@ export default (routesRelativePath, {
...
+				server.use('/', compression(), express.static(`__clientTemp/build`));
+				logger.info("Starting static JavaScript server...");
 
-			http.createServer(server).listen(port, () => {
-				logger.info(`Started static JavaScript server on port ${port}`);
-				resolve();
+				httpServer.on('error', (e) => {
+					console.error("Error starting up JS server");
+					console.error(e);
+					reject(e)
```
gigabo: This looks like more robust handling of server start failure than https://github.com/redfin/react-server/pull/80.

packages/react-server-integration-tests/src/__tests__/integration/containers/pages/SingleRootContainerInArray.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import RootContainer from "../../../../components/RootContainer";
+import {RootContainer} from "react-server";
```
gigabo: That's nice. :)

gigabo:
This is cool.  Next step in my mind is pulling `startServer` out of `react-server-cli` and into something that the CLI wraps.  These tests don't actually use the CLI. ;)

aickin:
I've delved into the build a bit, and it looks like the way I set up the test package is not to lerna's liking; seems like lerna can't correctly handle projects with peer dependencies. Dunno exactly what to do, other than to pull `react-server-integration-tests` out of the `packages` directory entirely, hoping that lerna will leave it alone. That means then I have to write a gulp script to npm install the test dir, though, which is annoying, but I don't have any other good options.

aickin:
Hooray, my scheme in 4fc2b67 of moving the integration tests out of `packages` pleased the build gods!

In summary, if you're interested: at first I had the integration tests under `packages/react-server-integration-tests` and in the package.json had lines like this:

```javascript
dependencies: {
  <snip>
  "react-server": "file:../react-server",
  "react-server-cli": "file:../react-server-cli"
}
```

This worked fine for testing locally, but lerna was not a fan, and `lerna bootstrap` errored out with this format in the package.json. I then tried to use normal dependencies and have lerna figure it out, but this failed because `react-server` is actually a peer dependency of `react-server-cli`, and lerna doesn't handle those kinds of dependencies correctly.

Eventually, I moved the integration tests out of `packages` altogether (which makes a certain amount of sense, since they aren't a package we ever intend to publish). This let me use the `file:` style local references in the package.json. I had to add a few extra gulp tasks to the root gulp file, one to npm install the integration test directory, and one to run the integration tests, but now it seems to work.

aickin:
As an added bonus, in cdf17e3 I think I finally squashed the intermittent test failures I was seeing; looks like they were due to the module cache being poisoned, and my strategy of manually deleting from the cache was not working consistently.

aickin:
And back to the merge conflict mines.

Siri, remind me to not commit three simultaneous conflicting branches next time, ok?

aickin:
Can I gently bump this PR? I'm starting to accumulate things behind it.

packages/react-server-cli/src/compileClient.js
```diff
@@ -194,8 +195,12 @@ module.exports = {
 	}
 };`);
 
-	const routesFilePath = `${workingDirAbsolute}/routes_${isClient ? "client" : "server"}.js`;
-	fs.writeFileSync(routesFilePath, routesOutput.join(""));
+	const routesContent = routesOutput.join("");
+	// make a unique file name so that when it is required, there are no collisions
+	// in the module loader between different invocations.
+	const routesCRC = crypto.createHash('md5').update(routesContent).digest("hex");
```
roblg: `routesCRC` --> `routesMD5`?

aickin: Yikes. An earlier version was CRC; I'll change it. Thanks.

packages/react-server-cli/src/startServer.js
```diff
@@ -160,40 +195,72 @@ const startHotLoadJsServer = (compiler, port, httpsOptions) => {
...
 };
 
+// takes in the err and stats object returned by a webpack compilation and returns
+// an error object if something serious happened, or null if things are ok.
 const handleCompilationErrors = (err, stats) => {
 	if(err) {
 		logger.error("Error during webpack build.");
 		logger.error(err);
-		throw new Error(err);
+		return new Error(err);
```
roblg: do `Error` objects get a stack in JS if you return them as opposed to throwing them?

roblg:
LGTM. I like the simplification+reuse for test startup.

packages/react-server-cli/src/startServer.js
aickin: Helpfully, it's completely non-standard! From [the MDN doc on Error.prototype.stack](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack):

> Different browsers set this value at different times. For example, Firefox sets it when creating an Error object, while PhantomJS sets it only when throwing the Error, and MSDN docs also seem to match the PhantomJS implementation.

Given that the method that calls this method doesn't throw the error (because it uses it for a Promise `reject()`); I think the question is moot? (Although admittedly, promise error handling consistently makes me feel stupid, so I may be thinking about this wrong.)

roblg: In the past when using a promise library, we just throw, and it rejects automatically. It seems like that's probably not true w/ native promises. In any event, I'm OK with trying out this pattern and changing it if it doesn't work for some reason.

gulpfile.js
```diff
@@ -13,3 +14,16 @@ gulp.task('travis-ci', ['test', 'eslint']);
...
+// to call npm install in that dir as part of the build.
+gulp.task('install-integration-test', () => gulp.src(['./integration-tests/package.json'])
+  .pipe(install())
+);
+
+// run the integration tests.
+gulp.task('integration-test', ['install-integration-test'], () => gulp
+	.src(['./integration-tests/gulpfile.babel.js'], {read: false})
+	.pipe(chug({tasks: ['test']}))
+);
```
gigabo: This looks good.  I have a feeling we'll want to generalize it pretty soon.  We have other packages that don't really make sense to publish (e.g. `react-server-test-pages`).

gigabo:
Nice, thanks @aickin!

PR #84, opened 2016-04-08
Link bundle and reuse

gigabo:
Added two new optional props to `<Link>` elements (and the `ClientRequest` constructor):

- `reuseDom`: Don't blow away the DOM on client transition.
- `bundleData`: Fetch the data as a bundle (instead of request-by-request).

The `bundleData` option currently also _caches_ the bundle forever.  Still need to deal with invalidation.

This PR also adds a navigation playground for testing out the various forms of navigation in `react-server`.  Pretty snappy with bundling and DOM reuse. :)

Frameback needs some love.  Gonna tackle that next.

### DIFF
--- a/packages/react-server-test-pages/middleware/RequestToPort.js
+++ b/packages/react-server-test-pages/middleware/RequestToPort.js
@@ -1,10 +1,12 @@
 import {ReactServerAgent} from "react-server";
 export default class RequestToPortMiddleware {
 	handleRoute(next){
-		ReactServerAgent.plugRequest(req => {
-			// TODO: Get port dynamically?
-			req.urlPrefix('localhost:3000');
-		});
+		if (typeof window === "undefined"){
+			ReactServerAgent.plugRequest(req => {
+				// TODO: Get port dynamically?
+				req.urlPrefix('localhost:3000');
+			});
+		}
 		return next();
 	}
 }
--- a/packages/react-server-test-pages/pages/data/delay.js
+++ b/packages/react-server-test-pages/pages/data/delay.js
@@ -7,10 +7,9 @@ export default class DelayDataPage {
 	getContentType() {
 		return 'application/json';
 	}
-	handleRoute(next) {
-		return Q.delay(this.getRequest().getQuery().ms||0).then(next);
-	}
 	getResponseData() {
-		return JSON.stringify({'ok': true});
+		const { ms, val } = this.getRequest().getQuery();
+		return Q.delay(ms||0)
+			.then(() => val||JSON.stringify({'ok':true}));
 	}
 }
--- a/packages/react-server-test-pages/pages/navigation/playground.js
+++ b/packages/react-server-test-pages/pages/navigation/playground.js
@@ -34,6 +34,8 @@ const ClientTransitionLink = ({row}) => <Link path={`${BASE}?page=${row}`}>Clien
 
 const FramebackLink = ({row}) => <Link frameback={true} path={`${BASE}?page=${row}`}>Frameback</Link>
 
+const ReuseDom = ({row}) => <Link reuseDom={true} path={`${BASE}?page=${row}`}>Reuse DOM</Link>
+
 class ClientRenderIndicator extends React.Component {
 	constructor(props){
 		super(props);
@@ -64,7 +66,8 @@ export default class NavigationPlaygroundPage {
 			<ClientRenderIndicator />
 			<NormalLink />
 			<ClientTransitionLink />
-			<FramebackLink/>
+			<FramebackLink />
+			<ReuseDom />
 		</RootContainer>);
 	}
 }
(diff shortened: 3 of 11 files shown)


packages/react-server-test-pages/middleware/RequestToPort.js
```diff
@@ -1,10 +1,12 @@
...
-		ReactServerAgent.plugRequest(req => {
-			// TODO: Get port dynamically?
-			req.urlPrefix('localhost:3000');
-		});
+		if (typeof window === "undefined"){
+			ReactServerAgent.plugRequest(req => {
+				// TODO: Get port dynamically?
+				req.urlPrefix('localhost:3000');
+			});
+		}
```
gigabo: This just disables the prefix in the browser since it causes cross origin errors.

packages/react-server-test-pages/pages/data/delay.js
```diff
@@ -7,10 +7,9 @@ export default class DelayDataPage {
...
 		return 'application/json';
 	}
-	handleRoute(next) {
-		return Q.delay(this.getRequest().getQuery().ms||0).then(next);
-	}
 	getResponseData() {
-		return JSON.stringify({'ok': true});
+		const { ms, val } = this.getRequest().getQuery();
+		return Q.delay(ms||0)
+			.then(() => val||JSON.stringify({'ok':true}));
```
gigabo: Added a `val` parameter to `/data/delay`.  It just gets echoed back after `ms` milliseconds.

packages/react-server-test-pages/pages/navigation/playground.js
```diff
@@ -0,0 +1,89 @@
...
+			<ClientRenderIndicator />
+			<NormalLink />
+			<ClientTransitionLink />
+			<ReuseDom />
+			<BundleData />
+			<BundleDataAndReuseDom />
+			<FramebackLink />
+		</RootContainer>);
+	}
+}
```
gigabo: Page looks like this:
image

Those check-marks are hour-glasses until the client render happens (between when they're visible and when they're interactive).  They revert to hour-glasses on `navigateStart` and switch back on re-render.  Nice way to see the flow of render even when the DOM is preserved.

packages/react-server/core/ClientController.js
```diff
@@ -457,9 +457,10 @@ class ClientController extends EventEmitter {
 					}
 				}
 
-				if (element.containerOpen || element.containerClose){
-					return; // Nothing left to do.
-				}
+			}
+
+			if (element.containerOpen || element.containerClose){
+				return; // Nothing left to do.
```
gigabo: I think this broke with DOM preservation, though I'm not sure.  Don't want to hand container control objects off to React.

gigabo: And, to be clear, this just eliminates log noise.

packages/react-server/core/components/Link.jsx
```diff
@@ -8,13 +8,17 @@ module.exports = React.createClass({
...
+		path       : React.PropTypes.string.isRequired,
+		bundleData : React.PropTypes.bool,
+		frameback  : React.PropTypes.bool,
+		reuseDom   : React.PropTypes.bool,
 	},
 
 	getDefaultProps(){
 		return {
-			frameback: false,
+			bunndleData : false,
```
aickin: I think this is misspelled (`bunndleData` vs. `bundleData`).

aickin:
Bundling data is pure :fire:. My questions:

1. Why is false the default? I would have assumed that true would be the default given current browser performance characteristics with lots of HTTP requests, but I have not thought about it deeply.
1. Why is this an attribute of a link and not of the page that you are transitioning to? It seems to me that bundling is more likely to always be good for a particular page class or not, versus desirable on some links to that page and not on others.
1. Have you thought at all about what will happen in an HTTP/2 world? I'm guessing this is probably HTTP/1-only given the perf characteristics of HTTP/2, but I'm not entirely sure.

gigabo:
> Bundling data is pure :fire:.

More to come! 🚀 

> questions

1. A data bundle is only as fast as the slowest contributing upstream request.  If the upstream requests necessary to render content above the fold are much faster than that long pole it's probably better not to use a data bundle.
2. That's a good point.  A page that makes many upstream requests, none of which are very slow, will generally benefit from data bundling.  I like having control in the linking page because it makes it easy to enable/disable bundling at run time and monitor the impact.  It may also turn out that some pages benefit from bundling in some contexts but not others (e.g. mobile vs desktop).  Maybe a page could override the default of `false` via `setConfigValues()`, but a request has ultimate authority?
3. Yeah, I guess in theory this should be a non-issue with HTTP/2.  We'll see. :)

packages/react-server/core/components/Link.jsx
gigabo: Nice catch!

gigabo:
Ticket to set default bundling policy via page config: https://github.com/redfin/react-server/issues/92.

Anybody feel ready to give this a 👍 ?

gigabo:
Hi guys.  I've got a big batch of frameback improvements nearly ready to drop that's based on this branch.  If you get a chance, please give this a look over and 👍 / 👎 .  Thanks!

aickin:
I'm happy to give a :+1:, given that you filed #92. Cheers.

gigabo:
Sweet.  Thanks @aickin!

roblg:
Seems fine to me.

PR #85, opened 2016-04-09
Blank the screen immediately on client transition

gigabo:
If we're not going to reuse the DOM this improves the perception of
responsiveness for navigation controls.  Click a link and the screen blanks
immediately.

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -539,7 +539,7 @@ class ClientController extends EventEmitter {
 	 */
 	_cleanupPreviousRender(index) {
 
-		index || (index = 0); // Default to everything.
+		index = index || 0; // Default to everything.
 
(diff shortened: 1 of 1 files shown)


packages/react-server/core/ClientController.js
```diff
@@ -532,6 +538,9 @@ class ClientController extends EventEmitter {
 	 * DOM node(s) that were created.
 	 */
 	_cleanupPreviousRender(index) {
+
+		index || (index = 0); // Default to everything.
```
roblg: Nit: can we use `index = index || 0` here? I think it's a little easier to understand the intent. (And it's actually two characters shorter ;))

roblg:
Seems good to me.

packages/react-server/core/ClientController.js
gigabo: > it's a little easier to understand

I disagree but acquiesce.

PR #87, opened 2016-04-10
Added ability to configure react-server-cli with a .reactserverrc file or through package.json

aickin:
@gigabo mentioned in [a comment](https://github.com/redfin/react-server/pull/49#issuecomment-205476978) on PR #49 that it would be nice if there was a config file way to launch `react-server-cli`. I agree, and that's what this PR does.

If you send in any arguments at the command line, nothing changes. But if you don't, then it starts looking in the current directory for either a `.reactserverrc` file or a `reactServer` section of package.json. If it doesn't find either of those, it starts crawling up the directory hierarchy to find a config file. It uses the first config file it finds; it doesn't try to find multiple configs and merge them. For any values that aren't specified, it uses the same defaults it used before.

Additionally, as in Babel, it allows for config overrides on a per-environment basis. All you do is include a `env` attribute in your config, like this:

```json
{
  "port": 4000,
  "env": {
    "production": 5000
  }
}
```

In this example, the port will be 5000 when NODE_ENV is set to "production" and 4000 otherwise.

Other notes:

* I got rid of the `--production` flag; it seemed like it was excessively confusing to me, both at the command line and in terms of the code. It also felt like setting NODE_ENV to production is a fairly normal thing these days, and we should just go with that convention and not provide a second way to do it. Let me know if you disagree.
* I moved the logging warnings from the CLI-specific code into `startServer` so that they will happen for API uses as well as CLI uses.
* One side effect of this PR is that the API now has different config defaults when NODE_ENV is production, which was not true before. This is by design, but I wanted to call it out.

### DIFF
--- a/packages/react-server-cli/README.md
+++ b/packages/react-server-cli/README.md
@@ -55,7 +55,7 @@ There are three ways to pass options to the CLI, through the command line, `.rea
  1. If there is a `package.json` file in the current directory with an entry named `reactServer`, it is used. Otherwise:
 1. Go back to step 2 in the parent of the current directory. Repeat until you either find a config or hit the root directory.
 
-Note that the programmatic API also searches for config files if you don't pass in an options object.
+Note that the programmatic API also searches for config files, although options sent in explicitly override the config file.
 
 ###JSON options can be set per environment
 
@@ -163,9 +163,7 @@ start(routesRelativePath, {
 })
 ```
 
-All of the values in the second argument are optional, and they have the same defaults as the corresponding CLI arguments explained above.
-
-Note that the programmatic API also searches for config files if you don't pass in an options object. If you want to disable this behavior, send in an empty object as the options object.
+All of the values in the second argument are optional, and they have the same defaults as the corresponding CLI arguments explained above. (Also note that if an option isn't present, the programmatic API will search for a config file in the same way as the CLI.)
 
 ##Future Directions
 
--- a/packages/react-server-cli/README.md
+++ b/packages/react-server-cli/README.md
@@ -43,11 +43,39 @@ It's rare to see a project these days in the JavaScript world that isn't at leas
 To take advantage of the Babel compilation, you need to install the Babel plugins and presets you want and reference them in a `.babelrc` file in your code directory. For more on the `.babelrc` format, see [its documentation here](https://babeljs.io/docs/usage/babelrc/).
 
 ##Options
-Smart defaults are the goal, and `react-server-cli` has two base modes: **development** and **production**. `react-server-cli` will determine which base mode it's in by looking at (in order):
+Smart defaults are the goal, and `react-server-cli` has two base modes: **development** and **production**. `react-server-cli` will determine which base mode it's in by looking at the NODE_ENV environment variable. If it's not "production", then `react-server-cli` will assume we are in development mode.
 
-1. If the `--production` flag was sent in, it's **production**.
-1. If `process.env.NODE_ENV` is `'production'`, it's **production**.
-1. Otherwise, the base mode is **development**.
+###Ways to add options
+
+There are three ways to pass options to the CLI, through the command line, `.reactserverrc` JSON files, or as a `reactServer` entry in `package.json` files. Once `react-server` finds one of these sources of options, it will not search for more. It searches for options this way:
+
+1. If there are any options arguments on the command line, they are used. Otherwise:
+1. `react-server-cli` looks at the current directory.
+ 1. If there is a JSON file named `.reactserverrc` in the directory, it is used. Otherwise:
+ 1. If there is a `package.json` file in the current directory with an entry named `reactServer`, it is used. Otherwise:
+1. Go back to step 2 in the parent of the current directory. Repeat until you either find a config or hit the root directory.
+
+Note that the programmatic API also searches for config files if you don't pass in an options object.
+
+###JSON options can be set per environment
+
+If you are using either `.reactserverrc` or `package.json` to set your react-server options, you can provide environment-specific values in a sub-object at the key `env`. It looks like this:
+
+```json
+{
+	"port": "5000",
+	"env": {
+		"staging": {
+			"port": "4000"
+		},
+		"production": {
+			"port": "80"
+		}
+	}
+}
+```
+
+The values in a particular environment override the main settings. In this example configuration `port` will be set to 80 if `process.env.NODE_ENV` is `production`, 4000 if `process.env.NODE_ENV` is `staging`, and 5000 for any other situation.
 
 ###Development mode: making a great DX
 
@@ -73,7 +101,7 @@ To use `react-server-cli` in this sort of production setup, follow these steps:
 
 ###Setting Options Manually
 
-While development and production mode are good starting points, you can of course choose to override any of the setup by passing in options at the command line:
+While development and production mode are good starting points, you can of course choose to override any of the setup by setting the following options:
 
 #### --routes
 The routes file to load.
@@ -136,6 +164,8 @@ start(routesRelativePath, {
 ```
 
 All of the values in the second argument are optional, and they have the same defaults as the corresponding CLI arguments explained above.
+
+Note that the programmatic API also searches for config files if you don't pass in an options object. If you want to disable this behavior, send in an empty object as the options object.
 
 ##Future Directions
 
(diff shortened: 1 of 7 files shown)


gigabo:
> a `.reactserverrc` 

Aw yeah!

> a `reactServer` section of package.json

Huh, cool.  I have mixed feelings about tucking config into `package.json` generally.  I see the appeal of having package config in one place.  On the other hand small single-purpose configuration files provide good separation of concerns.  Not opposed to this, and it seems to be a trend.  Just voicing my skepticism. :)

> If you send in any arguments at the command line, nothing changes.

This seems unusual.  More typical behavior with configuration files is to allow command-line options to override values _individually_, right?  Whereas here if I add _any_ command-line option (e.g. `--log-level=debug`) I'm then responsible for also re-specifying everything that would otherwise have been picked up from my config?  I'd prefer a merge sieve of `defaults => config file => options`.

> it doesn't try to find multiple configs and merge them.

This is another common pattern.  Not sure if it makes sense for `react-server` or not, but it's a logical extension of per-option command-line config overrides.

> config overrides on a per-environment basis.

Awesome!  I don't get the example, though.  I would have expected the value associated with `production` to be an object with key/value configuration pairs rather than a single value:

```javascript
{
  "port": 4000,
  "env": {
    "production": {
      "port": 5000
    }
  }
}
```

aickin:
> I have mixed feelings about tucking config into package.json generally... it seems to be a trend. Just voicing my skepticism.

I hear that. I was basing it off `babel` (which you probably knew), and I thought it was kind of neat, but I'm not wedded to it. It's very removable if you want.

> More typical behavior with configuration files is to allow command-line options to override values individually, right?

I'm actually not sure which is more common; I'll investigate with babel.

What I was worried about is reproducibility of a particular CLI invocation. It just felt potentially confusing to me to have the same command line set of options do different things depending on your working directory. (Of course, I recognize at some level that having `react-server-cli` work differently in different directories is **the entire point** of this PR, but it somehow felt more constrained when it only applied to option-less invocations of the CLI. I'm not sure that's rational on my behalf.) 

I also was thinking that a single config's values probably have some interdependencies (which is why I also didn't want to merge config from different sources in the file system), but I may be wrong about that, too. Your example of just changing the log level is illuminating.

I'll look at babel's behavior and think about it a little more. Implementation should be easy no matter what the decision.

> Awesome! I don't get the example, though.

Oh, yeah, the example is just plain wrong. D'oh. Your version is correct, as is the example in the README.md. Thanks for catching it.

aickin:
Played around with babel, and it follows your proposal for merging (`defaults => config file => options`). I've also just thought about it more, and I think you are right and my original idea was wrong. 

Implemented the new options merging in 4120e21, which actually meant I got to delete a bit of code. Yay!

As for `package.json` config, in the absence of major resistance, I'm going to leave it in. If you change your mind and want it out, though, just let me know.

Thanks as always for the feedback!

packages/react-server-cli/README.md
```diff
@@ -43,11 +43,39 @@ It's rare to see a project these days in the JavaScript world that isn't at leas
...
 ##Options
-Smart defaults are the goal, and `react-server-cli` has two base modes: **development** and **production**. `react-server-cli` will determine which base mode it's in by looking at (in order):
+Smart defaults are the goal, and `react-server-cli` has two base modes: **development** and **production**. `react-server-cli` will determine which base mode it's in by looking at the NODE_ENV environment variable. If it's not "production", then `react-server-cli` will assume we are in development mode.
 
-1. If the `--production` flag was sent in, it's **production**.
-1. If `process.env.NODE_ENV` is `'production'`, it's **production**.
-1. Otherwise, the base mode is **development**.
+###Ways to add options
+
+There are three ways to pass options to the CLI, through the command line, `.reactserverrc` JSON files, or as a `reactServer` entry in `package.json` files. Once `react-server` finds one of these sources of options, it will not search for more. It searches for options this way:
```
gigabo: I think this wording isn't quite right anymore with your latest changes.  Command-line options no longer cause `react-server` to stop searching for additional sources of options.  Same with the pseudo-code below.

packages/react-server-cli/src/defaultOptions.js
```diff
@@ -0,0 +1,16 @@
+// these are the default options for startServer.
```
gigabo: Nice to have this pulled out into a dedicated module!  Also nice to have consistent defaults between the CLI and a direct call `startServer`.

packages/react-server-cli/src/findOptionsInFiles.js
```diff
@@ -0,0 +1,41 @@
...
+			packageJson = fs.readFileSync(path.join(dir, PACKAGE_JSON));
+		} catch (e) {} //eslint-disable-line no-empty
+
+		if (packageJson) {
+			const parsedPackageJson = JSON.parse(packageJson);
+			if (parsedPackageJson.reactServer) {
+				return parsedPackageJson.reactServer;
+			}
+		}
+	} while (dir !== (dir = path.dirname(dir)))
```
gigabo: > do { ... } while (dir !== (dir = path.dirname(dir)))

Yeah! 😈

packages/react-server-cli/src/parseCliArgs.js
```diff
@@ -9,52 +9,56 @@ export default (isProduction, args = process.argv) => {
...
+const removeUndefinedValues = (input) => {
+	const result = Object.assign({}, input);
+
+	for (let key of Object.keys(input)) {
+		if (result[key] === undefined) {
+			delete result[key];
+		}
+	}
+
+	return result;
```
gigabo: What's up with the explicit `undefined` => remove `undefined` dance?

packages/react-server-cli/src/startServer.js
```diff
@@ -3,19 +3,34 @@ import http from "http"
...
 
 const logger = logging.getLogger(__LOGGER__);
 
-export default (routesRelativePath, {
-		port = 3000,
-		jsPort = 3001,
-		hot = true,
-		minify = false,
-		compileOnly = false,
+export default (routesRelativePath, options = {}) => {
```
gigabo: Ahhh... That's nice. 😌

packages/react-server-cli/src/startServer.js
```diff
@@ -128,3 +143,33 @@ const handleCompilationErrors = (err, stats) => {
...
 }
+
+const setupLogging = (logLevel) => {
+	logging.setLevel('main',  logLevel);
+	// TODO: the time and gauge log levels should also be parameters.
+	if (process.env.NODE_ENV !== "production") { //eslint-disable-line no-process-env
+		logging.setLevel('time',  'fast');
+		logging.setLevel('gauge', 'ok');
+	}
+}
```
gigabo: I like that this happens when I call `startServer` directly now.  I'd like it _more_ if it happened in the _browser_, too! ;)

gigabo:
This is awesome!

Next question: Can this merge with [this config](https://github.com/redfin/react-server/blob/master/packages/react-server/core/config.js)?  Should it?

Some values here probably shouldn't be exposed to the client.  Feels weird to have _two_ config files, though...

packages/react-server-cli/src/parseCliArgs.js
aickin: Values that aren't sent in on the command line come in as `undefined` (except for boolean values, which come in as `false`.

Sending in a value with `undefined` overrides the default, because I used `Object.assign` to merge options, so I remove them before sending in to startServer.

I'll comment it.

aickin:
Changed the README and added comments on the CLI `undefined` dance. Thanks!

> Next question: Can this merge with this config? Should it?

Oh hey. I'd completely forgotten about that file's existence (even though blame says I wrote significant parts of it... sigh). I can't easily figure out what was in that config; do you know, or do I need to go code spelunking?

gigabo:
I don't think _anything_ is put in it or used from it by `react-server` itself.  But `react-server` transfers it automagically from the server to the browser, which is convenient.  We use it for things like asset versions and server domains.  I guess all stuff that can be transferred via a separate script include?

@roblg - Any reason not to just kill it?

aickin:
> I don't think anything is put in it or used from it by react-server itself.
> @roblg - Any reason not to just kill it?

If that's the case, how would you feel about filing it as a bug and separating it from this PR? It seems like it's not something that most folks writing app code need to know about?

roblg:
> Any reason not to just kill it?

Only the reasons that you already outlined -- mainly that it's available both on the client and server. I think it's baked into the lifecycle in `ClientController` a little bit, but I can't imagine it'd be that hard to replace w/ a separate script.

gigabo:
> filing it as a bug and separating it from this PR

[Done](https://github.com/redfin/react-server/issues/91).

PR #89, opened 2016-04-11
Point to `react-server-cli` from the top-level README

gigabo:
It's a much faster way to get up and running.


gigabo:
Okay, I'm done fiddling with it. :)

aickin:
LGTM!

PR #71, opened 2016-03-31
Upgrade eslint

doug-wade:
Clean up a bunch of eslint errors while I'm at it.

PR #90, opened 2016-04-12
Add hello world example

aickin:
This PR ports over my super barebones hello, world example from [aickin/react-server-examples](https://github.com/aickin/react-server-examples). It really only demos a working site that has server and client rendering working.

I think it's pretty self explanatory. The only real thing to note is that I did enable two React eslint rules project wide so that it wouldn't complain about having an unused React variable.


aickin:
Ugh. I think lerna doesn't like the fact that `hello-world` is not a direct subdirectory of `packages`. It was erroring for me locally too but I figured I had done something wrong.

.eslintrc
```diff
@@ -91,6 +91,8 @@
       0,
       "double"
     ],
+    "react/jsx-uses-react": 2,
+    "react/react-in-jsx-scope": 2,
```
gigabo: Nice.  Will these need to change at all when the [eslint upgrade](https://github.com/redfin/react-server/pull/71) lands?

gigabo:
>  I think lerna doesn't like the fact that hello-world is not a direct subdirectory of packages.

Hmm... I think lerna might be on to something there, too.  `hello-world` is [already taken](https://www.npmjs.com/package/hello-world), for one thing.

Would it make sense for `react-server-examples` ([still available](https://www.npmjs.com/package/react-server-examples)!) to be the public module and have the individual examples be private?

packages/react-server-examples/hello-world/HelloWorldPage.js
```diff
@@ -0,0 +1,11 @@
+import React from "react"
+import HelloWorld from "./HelloWorld"
+
+export default class SimplePage {
+	getElements () {
+		return [
+			<HelloWorld/>,
+		];
```
gigabo: Maybe simpler still to just `return <HelloWorld/>;`?

aickin:
> I think lerna might be on to something there, too. hello-world is already taken, for one thing.

Yeah, but the actual module name is react-server-examples-hello-world, so maybe that's not an issue?

> Would it make sense for react-server-examples (still available!) to be the public module and have the individual examples be private?

Perhaps, although do we really need a public module at all? I'm not sure what we gain from making any of it public. Do you know of any good documentation for lerna (other than its git README)? I feel like I don't really entirely even understand the problem it's trying to solve.

I'll fiddle with it today or tomorrow.

gigabo:
> do we really need a public module at all?

Yeah, I guess not.  Doesn't even really make sense.  Same with `react-server-test-pages`.  Maybe we need something next to `packages` to contain modules that we don't publish?  Or maybe we need to augment `lerna` to bypass private modules (if it doesn't already)?  Or we need to ditch `lerna`?

aickin:
Figured it out. lerna doesn't care about whether a module is private or not, but it does throw an error on modules without a name. Added `name` to the `react-server-examples` package.json, and lerna stopped complaining. 

Also addressed @gigabo's suggestion to not return an array in `getElements()`.

Should be good to merge now!

packages/react-server-examples/hello-world/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "0.0.1",
+  "description": "",
+  "main": "HelloWorld.js",
+  "scripts": {
+		"start": "react-server-cli",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Sasha Aickin",
+  "license": "Apache-2.0",
+  "dependencies": {
```
RobKopp: Mind adding a reference to 'babel-runtime' here to make this truly plug and play?  Because of the routes file construction the error is consumed if installing this example onto a totally fresh machine.  And yes, I did not read the dependencies :)

aickin:
Eek. Thanks, @RobKopp!

Should be fixed in e686258, and I filed bug #95 to look into the root cause for the error swallowing.

packages/react-server-examples/hello-world/package.json
aickin: Eek. Thanks, @RobKopp!

Should be fixed in e686258, and I filed bug #95 to look into the root cause for the error swallowing.

gigabo: That's one of the error types addressed by #80.

gigabo:
Thanks @aickin! 🎉

PR #93, opened 2016-04-13
Focus on the child frame is being called too early

lidawang:
Keypress events are not getting caught by the child frame until the user interacts with it, we need to wait until the child frame is fully loaded before trying to give it focus.


gigabo:
@lidawang - It looks like this will prevent `focus` from being called on a second navigation to the same frame (when the frame URL is the same).  Have you tested that case?

```
navigate to page in frame => click back => navigate to _same_ page in frame
```

lidawang:
Hoho! Nice catch - it does not handle that case.

lidawang:
This PR should have the fix for the 2nd-navigation-to-child-frame-not-getting-focus + no conflicts with the base branch ☀️

packages/react-server/core/FramebackController.js
```diff
@@ -310,6 +310,7 @@ class FramebackController extends EventEmitter {
 
 		// We've got it now, so let's set it.
 		this.setTitleFromFrame();
+		this.frame.contentWindow.focus();
```
gigabo: Nice one-line fix, @lidawang. :)

A blank line above this would make it clearer that the comment for the `setTitleFromFrame` isn't also meant to apply to the `.focus()` call.  A _comment_ for the `.focus()` call would make it even _more_ clear. ;)

gigabo:
Thanks @lidawang!

PR #94, opened 2016-04-14
Adding needed dependencies so test pages project can startup.

aickin:
The test pages project wasn't starting up correctly because it needs `react`, `react-dom`, and `superagent` to start up. I added them to the `package.json`.


roblg:
LGTM.

gigabo:
Nice catch.  It also needs the _next_ version of `react-server`, which hasn't been published yet.

Should probably publish again soon.

aickin:
> It also needs the next version of react-server

Oh yeah, didn't notice that because lerna linked them locally.

PR #96, opened 2016-04-15
Long-term caching

aickin:
Adds the `longTermCaching` option to `react-server-cli` to make the static files cacheable with far-future expires headers.

The implementation was based on [this article](https://medium.com/@okonetchnikov/long-term-caching-of-static-assets-with-webpack-1ecb139adb95#.y5tar3n8h) of webpack caching best practices. It seems to be working in both production and non-production modes, as well as with static assets that are built independently and uploaded to a separate server (using `compileOnly` in concert with `jsUrl`). A few things to note:

- I was never able to get the common chunk hash 100% deterministic when unrelated modules change, though it should remain deterministic when the codebase doesn't change. Also, entry point files that are unrelated to the code change should continue to have the same hash.
- Even when you provide `jsUrl`, the client code is built so that the server knows the static file URLs to include.
- This code writes a `manifest.json` file to the static code directory that includes all the resulting file names, along with some other information. I think it can be a useful general place for metadata about the compilation.
- When using `compileOnly` & `jsUrl` to serve static files from somewhere else, there is a reasonably big danger that the client and server codebases are slightly different. If this happens, everything will fail pretty spectacularly at runtime, since the file hashes will all be wrong. In the future, I think it's a good idea if the server downloads the static files' `manifest.json` and does an integrity check to make sure that server and client codebases are the same.


roblg:
LGTM

packages/react-server-cli/src/coreJsMiddleware.js
```diff
@@ -1,17 +1,33 @@
...
-					text: baseUrl ? `
-						if (typeof window !== "undefined" && window.__setReactServerBase) {
-							window.__setReactServerBase(${JSON.stringify(baseUrl)});
-						}` : "",
+					text: `
+						if (typeof window !== "undefined") {
+							if (window.__setReactServerBase) {
+								window.__setReactServerBase(${JSON.stringify(baseUrl)});
+							}
+							window.__reactServerManifest = ${manifest ? JSON.stringify(manifest) : "{}"};
```
gigabo: This could get big for a large site, right?  And it is, itself, static for a build?  I wonder if there's a way we could sneak it into the common bundle...

aickin: It could get big; it's probably ~100 bytes per entry point (that's a guess; I haven't looked at it to measure).

I believe the issue with it being in the bundle is that putting it in the bundle would change all bundle hashes any time any file changes and defeat the purpose of long-term caching.

gigabo: Yeah, I was thinking something sneakier, like just appending some extra JS to the end of the common bundle after webpack's done with it. 👹 

Maybe not worth the evil until someone with a large site decides it would make a difference.

aickin: Oh that is sneaky. I like it! But yes, probably best to address if/when it becomes a real problem. ;)

gigabo:
This looks good to me, too.  Unfortunately some conflicts with master have snuck in.

aickin:
>  Unfortunately some conflicts with master have snuck in.

I have no one to blame but myself 😄 

Should be fixed in de4d003. Had to do a little bit of surgery to startServer.js, all concentrated at the end of `startImpl` and the new function `startDummyJsServer`. The HTTPS changes in master which moved the starting of the servers into `startServers` were not playing well with the changing of serverRoutes from value to Promise in this branch.

The rest of the merge was pretty trivial.

gigabo:
Thanks @aickin! 🚀

PR #97, opened 2016-04-15
Better frameback

gigabo:
This is a big change to how frameback (and the history stack generally) is managed, but the changes in behavior are pretty straightforward.

Previously:
- Frameback
    - Each click into a frame destroyed and recreated the frame.
    - Any click _within_ the frame caused a full browser navigation.
- Generally
    - Client transition options (`reuseDom`, `bundleData`) only applied on first click, not on forward/backward browser nav.

With this changeset:
- Frameback
    - A click into a frame _may_ perform a client transition _within_ the frame to the new page.
        - Client transition options (`reuseDom`, `bundleData`) are available.
    - A click _within_ the frame may perform a client transition.
- Generally
    - Client transition options work on forward/backward browser nav.

Sounds simple, right? :feelsgood: 

Some things actually _did_ get simpler.  Some things will make your head hurt.  I've tried to address those with commentary.  Please don't be shy about calling out confusing bits.

### DIFF
--- a/packages/react-server-test-pages/pages/navigation/playground.js
+++ b/packages/react-server-test-pages/pages/navigation/playground.js
@@ -50,10 +50,21 @@ class ClientRenderIndicator extends React.Component {
 		super(props);
 		this.state = {ready: false};
 	}
+	_handleNavEvent(evt) {
+		if (!this._mounted) return;
+		this.setState({ready: evt === 'loadComplete'});
+	}
 	componentDidMount() {
+		this._mounted = true;
 		this.setState({ready: true});
-		window.__reactServerClientController.context
-			.onNavigateStart(() => this.setState({ready: false}));
+		const nav = window.__reactServerClientController.context.navigator;
+		this._navListeners = ['navigateStart', 'loadComplete']
+			.reduce((m,e) => (nav.on(e, m[e] = this._handleNavEvent.bind(this, e)), m), {});
+	}
+	componentWillUnmount() {
+		this._mounted = false;
+		const nav = window.__reactServerClientController.context.navigator;
+		_.forEach(this._navListeners, (v,k) => nav.removeListener(k, v));
 	}
 	componentWillReceiveProps() {
 		this.setState({ready: true});
--- a/packages/react-server-test-pages/pages/navigation/playground.js
+++ b/packages/react-server-test-pages/pages/navigation/playground.js
@@ -37,13 +37,32 @@ const NormalLink = ({row}) => <a href={LINK(row)}>Normal Link</a>
 
 const ClientTransitionLink = ({row}) => <Link path={LINK(row)}>Client Transition</Link>
 
-const FramebackLink = ({row}) => <Link frameback={true} path={LINK(row)}>Frameback</Link>
-
 const ReuseDom = ({row}) => <Link reuseDom={true} path={LINK(row)}>Reuse DOM</Link>
 
 const BundleData = ({row}) => <Link bundleData={true} path={LINK(row)}>Bundle Data</Link>
 
 const BundleDataAndReuseDom = ({row}) => <Link bundleData={true} reuseDom={true} path={LINK(row)}>Bundle AND Reuse</Link>
+
+class FramebackCell extends React.Component {
+	constructor(props){
+		super(props);
+		this.state = {available: true};
+	}
+	componentDidMount() {
+		if (window.__reactServerIsFrame) {
+			this.setState({available: false});
+		}
+	}
+	render() {
+		return <span className={this.state.available?'available':'not-available'}>
+			<Link path={LINK(this.props.row)} frameback={true} {...this.props.link}>{
+				this.props.children
+			}</Link>
+		</span>
+	}
+}
+
+const FramebackLink = props => <FramebackCell {...props}>Frameback</FramebackCell>
 
 class ClientRenderIndicator extends React.Component {
 	constructor(props){
--- a/packages/react-server-test-pages/pages/navigation/playground.js
+++ b/packages/react-server-test-pages/pages/navigation/playground.js
@@ -64,6 +64,23 @@ class FramebackCell extends React.Component {
 
 const FramebackLink = props => <FramebackCell {...props}>Frameback</FramebackCell>
 
+const FramebackCTLink = props => <FramebackCell link={{reuseFrame:true}} {...props}>
+	Frameback with Client Transition
+</FramebackCell>
+
+const FramebackBDLink = props => <FramebackCell link={{reuseFrame:true, bundleData:true}} {...props}>
+	Frameback with Bundle
+</FramebackCell>
+
+const FramebackRDLink = props => <FramebackCell link={{reuseFrame:true, reuseDom:true}} {...props}>
+	Frameback with DOM Reuse
+</FramebackCell>
+
+const FramebackBDRDLink = props => <FramebackCell link={{reuseFrame:true, bundleData:true, reuseDom:true}} {...props}>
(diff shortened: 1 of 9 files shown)


packages/react-server-test-pages/pages/navigation/playground.js
```diff
@@ -33,27 +33,58 @@ const PagePointer = ({page, row}) => <div className="page-pointer">
...
+const NL       = ({row}) => <a href={LINK(row)}>Normal Link</a>
+const CT       = ({row}) => <Link path={LINK(row)}>CT</Link>
+const RD       = ({row}) => <Link reuseDom={true} path={LINK(row)}>CT/RD</Link>
+const BD       = ({row}) => <Link bundleData={true} path={LINK(row)}>CT/BD</Link>
+const BDRD     = ({row}) => <Link bundleData={true} reuseDom={true} path={LINK(row)}>CT/BD/RD</Link>
+const FB       = (props) => <FBL {...props}></FBL>
+const FBCT     = (props) => <FBL {...props} link={{reuseFrame:true}}>CT</FBL>
+const FBCTBD   = (props) => <FBL {...props} link={{reuseFrame:true, bundleData:true}}>CT/BD</FBL>
+const FBCTRD   = (props) => <FBL {...props} link={{reuseFrame:true, reuseDom:true}}>CT/RD</FBL>
+const FBCTBDRD = (props) => <FBL {...props} link={{reuseFrame:true, bundleData:true, reuseDom:true}}>CT/BD/RD</FBL>
```
gigabo: Have a lot of links now.  Gave them shorter names.  Added a legend to the page.

image

packages/react-server/core/ClientController.js
```diff
@@ -69,7 +76,6 @@ class ClientController extends EventEmitter {
 		ReactServerAgent.cache().rehydrate(dehydratedState.InitialContext['ReactServerAgent.cache']);
 		this.mountNode = document.getElementById('content');
 
-		this._setupFramebackController();
```
gigabo: This is attached to the request context now, since the navigator needs to get at it.

packages/react-server/core/ClientController.js
```diff
@@ -172,27 +243,6 @@ class ClientController extends EventEmitter {
...
-				}
-				this._history.pushState({
-					reactServerFrame: pushFrame(),
-				}, null, path);
-			} else if (type === History.events.PAGELOAD && this._history && this._history.canClientNavigate()) {
-				this._history.replaceState({
-					reactServerFrame: pushFrame(),
-				}, null, path);
-			}
-
```
gigabo: We were pushing state after running `handleRoute`.  Moved it up into `_startRequest`.

packages/react-server/core/ClientController.js
```diff
@@ -558,7 +608,7 @@ class ClientController extends EventEmitter {
 					// and should destroy it. Destruction means
 					// first unmounting from React and then
 					// destroying the DOM node.
-					React.unmountComponentAtNode(root);
+					ReactDOM.unmountComponentAtNode(root);
```
gigabo: Oops, this is just an incidental fix to hush React's deprecation warning.

roblg:
I tried to pull this branch to play with it, and I can't get `react-server-cli` to work in `react-server-test-pages`.

I ran:
```
lerna bootstrap
cd packages/react-server-test-pages
npm start
```

and got:

`sh: react-server-cli: command not found`

I can't run `react-server-cli` via `node_modules` because `lerna bootstrap` does the funky not-an-npm-link thing.

I also ran into #98 futzing around with things.

Any advice on how I might go about getting the server running? :grin:

roblg:
Figured out a workaround: used `npm install ../react-server-cli` from the `react-server-test-pages` package and all was well.

packages/react-server/core/ClientController.js
```diff
@@ -151,10 +156,76 @@ class ClientController extends EventEmitter {
 				this.context.registerRequestLocal();
 			}
 		}
+
+		// If this is a History.events.PUSHSTATE navigation,
+		// and we have control of the navigation bar (we're
+		// not in a frameback frame) we should change the URL
+		// in the bar location bar before rendering.
```
roblg: typo: "in the bar location bar"

packages/react-server/core/ClientController.js
roblg: Incidental or not, good change :)

gigabo:
> used npm install ../react-server-cli from the react-server-test-pages package and all was well.

Ugh.  Yeah, I'm not sure if we're somehow mis-using `lerna` or if it just doesn't do what we want it to do: Link together the cross dependencies within our monorepo.

The `react-server-test-pages` in this branch also needs the `react-server` from this branch.

roblg:
`react-server` worked, I think. I think the issue that I was running into w/ `react-server-cli` is that is has a binary defined in package.json.

packages/react-server/core/ClientController.js
gigabo: Heh, wonder how long [that's been there](https://github.com/redfin/react-server/pull/97/files#diff-db9055542b789c676642bfe7f729f1ecL175).  Will fix.

roblg:
BTW, I reviewed it I think as throughly as I'll be able to. I pulled it down and it works, and it seems to remember things through forward and back buttons. 👍

gigabo:
Awesome.  Thanks for looking at it.

BTW - I think I got one behavior _wrong_ here that I'm going to address before I merge: Frameback links in a frame are only safe to convert to normal client transitions _if `reuseFrame` is set_.  Currently they're _always_ converted.  If `reuseFrame` isn't set, I think it's (unfortunately) only safe to convert them to top-level full browser navigation.

packages/react-server/core/ClientController.js
```diff
@@ -169,30 +240,9 @@ class ClientController extends EventEmitter {
...
-						reactServerFrame: pushFrame(),
-					}, null, location.path);
-				}
-				this._history.pushState({
-					reactServerFrame: pushFrame(),
-				}, null, path);
-			} else if (type === History.events.PAGELOAD && this._history && this._history.canClientNavigate()) {
-				this._history.replaceState({
-					reactServerFrame: pushFrame(),
-				}, null, path);
```
gigabo: Oh, just realized I introduced a bug here.  It's tough to see because this whole block of code had to _move_, but I also eliminated the `replaceState` on `PAGELOAD`.  My reasoning was that any client transition away from the initial page would handle the `replaceState`, so the one at load time seemed redundant.  I had forgotten about non-`react-server` users of the history navigation API. :hurtrealbad:

gigabo:
> If reuseFrame isn't set, I think it's (unfortunately) only safe to convert them to top-level full browser navigation.

Turned out not be as bad as all that.  Converting instead to a direct non-`reuseFrame` frameback navigation.  Nice that things like this can finally be [pretty straightforward](https://github.com/redfin/react-server/pull/97/commits/cc524d34e72d61e6087474671be8fbce0eaa7c2e).  Used to be that any change to frameback logic was a nightmare. 😱

gigabo:
(Moving this from edited PR description.  Progress bar didn't how up in list view... maybe it will if this is in a new comment?)

Still have a few of items to deal with that came up since this PR was created:

- [x] Handle non-`reuseFrame` frameback links within the frame.
- [ ] Set history nav state on page load
- [ ] Test interaction with non-`react-server` history nav users

gigabo:
Okay, I've addressed the remaining issues here.

Going to merge once the build passes.  If you want to look at this before I do, stop me before then. 😛

PR #101, opened 2016-04-18
Frame skipping

gigabo:
A bug had crept in here where an infrequently used value that used to be at one index in an array had been pushed to a different index.  Changed the array to an object to help future-proof it a bit.


gigabo:
This one's only three lines if someone has a moment to review. 😇

roblg:
👍

PR #106, opened 2016-04-20
Data bundle cache

gigabo:
The `bundleData` option for requests was implicitly tied to a crude cache with no expiry, eviction, or TTL.

This change set eliminates that cache in `react-server` (along with data bundle preloading) and adds instead a separate package that provides more full-featured data bundle caching (and preloading) as an optional plugin.

Features:
- Configurable TTL (lifetime of bundles)
- Configurable cache size (number of bundles)
- Page-level opt-in

The [`README`](https://github.com/gigabo/react-server/blob/data-bundle-cache/packages/react-server-data-bundle-cache/README.md) gives an overview.  That might be a good place to start with this PR.


packages/react-server-cli/src/compileClient.js
```diff
@@ -250,7 +250,11 @@ module.exports = {
 							});`);
 			} else {
 				routesOutput.push(`
-							cb(unwrapEs6Module(require("${relativePathToPage}")));`);
+							try {
+								cb(unwrapEs6Module(require("${relativePathToPage}")));
+							} catch (e) {
+								console.error('Failed to load page at "${relativePathToPage}"', e.stack)
+							}`);
```
gigabo: This is part of #80 that I re-did in place because I needed to unmask some errors during development.  Forgot to undo before committing, and now it's buried in history.  I can squash if it bugs you, but I won't unless prompted.

gigabo: Rebased away. :)

gigabo:
Hi.  I want to start using this.  Please give it a look over if you get a chance.

packages/react-server-data-bundle-cache/src/wrap.js
```diff
@@ -0,0 +1,25 @@
+import _ from "lodash";
```
roblg: Does this result in all of `lodash` being pulled in to the client-side bundle?

packages/react-server-test-pages/pages/navigation/common.js
```diff
@@ -0,0 +1,57 @@
...
+		this._mounted = false;
+		const nav = window.__reactServerClientController.context.navigator;
+		_.forEach(this._navListeners, (v,k) => nav.removeListener(k, v));
+	}
+	componentWillReceiveProps() {
+		this.setState({ready: true});
+	}
+	render() {
+		return <div className="render-indicator">
+			{this.state.ready?"✓":"⌛️"}
```
roblg: :heart: ⌛

packages/react-server-test-pages/pages/navigation/data-bundle-cache.js
```diff
@@ -0,0 +1,89 @@
...
+	PagePointer,
+	RowIndex,
+	RowMS,
+	ClientRenderIndicator,
+} from "./common";
+
+require("./common.css");
+require("./data-bundle-cache.css");
+
+DataBundleCache.install();
```
roblg: Just to make sure I'm understanding: `DataBundleCache.install()` is something we'd typically put somewhere central, rather than per-page, right?

roblg:
Left a couple questions, but otherwise SGTM. (For what that's worth... I'm pretty tired today ;) )

packages/react-server-data-bundle-cache/src/wrap.js
gigabo: Yep.  Only if you use the plugin.

And it would be anyway by `stratocacher`, so I figured I might as well use it here, too.

packages/react-server-test-pages/pages/navigation/data-bundle-cache.js
gigabo: Yep, that's right.

PR #107, opened 2016-04-21
Support promises returned in getHeadStylesheets

roblg:
We can return promises (almost) everywhere else. Seems reasonable to return them here too. (Plus it unblocks having a middleware that reads a route-->stylesheet mapping from a file server-side and provides it to the client.)

Risks: timing of stylesheets could delay page rendering.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,7 @@ root = true
 
 [*]
 indent_style = tab
-indent_size = 1
+indent_size = 4
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/packages/react-server-test-pages/pages/styles/promises.js
+++ b/packages/react-server-test-pages/pages/styles/promises.js
@@ -1,0 +1,31 @@
+
+import React from "react";
+
+export default class StylePromises {
+	getHeadStylesheets () {
+		return [
+			delay(1000)
+				.then(() => "./styles/first.css"),
+			delay(100)
+				.then(() => "./styles/second.css"),
+		];
+	}
+
+	getElements () {
+		return [
+			<div>If first.css shows up before second.css in the body of this page, all is well.</div>,
+		]
+	}
+
+}
+
+function delay(ms) {
+	let _res;
+	const p = new Promise(function (res) {
+		_res = res;
+	});
+	setTimeout(() => {
+		_res()
+	}, ms);
+	return p;
+}
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -304,39 +304,40 @@ class ClientController extends EventEmitter {
 				this._renderLinkTags(page);
 			}
 
-			cssHelper.ensureCss(routeName, page);
+			cssHelper.ensureCss(routeName, page)
+				.then(() =>
+					page.getBodyClasses().then((classes) => {
+						classes.push(`route-${routeName}`);
+						document.body.className = classes.join(' ');
+					}))
+				.then(() => this._render(page)).catch((err) => {
+					logger.error("Error during client transition render", err);
+				}).then(() => {
 
-			page.getBodyClasses().then((classes) => {
-				classes.push(`route-${routeName}`);
-				document.body.className = classes.join(' ');
-			}).then(() => this._render(page)).catch((err) => {
-				logger.error("Error during client transition render", err);
-			}).then(() => {
-
-				// We're responsible for letting the navigator
-				// know when we're more or less done stomping
-				// around in the current request context
-				// setting things up.
-				//
-				// We can't _guarantee_ that pages/middleware
-				// haven't set timers to mess with things in
-				// the future, so we need to wait a bit before
-				// letting the navigator yank our context if
-				// an immediate subsequent navigation is
-				// scheduled.
-				//
-				// I don't like this magic delay here, but it
-				// gives us a better shot at falling after
-				// things like lazy load images do their
-				// post-render wire-up.
-				//
-				// Anything that the current page does in the
-				// request context _after_ this timeout has
-				// elapsed and we've started a subsequent
-				// navigation is pure corruption. :p
-				//
-				setTimeout(() => context.navigator.finishRoute(), 200);
-			}).done();
+					// We're responsible for letting the navigator
+					// know when we're more or less done stomping
+					// around in the current request context
+					// setting things up.
+					//
+					// We can't _guarantee_ that pages/middleware
+					// haven't set timers to mess with things in
+					// the future, so we need to wait a bit before
+					// letting the navigator yank our context if
+					// an immediate subsequent navigation is
+					// scheduled.
+					//
+					// I don't like this magic delay here, but it
+					// gives us a better shot at falling after
+					// things like lazy load images do their
+					// post-render wire-up.
+					//
+					// Anything that the current page does in the
+					// request context _after_ this timeout has
+					// elapsed and we've started a subsequent
+					// navigation is pure corruption. :p
+					//
+					setTimeout(() => context.navigator.finishRoute(), 200);
+				}).done();
 
 		});
 
--- a/packages/react-server/core/renderMiddleware.js
+++ b/packages/react-server/core/renderMiddleware.js
@@ -299,10 +299,12 @@ function writeHeader(req, res, context, start, pageObject) {
 		// PLAT-602: inline scripts come before stylesheets because
 		// stylesheet downloads block inline script execution.
 		renderScripts(pageObject, res),
-		renderStylesheets(pageObject, res),
-		renderMetaTags(pageObject, res),
-		renderLinkTags(pageObject, res),
-		renderBaseTag(pageObject, res),
+		renderStylesheets(pageObject, res)
+			.then(() => Q.all([
+				renderMetaTags(pageObject, res),
+				renderLinkTags(pageObject, res),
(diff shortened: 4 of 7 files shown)


packages/react-server/core/ClientController.js
```diff
@@ -304,39 +304,40 @@ class ClientController extends EventEmitter {
...
+					// things like lazy load images do their
+					// post-render wire-up.
+					//
+					// Anything that the current page does in the
+					// request context _after_ this timeout has
+					// elapsed and we've started a subsequent
+					// navigation is pure corruption. :p
+					//
+					setTimeout(() => context.navigator.finishRoute(), 200);
+				}).done();
```
roblg: Changes above here are almost entirely whitespace. (minus `ensureCss` now returning a promise). Github using 2 spaces for tabs makes the indentation hard to read. Sorry :(

packages/react-server/core/renderMiddleware.js
```diff
@@ -299,10 +299,12 @@ function writeHeader(req, res, context, start, pageObject) {
...
-		renderStylesheets(pageObject, res),
-		renderMetaTags(pageObject, res),
-		renderLinkTags(pageObject, res),
-		renderBaseTag(pageObject, res),
+		renderStylesheets(pageObject, res)
+			.then(() => Q.all([
+				renderMetaTags(pageObject, res),
+				renderLinkTags(pageObject, res),
+				renderBaseTag(pageObject, res),
+			])),
```
roblg: This forces meta/link/base writing _after_ stylesheets. I don't know if it's necessary, but I wanted to draw attention to this section and any ordering issues I might be introducing.

packages/react-server/core/renderMiddleware.js
```diff
@@ -590,18 +592,27 @@ function renderScripts(pageObject, res) {
...
+				res.write(`<style type="${styleSheet.type}" media="${styleSheet.media}" ${PAGE_CSS_NODE_ID}>${styleSheet.text}</style>`);
+			} else {
+				throw new Error("Style cannot be rendered because it has neither an href nor a text attribute: " + styleSheet);
+			}
+		});
 	});
 
-	// resolve immediately.
-	return Q("");
+	return Q.all(stylePromises);
```
roblg: Hmm... should probably change this to `Q.allSettled`.

packages/react-server/core/util/ClientCssHelper.js
```diff
@@ -29,49 +31,51 @@ module.exports = {
 			throw new Error("ClientCssHelper.registerPageLoad can't be called server-side");
 		}
 
-		var newCss = pageObject.getHeadStylesheets();
+		return Q.all(PageUtil.standardizeStyles(pageObject.getHeadStylesheets())).then(newCss => {
```
roblg: Will change this to `Q.allSettled` too.

packages/react-server/core/renderMiddleware.js
roblg: `Q.all` maintains backwards-compatibility in the case that a malformed style object was passed, but it introduces new behavior where failure to figure out what stylesheet to load can cause the page to fail to render. Thoughts?

packages/react-server/core/ClientController.js
gigabo: >  Github using 2 spaces for tabs makes the indentation hard to read. Sorry :(

I think that's because of [this](https://github.com/redfin/react-server/blob/master/.editorconfig#L5)?

packages/react-server/core/renderMiddleware.js
gigabo: Seems like a good reason to fail the render...

packages/react-server/core/util/ClientCssHelper.js
gigabo: Seems like these should be chained?  We want to get the _first_ stylesheet `<link/>` tags down to the browser ASAP.  If _later_ stylesheet promises are slower they could delay _earlier_ writes to the browser.

packages/react-server/core/ClientController.js
roblg: Oh. That makes sense. I apparently have no idea what that `.editorconfig` file does.

packages/react-server/core/util/ClientCssHelper.js
roblg: Hmmm... good point.

Previously (and as written in this PR) the swap of "remove old and add new" is relatively atomic. The remove/add of the dom nodes happen in the same time slice (which I assume matters, but I haven't actually tested extensively, I don't think). I could imagine adding all the new styles as they come in, then removing the old ones all at once? Seems like that could have some unintentional side effects with styles though. 

Thoughts?

gigabo: Oh, I totally mistook the context here.  Thought we were in `renderMiddleware`.  I'll go gripe over there. 👹

packages/react-server/core/renderMiddleware.js
```diff
@@ -590,18 +592,27 @@ function renderScripts(pageObject, res) {
...
-			res.write(`<link rel="stylesheet" type="${styleSheet.type}" media="${styleSheet.media}" href="${styleSheet.href}" ${PAGE_CSS_NODE_ID}>`);
-		} else if (styleSheet.text) {
-			res.write(`<style type="${styleSheet.type}" media="${styleSheet.media}" ${PAGE_CSS_NODE_ID}>${styleSheet.text}</style>`);
-		} else {
-			throw new Error("Style cannot be rendered because it has neither an href nor a text attribute: " + styleSheet);
-		}
+
+	const styles = PageUtil.standardizeStyles(pageObject.getHeadStylesheets());
+
+	const stylePromises = styles.map(styleP => {
```
gigabo: This needs to be a _chain_, right?  `<link/>` tags could get written out of order with this map.

roblg: Yep, probably 👹

.editorconfig
```diff
@@ -2,7 +2,7 @@ root = true
 
 [*]
 indent_style = tab
-indent_size = 1
+indent_size = 4
```
roblg: 😌

roblg: This obviously assumes that this is a desirable change. I can modify it / remove it if not.

gigabo: I'm a firm believer in 8-column tabs, but I'll gladly take 4 over 1!

roblg: > I'm a firm believer in 8-column tabs

I... don't think there's an emoji for how I feel about 8-column tabs. 👺

gigabo: Try `cat`-ing the source and see how many columns the tabs are.

Or try `less`.  Or `git diff`, for that matter. ;)

That's not to say that I prefer tabs for _indentation in source code_.  I _like_ 4-column _indentation_!

packages/react-server/core/renderMiddleware.js
```diff
@@ -590,18 +592,27 @@ function renderScripts(pageObject, res) {
...
-	});
+	};
 
-	// resolve immediately.
-	return Q("");
+	const styles = PageUtil.standardizeStyles(pageObject.getHeadStylesheets());
+
+	return styles.reduce( (prev, styleP) => {
+		return prev.then(() => styleP.then(writeTag));
+	}, Q());
```
roblg: @gigabo does this chaining look better?

gigabo: Yep, that should do it!

packages/react-server-test-pages/pages/styles/promises.js
```diff
@@ -0,0 +1,31 @@
...
+			delay(1000)
+				.then(() => "./styles/first.css"),
+			delay(100)
+				.then(() => "./styles/second.css"),
+		];
+	}
+
+	getElements () {
+		return [
+			<div>If first.css shows up before second.css in the body of this page, all is well.</div>,
```
gigabo: Nice!

packages/react-server-test-pages/pages/styles/promises.js
```diff
@@ -0,0 +1,31 @@
...
+function delay(ms) {
+	let _res;
+	const p = new Promise(function (res) {
+		_res = res;
+	});
+	setTimeout(() => {
+		_res()
+	}, ms);
+	return p;
+}
```
gigabo: Trying to eschew `Q`?  It's already in the deps for the test pages...

gigabo:
This looks good to me.  Nice improvement!

packages/react-server-test-pages/pages/styles/promises.js
roblg: I... have no idea why I felt like I needed to do that...

packages/react-server-test-pages/pages/styles/promises.js
```diff
@@ -0,0 +1,31 @@
+
+import React from "react";
+
+export default class StylePromises {
+	getHeadStylesheets () {
+		return [
+			delay(1000)
+				.then(() => "./styles/first.css"),
+			delay(100)
+				.then(() => "./styles/second.css"),
```
gigabo: Oh, did you `git add` these CSS files?  I don't see them here.

roblg: I was lazy and just didn't add any. But per our conversation on slack, I added some: https://github.com/redfin/react-server/pull/115

PR #108, opened 2016-04-21
Documentation whitespace changes

gigabo:
Going to make some updates to these docs, and wanted to get some
non-substantive changes out of the way first to make diffs easier to read down
the road.

- Broke up long lines
    - GitHub's diff UI breaks on very long lines
    - Linewise diffs in general are easier to read with shorter lines
    - Long lines take more cursor time to traverse (linear seek vs 2d seek)
- Removed some zero-width space characters
- Wrapped a `<head>` in backticks so it shows up


roblg:
👍

PR #109, opened 2016-04-21
Update why react server

gigabo:
Doc updates can make eyes glaze over, so I'm going to do these PRs in small file-sized chunks to keep individual PR review cost in check.

This has a few just cleanup-y changes (e.g. `React` => `ReactDOM`), but also two substantive changes:

- Replaced references to a "cache" with a "data bundle".  Don't want to give the impression that there's any _caching_ going on. There's really _transfer of data_ from the server to the client.  We're not holding onto data for longer than a single request.
- Added a section on streaming initialization to go along with streaming HTML.  Hope I Englished OK.

### DIFF
--- a/packages/react-server/docs/why-react-server.md
+++ b/packages/react-server/docs/why-react-server.md
@@ -339,8 +339,8 @@ translate very well to HTTP/2. A few of the ideas we have for HTTP/2 are:
   potentially even help overcome TCP slowstart.
 
 * **Server push of API endpoints**: In the current version of `react-server`,
-  the backend API results are sent down in a client-side cache, as noted
-  above. That cache is implemented as an inline `<script>` tag in the HTML
+  the backend API results are sent down in a data bundle, as noted
+  above. That bundle is implemented as an inline `<script>` tag in the HTML
   page, and it works, but it could be better. First, it doesn't use the
   browser cache when an API result is cacheable. Second, we currently wait
   until the end of the page to send down the client-side cache, as we don't
--- a/packages/react-server/docs/why-react-server.md
+++ b/packages/react-server/docs/why-react-server.md
@@ -3,7 +3,7 @@
 React is an amazing library for client-side user interface, and it has the
 added benefit of being able to be rendered on the server, which is crucial for
 SEO, SEM, and user experience. However, server-side rendering in practice is
-significantly more complicated than just calling `ReactDOM.renderToString` and
+significantly more complicated than just calling `ReactDOMServer.renderToString` and
 piping out the result, especially if you want to make sure your site loads
 quickly in a mobile-first world.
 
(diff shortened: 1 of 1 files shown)


roblg:
LGTM

packages/react-server/docs/why-react-server.md
```diff
@@ -3,7 +3,7 @@
 React is an amazing library for client-side user interface, and it has the
 added benefit of being able to be rendered on the server, which is crucial for
 SEO, SEM, and user experience. However, server-side rendering in practice is
-significantly more complicated than just calling `React.renderToString` and
+significantly more complicated than just calling `ReactDOM.renderToString` and
```
aickin: Should be `ReactDOMServer` to follow conventions elsewhere, I think.

packages/react-server/docs/why-react-server.md
```diff
@@ -31,8 +31,9 @@ the total time to full completion.
 In order to facilitate blazing perceived performance, `react-server` implements the following:
 
 * Parallelize backend queries
-* Use a client-side data cache
+* Data bundle transfer
```
aickin: nitpick: this doesn't have parallel structure; the rest of the bullets are all actions (i.e. have a verb). How about "Bundle data requests" or "Bundle data for transfer"? 

And same for the section header below.

packages/react-server/docs/why-react-server.md
```diff
@@ -108,30 +109,30 @@ rendering.
...
 
 ## Stream pre-rendered HTML
 
-Once we have parallel backend services and a client-side data cache, we're
+Once we have parallel backend services and data bundle transfer, we're
 well on our way to developing a web experience that has good perceived
 performance.
 
-However, as our page gets more longer and more fully featured, we run into
+However, as our page gets longer and more fully featured, we run into
```
aickin: Nice. I write more better English. 😛

packages/react-server/docs/why-react-server.md
```diff
@@ -151,6 +152,22 @@ a few milliseconds. Browsers are well equipped to render partial pages, and
 the user will be shown some content while the primary part of the page is
 still loading.
 
+## Streaming client initialization
+
+Beyond simple streaming of the HTML content, `react-server` also initializes
+the `React` components in the browser _as they arrive_.  You tell
```
aickin: Oh, I didn't know it did this. THAT'S AWESOME.

aickin:
Definitely better; love the section on aggressive rendering. 

LGTM after my comments are addressed.

packages/react-server/docs/why-react-server.md
gigabo: Yeah!  That was the flip side to dropping the synchronous render (which turned out to be pretty confusing for devs and the source of tricky bugs).

Now data is either available for the server render or it's not.  We've got a helper in our store class that makes it easy to configure which values will block the server render and which will get populated _after_ the initial client render (to avoid checksum mismatch).  There's nothing in between.

Haven't figured out how to split that scheduling/filtration out of our stores yet to make it open.  Our stores are based on reflux, and I wouldn't want to impose that as a requirement for core functionality in `react-server`.

PR #111, opened 2016-04-22
Add a doc on client transitions

gigabo:
Include info on:
- `<Link>` tag usage
- `ClientRequest` object usage
- Options for client transitions:
    - `bundleData`
    - `reuseDom`
    - `frameback`
    - `reuseFrame`

### DIFF
--- a/packages/react-server/docs/client-transitions.md
+++ b/packages/react-server/docs/client-transitions.md
@@ -1,0 +1,131 @@
+# Client Transitions
+
+React-server provides blazing fast server-side render of web pages.  This is
+important for landing experience and for web crawlers, but it's only part of
+the picture!  The great strength of `react-server` is that the _same code_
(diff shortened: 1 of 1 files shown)


doug-wade:
👍

packages/react-server/docs/client-transitions.md
```diff
@@ -0,0 +1,131 @@
...
+	ClientRequest
+	History,
+	getCurrentRequestContext,
+} = require("react-server");
+
+getCurrentRequestContext().navigate(
+	new ClientRequest(this.props.path),
+	History.events.PUSHSTATE
+);
+```
```
gigabo: We've got to make this easier.  I want to be able to initiate a client transition with a single import.

Something as simple as:

```javascript
const {navigateTo} = require("react-server");

navigateTo(path, {options});
```

PR #112, opened 2016-04-22
Remove dependency on `debug`

gigabo:
I don't think we're actually using this?

I don't think we've used this since before `logging` (~January 2015)?

Anyway, `nsp` reported a regexp denial of service vulerability.  So, instead
of upgrading I'm just removing, since... I don't think we're actually using
this? :grin:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,4 @@ addons:
       - g++-4.8
 
 before_script:
-  - npm install -g gulp lerna
-  - lerna bootstrap
-
-script: gulp travis-ci
+  - npm run bootstrap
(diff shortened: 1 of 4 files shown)


gigabo:
Note that travis is failing on this in master right now, so PRs may need to be rebased once this lands.

roblg:
👍

gigabo:
Added a few more changes here: Upgraded to the latest beta of `lerna` because that's what it took to get the dependency change in `react-server` picked up at bootstrap time.  I also cleaned up our `.travis.yml` (modeled after what I did for [stratocacher](https://github.com/redfin/stratocacher/blob/master/.travis.yml)) because that had been bugging me for a while.  Runs a little faster without the redundant global installs of gulp and lerna.

gigabo:
Gonna merge this to get other PRs unblocked.

Next time you update from master be sure to upgrade `lerna`.

PR #113, opened 2016-04-23
Get rid of base gulpfile and gulp-chug

gigabo:
Just run tests with `npm test`

Briefly tried to get integration tests back into `packages/`, but gave up
after an hour or so.  Layers of impediment there.


gigabo:
This PR moves all dev dependencies into the individual packages.  I think a next step might actually be to move in the _opposite_ direction and get all of the dev dependencies factored out to the root.  That would speed up bootstrap and keep us on consistent tooling across packages.

Definitely a separate PR, though.

gigabo:
This is pretty small.  Please take a look when you have a moment.

roblg:
👍

PR #115, opened 2016-04-27
Cleaning up promise-from-getHeadStylesheets test page

roblg:
Cleaning up getHeadStylesheets with promises test page.

@gigabo encouraged me to be cute, so I added a date endpoint for test pages that echoes CSS back.

### DIFF
--- a/packages/react-server-test-pages/pages/styles/promises.js
+++ b/packages/react-server-test-pages/pages/styles/promises.js
@@ -1,19 +1,32 @@
 
 import React from "react";
 
+const greenBackground = `
+	body {
+		background: green;
+	}
+`;
+
+const redBackground = `
+	body {
+		background: red;
+	}
+`;
+
 export default class StylePromises {
 	getHeadStylesheets () {
 		return [
-			delay(1000)
-				.then(() => "./styles/first.css"),
-			delay(100)
-				.then(() => "./styles/second.css"),
+			delay(1000).then(() => `/data/echo-css?css=${encodeURIComponent(redBackground)}`),
+			delay(100).then(() => `/data/echo-css?css=${encodeURIComponent(greenBackground)}`),
 		];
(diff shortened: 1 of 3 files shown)


roblg:
BTW, I noticed some weird behavior that I filed as #116. I didn't have time to dig into it though.

packages/react-server-test-pages/pages/styles/promises.js
```diff
@@ -1,19 +1,32 @@
...
+			delay(100).then(() => `/data/echo-css?css=${encodeURIComponent(greenBackground)}`),
 		];
 	}
 
 	getElements () {
 		return [
-			<div>If first.css shows up before second.css in the body of this page, all is well.</div>,
+			<div>
+				<h2>If the background is GREEN at page load, all is well!</h2>
+				<div>If it flashes red first, all is _not_ well.</div>
```
gigabo: Isn't it actually the page _winding up_ red that we're looking out for?  In other words _out of order_ CSS includes?

gigabo:
> @gigabo encouraged me to be cute

😘 

This is great!  Plus you found a bug while you were working on it!

packages/react-server-test-pages/pages/styles/promises.js
roblg: It would be, but for #116. ClientCssHelper getting confused + the fact that it actually waits until all promises are resolved means that everything currently ends up correct at the end (after a 1s pause w/ a red background)

roblg: Er. That is, everything works in this PR. You have to hand-modify renderMiddleware to the not-chained behavior in order to see the flash of red

gigabo: Got it.  Thanks!

PR #117, opened 2016-04-27
Frame bleed

gigabo:
There are two navigation patterns that result in a frame transition turning into a client transition.  This isn't appropriate.  There's no guarantee that a client transition is viable.

This change set addresses these two navigation patterns:

- `frameback in => refresh => back-button`
- `frameback in => back-button => refresh => forward-button`

Note that `client-transition => frameback in => back => back => refresh => forward => forward` and variants _were_ previously handled.  It was only the refesh _on_ the outer frame (immediately prior to frameback) followed immediately by navigation forward that was broken from outside.  A refresh from _any_ depth _inside_ was broken.

Frameback is fun! :feelsgood:

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -128,25 +128,41 @@ class ClientController extends EventEmitter {
 				});
 			}
 
-		} else {
+		} else if (this._previouslyRendered) {
+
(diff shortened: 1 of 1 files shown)


roblg:
I think the changes make sense to me. 

Also, I think if you pull from master, you'll get more than 1 space for tabs in your diffs :)

PR #123, opened 2016-04-29
Npm optimization

gigabo:
A better description (matches the repo description and first line of README) and some keywords.

BTW - Anyone have a better one-line description?  This thing shows up in search results, unfurled slack links, etc.  Needs to be good.

### DIFF
--- a/packages/react-server/package.json
+++ b/packages/react-server/package.json
@@ -11,6 +11,12 @@
     "prepublish": "gulp build",
     "test": "gulp test"
   },
+  "keywords": [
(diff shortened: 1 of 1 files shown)


gigabo:
Also, any other keywords that we should add?

doug-wade:
Other keywords: express framework middleware?

gigabo:
Thanks @doug-wade.  Added those + "universal".

Gonna merge when this build passes unless someone comes up with more improvements first.

PR #128, opened 2016-05-05
The path is at `location.pathname`

gigabo:
`location.path` is undefined.  This is a real bug.

Most browsers just default to the current URL when `undefined` is passed as
the third argument to `history.pushstate`, which masks the bug (makes it
irrelevant, actually).

IE11 replaces the last path element with the string "undefined".

Example: "/navigation/playground" becomes "/navigation/undefined".


doug-wade:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/pathname

checks out 

👍

PR #130, opened 2016-05-05
Rework frameback initialization for IE compatibility

gigabo:
Working across Chrome, FF, Safari and IE now.

Much simpler, too, which is nice.  Pretty close to my initial stab last year,
which wound up having trouble in FF, but seems to work now? :fingers_crossed:

### DIFF
--- a/packages/react-server/core/FramebackController.js
+++ b/packages/react-server/core/FramebackController.js
@@ -56,6 +56,7 @@ var {PAGE_CONTENT_NODE_ID} = require('./constants');
 
 var _ = {
 	assign: require('lodash/object/assign'),
+	forEach: require('lodash/collection/forEach'),
 }
 
 // Cover the whole viewport.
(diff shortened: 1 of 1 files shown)


packages/react-server/core/FramebackController.js
```diff
@@ -228,40 +229,19 @@ class FramebackController extends EventEmitter {
 
 	createFrame(url){
```
gigabo: This is probably easier to look at with a split diff rather than unified.  The entire body of this method changed.

packages/react-server/core/FramebackController.js
```diff
@@ -228,40 +229,19 @@ class FramebackController extends EventEmitter {
...
 
-		this.frame = document.createElement("iframe");
+		const frame = this.frame = document.createElement("iframe");
```
roblg: It looks like the main changes here are some 1:1 refactorings, plus removing the `this.frame.contentWindow` unload listener layer of indirection?

gigabo: Yep, it's doing the same stuff in a different way.  Each browser has a different set of implementation quirks.  Last October one such FF quirk caused the introduction of the awkward unload/load dance (see `RED-70527` for details).  That doesn't _seem_ to be necessary anymore.  Or else I've settled on a _different_ simple implementation than we were on previously that happens to also work in FF.

Or else I haven't tickled the repro case for the FF quirk correctly and this is going to bounce right back to me. :p

roblg:
👍

PR #132, opened 2016-05-06
Move devDependencies up to top level

gigabo:
This speeds up bootstrap and reduces storage requirements.

These packages are developed in the context of the monorepo.  No need for them
to independently install their dev dependencies.  This also helps us stay on
common versions across packages.

This required updating the top-level eslint config, and fixing some eslint
errors that the old version let slip through the cracks in the main
`react-server` package and `react-server-cli`.

Note that I left `react-server` on babel 5.  That's a whole other can of worms.


gigabo:
Any objections here?  Pretty basic improvement.  I've got another round of patches that I'd like to build on top of it, so please give it a quick look over if you get a chance.

roblg:
Seems reasonable to me

PR #136, opened 2016-05-09
Handle render before load in frameback frame

gigabo:
Frame may already be fully rendered during `_handleFrameLoad`.


doug-wade:
👍

roblg:
👍

gigabo:
Thanks!

PR #137, opened 2016-05-09
Upgrade to latest stable lodash

gigabo:
The new version doesn't support modularization in the same way as our old
version.  So, here, I've just abandoned it.

This adds ~20KB to minified compressed bundle.  It's simpler.  Worth it?

### DIFF
--- a/packages/react-server/package.json
+++ b/packages/react-server/package.json
@@ -28,7 +28,7 @@
     "cookie-parser": "1.3.3",
     "eventemitter3": "0.1.6",
     "express-state": "1.2.0",
-    "lodash": "3.10.1",
+    "lodash": "^4.12.0",
     "mkdirp": "0.5.0",
     "mobile-detect": "1.3.0",
     "q": "1.4.1",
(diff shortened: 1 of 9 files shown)


roblg:
I was under the impression that lodash v4 was supposed to be _better_ at modularization that v3? 

I can't link into headers in release notes, but... https://github.com/lodash/lodash/releases/tag/4.0.0

gigabo:
Oh... huh. :)

gigabo:
Okay, pushed a version that's modularized.

Still ~4KB bigger, but probably worth it to be off of creaky ancient technology.

doug-wade:
lgtm

http://i.imgur.com/vqD8jZh.gif

roblg:
👍

packages/react-server/package.json
```diff
@@ -28,7 +28,7 @@
     "cookie-parser": "1.3.3",
     "eventemitter3": "0.1.6",
     "express-state": "1.2.0",
-    "lodash": "3.10.1",
+    "lodash": "^4.12.0",
```
roblg: Just out of curiosity, did we intentionally unpin this? ("Yes" is a reasonable answer, just curious.)

gigabo: Nope, just did an `npm install --save lodash@latest`.

I'm not as paranoid about pinning as I once was.  I'd actually like to automatically pick up patch-level changes.

Unless they break everything.  That would suck.

PR #138, opened 2016-05-10
Avoid an infinite refresh loop in IE9

gigabo:
We do a `history.replaceState` on page load to mark the navigation stack frame
with request info.  On browsers that don't support the history navigation API
this was causing a _refresh_ on page load, which lead to an infinite refresh
loop.

This patch modifies the fallback behavior to only trigger a navigation if the
URL changes during a `replaceState`.


doug-wade:
👍

roblg:
👍

PR #134, opened 2016-05-09
Use relative dependencies in private modules

gigabo:
This makes test pages and integration tests work out of the box.

Also moved the integration tests into packages to speed up `npm test` and eliminate an odd-man-out.

PR #140, opened 2016-05-11
Move tagger back to top and trim prefixes

gigabo:
Eliminates a bootstrap race with relative deps.

(Prerequisite for #134)


roblg:
👍

PR #141, opened 2016-05-11
Upgrade to babel 6

gigabo:
This will require anybody who reaches into buildutils \*_cough_\* to upgrade as well when they pull this in.


doug-wade:
I dunno, both stage 0 proposals look pretty sweet: http://babeljs.io/docs/plugins/transform-function-bind/ http://babeljs.io/docs/plugins/transform-do-expressions/

gigabo:
Those are pretty sweet, but... I've already `npm install --save-dev babel-preset-stage-1` in like 200 places. :whine:

PR #142, opened 2016-05-12
Linking to source file

davidalber:
This is tiny. I'm just linking from documentation to a source file.


doug-wade:
:+1:

gigabo:
Thanks @davidalber.

PR #148, opened 2016-05-13
Issue#145 Replace __reactServerTimingStart with browser's navigation start event

TonyHYK:
Now that performance API is more readily available in browsers, we want to use the navigation start event for t0 on the client side. This will allow us to remove the script tag that we write synchronously to log the timing on the client side.

The current proxy for t0 comes down after the head is written. This change will inadvertently increase timings that were relying on `__reactServerTimingStart` with however long it takes to write the first byte (TTFB).

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -53,6 +53,7 @@ class ClientController extends EventEmitter {
 	constructor ({routes}) {
 		super();
 
+		window.__reactServerTimingStart = window.performance ? window.performance.timing.navigationStart : new Date;
 		var wakeTime = new Date - window.__reactServerTimingStart;
 
 		var dehydratedState = window.__reactServerState;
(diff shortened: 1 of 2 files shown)


gigabo:
@TonyHYK - I like the look of this!

Looks like travis caught a bug: Need to handle undefined `window.performance`.

TonyHYK:
Yeah I thought about having a fallback but I wasn't too sure if there's other things out there I can use. Navigation timing is supported by IE9+. So we could potentially not log the timings if the API is not available?
Thoughts?

gigabo:
That seems reasonable, though maybe it would be less work to just have a fallback?

Like:

```javascript
window.__reactServerTimingStart = window.performance
    ?window.performance.timing.navigationStart
    :new Date
```

Whatever's easiest.  Not too concerned about IE8. :)

TonyHYK:
Ah ok sounds reasonable, I wasn't sure to what extent we want to support old browsers. Fixed and rebased. Also, is there a way to run the build locally?

gigabo:
Yep, it's just `npm test` from the repo root.

gigabo:
This is great @TonyHYK!  Glad to have that `<script>` out of the beginning of the `<head>`!

PR #149, opened 2016-05-13
Ignore debug logs

doug-wade:
After fighting with some `EPEERINVALID`s this morning, I had a lot of debug logs laying around

```
dougwade code/react-server ‹master*› » git status
On branch master
Your branch is ahead of 'origin/master' by 54 commits.
  (use "git push" to publish your local commits)
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	lerna-debug.log
	packages/babel-preset-react-server/npm-debug.log.2854911462
	packages/react-server-cli/npm-debug.log.1636025210
	packages/react-server-cli/npm-debug.log.31f6baf1036b82852387dabe0669597a
node_modules
	packages/react-server-cli/npm-debug.log.9ad830a1b74a0705593796b07d637bba
	packages/react-server-data-bundle-cache/npm-debug.log.3960918523
	packages/react-server-examples/npm-debug.log.4116834042
	packages/react-server-examples/todo/
	packages/react-server-gulp-module-tagger/npm-debug.log.709866108
	packages/react-server-gulp-module-tagger/npm-debug.log.b8e5d50f060f9e51a728b854c65cd8d8
	packages/react-server-integration-tests/
	packages/react-server-test-pages/npm-debug.log.2842354137
	packages/react-server/npm-debug.log.2710223294
	packages/react-server/npm-debug.log.3197676522
	packages/react-server/npm-debug.log.3418084258
	packages/react-server/npm-debug.log.615368861
	packages/react-server/npm-debug.log.95e416bd17e539aeac73e5325db6c71f
	packages/react-server/npm-debug.log.a3051bf0c1223191840a0e690115f9ff

nothing added to commit but untracked files present (use "git add" to track)
```

I'm lazy and don't want to clean them up, so I'd like to .gitignore them instead.

PR #150, opened 2016-05-13
#44 add clean scripts

### LINKED ISSUE #44: Should "clean" before publishing

doug-wade:
Fix #44 ; unblock #149

### DIFF
--- a/packages/babel-preset-react-server/package.json
+++ b/packages/babel-preset-react-server/package.json
@@ -4,7 +4,9 @@
   "description": "A recommended Babel config for working with react-server",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"All good!\""
+    "test": "echo \"All good!\"",
+    "clean": "rimraf npm-debug.log*",
+    "prepublish": "npm run clean"
   },
   "repository": "redfin/react-server",
   "keywords": [
@@ -29,5 +31,8 @@
     "babel-plugin-transform-es2015-template-literals": "^6.3.13",
     "babel-plugin-transform-object-rest-spread": "^6.3.13",
     "babel-preset-react": "^6.5.0"
+  },
+  "devDependencies": {
+    "rimraf": "^2.5.2"
   }
 }
--- a/packages/react-server-gulp-module-tagger/package.json
+++ b/packages/react-server-gulp-module-tagger/package.json
@@ -4,7 +4,9 @@
   "description": "A gulp plugin to replace tags with module information",
   "main": "index.js",
   "scripts": {
-    "test": "gulp test"
+    "test": "gulp test",
+    "clean": "rimraf target npm-debug.log*",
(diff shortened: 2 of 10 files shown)


gigabo:
Nice!  I had been imagining putting `gulp clean` everywhere, but this is even easier. 👹 

👍

packages/babel-preset-react-server/package.json
```diff
@@ -29,5 +31,8 @@
     "babel-plugin-transform-es2015-template-literals": "^6.3.13",
     "babel-plugin-transform-object-rest-spread": "^6.3.13",
     "babel-preset-react": "^6.5.0"
+  },
+  "devDependencies": {
+    "rimraf": "^2.2.8"
```
gigabo: Oh, _actually_, this could be pulled out to the top level, right?

gigabo: Nice to avoid distributing `devDependencies` around, when possible.

Had just finally gotten rid of most packages `devDependencies`. ;)

doug-wade: The downside of pulling them to the monorepo root is reduced portability.  In particular, dependencies that define a `bin` script have the script added to `node_modules/.bin`, which is available to `package.json` scripts hash.  If I were to pull this dep to the monorepo root, I'd have to specify a path to the bin script (i.e. `../../node_modules/.bin/rimraf`), which would break our scripts for developers on windows machines.

gigabo: > If I were to pull this dep to the monorepo root, I'd have to specify a path to the bin script

I think `npm` just figures it out, actually.  As with `gulp`, for instance.

doug-wade:
So, @gigabo and I talked about this offline, and cleaning is different than our other build targets in that it _must_ be called before a module is published.  We thought a little bit about adding a `clean` target the the root `package.json` that was called in a `prepublish` step for the monorepo, but it would required a lot of gymnastics to get it to work, and that (like `gulp`) devs would have to have `rimraf` installed globally to use `lerna` instead of `npm` to do their builds, so we opted to keep distributed `rimraf` dependencies.

packages/react-server-gulp-module-tagger/package.json
```diff
@@ -4,7 +4,9 @@
   "description": "A gulp plugin to replace tags with module information",
   "main": "index.js",
   "scripts": {
-    "test": "gulp test"
+    "test": "gulp test",
+    "clean": "rimraf target npm-debug.log*",
+	"prepublish": "npm run clean"
```
gigabo: Indentation looks off here.

doug-wade:
I'm still working on this, just blocked on https://github.com/redfin/react-server/issues/210 for a quick sec

PR #151, opened 2016-05-13
npmignore npm-debug logs

doug-wade:
Unblock #149


gigabo:
Belt _and_ suspenders. 👍

doug-wade:
The travis build failed with `ECONNREFUSED 127.0.0.1:8769`, which is weird.  I restarted it with the hopes that it is transient (since it seems unrelated to my changes).

PR #143, opened 2016-05-12
Remove Sasha's early 2015 data proposal

gigabo:
This pains me, because it's a great doc with an overview of the problem and
the enviroment and a good explanation for where we went with data through most
of 2015.  But, it has gotten out of date and the _stores_ part of it actually
isn't discussing anything that's (currently) part of `react-server`.

This resolves #121.

PR #152, opened 2016-05-13
Add design doc

doug-wade:
Do the thing we talked about in #143


gigabo:
Sweet.  There are probably a number of design principles involved in `react-server` that aren't covered by this doc, but I'm glad to at least have _these_ there. :)

👍

doug-wade:
@gigabo if you know of any, let me know; I'm more than happy to add them to the doc before merging/announcing.

PR #153, opened 2016-05-13
Omit rootcontainers from counting againt getAboveTheFoldCount

### LINKED ISSUE #144: RootContainers count twice for getAboveTheFoldCount
We have a page that has an entire `RootContainer` that's above the fold, but we don't account for it in our `getAboveTheFoldCount` implementation. While debugging why our waterfall is looking weird (caused by this wrong count number), we noticed that the fix would be to a) take into account the `RootContainer` in our `getAboveTheFoldCount` implementation, but also b) increment our count by 2 for each entirely-above-the-fold `RootContainer`. The latter is extremely confusing and easy to miss - can we either omit `RootContainer`s from being part of the above-the-fold-count (they are elements, but they're container elements that can only contain `RootElements`, which already count for above-the-fold so it seems duplicative), or at least fixing this double-counting thing?

lidawang:
Should fix #144! Also added the test page I was using to make sure this worked.

### DIFF
--- a/packages/react-server-test-pages/pages/root/aboveTheFold.js
+++ b/packages/react-server-test-pages/pages/root/aboveTheFold.js
@@ -1,0 +1,25 @@
+import {ReactServerAgent, RootContainer, RootElement} from "react-server"; // eslint-disable-line
+
+export default class RootWhenPage {
+	handleRoute(next) {
+		this.data = ReactServerAgent.get('/data/delay?ms=200');
+		return next();
+	}
+	getElements() {
+		return [
+			<RootContainer>
+				<div>One</div>
+			</RootContainer>,
+			<RootElement when={this.data}><div>Two</div></RootElement>,
+			<RootContainer>
+				<div>Three</div>
+				<RootElement when={this.data}><div>Four</div></RootElement>
+			</RootContainer>,
+			<div>Five</div>,
+		]
+	}
+
+	getAboveTheFoldCount() {
+		return 3;
+	}
+}
--- a/packages/react-server/core/renderMiddleware.js
+++ b/packages/react-server/core/renderMiddleware.js
@@ -823,6 +823,8 @@ function writeElements(res, elements) {
 
 	// Pick up where we left off.
 	var start = RLS().nextElement||(RLS().nextElement=0);
+	// We keep a separate count for getAboveTheFoldCount in order to omit RootContainers
+	var elementCount = RLS().nextElementForATF||(RLS().nextElementForATF=0);
 
 	for (var i = start; i < elements.length; RLS().nextElement = ++i){
 
@@ -830,14 +832,14 @@ function writeElements(res, elements) {
 		if (elements[i] === ELEMENT_PENDING) break;
 
 		// Got one!
-		writeElement(res, elements[i], i);
+		var shouldCountForATF = writeElement(res, elements[i], i);
 
 		// Free for GC.
 		elements[i] = ELEMENT_ALREADY_WRITTEN;
 
 		if (PageUtil.PageConfig.get('isFragment')) continue;
 
-		if (i === RLS().atfCount - 1){
+		if (elementCount === RLS().atfCount - 1){
 
 			// Okay, we've sent all of our above-the-fold HTML,
 			// now we can let the client start waking nodes up.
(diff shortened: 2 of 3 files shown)


packages/react-server-test-pages/pages/root/aboveTheFold.js
```diff
@@ -0,0 +1,25 @@
...
+		return [
+			<RootContainer>
+				<div>One</div>
+			</RootContainer>,
+			<RootElement when={this.data}><div>Two</div></RootElement>,
+			<RootContainer>
+				<div>Three</div>
+				<RootElement when={this.data}><div>Four</div></RootElement>
+			</RootContainer>,
+			<div>Five</div>,
```
gigabo: Might be nice to have some text in the content here that says what the viewer should expect to see if things are working properly.

packages/react-server/core/renderMiddleware.js
```diff
@@ -823,40 +823,48 @@ function writeElements(res, elements) {
...
 		if (PageUtil.PageConfig.get('isFragment')) continue;
 
-		if (i === RLS().atfCount - 1){
+		if (elementCount === RLS().atfCount - 1){
 
 			// Okay, we've sent all of our above-the-fold HTML,
 			// now we can let the client start waking nodes up.
 			bootstrapClient(res)
 			for (var j = 0; j <= i; j++){
 				renderScriptsAsync([{ text: `__reactServerClientController.nodeArrival(${j})` }], res)
```
gigabo: Huh... just realized that this will emit `nodeArrival` events for container control elements... Guess that's okay?

packages/react-server/core/renderMiddleware.js
```diff
@@ -823,40 +823,48 @@ function writeElements(res, elements) {
...
-		writeElement(res, elements[i], i);
+		var shouldCountForATF = writeElement(res, elements[i], i);
 
 		// Free for GC.
 		elements[i] = ELEMENT_ALREADY_WRITTEN;
 
 		if (PageUtil.PageConfig.get('isFragment')) continue;
 
-		if (i === RLS().atfCount - 1){
+		if (elementCount === RLS().atfCount - 1){
```
gigabo: Could get rid of the awkward ` - 1` here if you move the `elementCount` increment above.

If you do, don't forget to adjust the conditional in the `else if`, below, too.

packages/react-server/core/renderMiddleware.js
lidawang: What exactly do the `nodeArrival` events do?

gigabo: They tell the `ClientController` that the server-rendered HTML for a `RootElement` has been loaded into the DOM, so it's safe to to the client-side `ReactDOM.render` into it.

gigabo:
Uh oh... looks like some of the automated tests are failing.

@lidawang - You can run those locally with `npm test` at the repo root.

lidawang:
Huhm I'm having some trouble with `npm test` - when I tried that I got errors, all originating from `react-server`'s `gulp test`, something about `eslint`?

```
Errored while running npm script 'test' in 'react-server'
Error: Command failed: /bin/sh -c npm run test
/Users/lida.wang/code/react-server/packages/react-server/node_modules/eslint/lib/eslint.js:646
                    throw new Error("Definition for rule '" + key + "' was not found.");
                    ^

Error: Definition for rule 'keyword-spacing' was not found.
    at /Users/lida.wang/code/react-server/packages/react-server/node_modules/eslint/lib/eslint.js:646:27
    at Array.forEach (native)
    at EventEmitter.module.exports.api.verify (/Users/lida.wang/code/react-server/packages/react-server/node_modules/eslint/lib/eslint.js:619:16)
    at processText (/Users/lida.wang/code/react-server/packages/react-server/node_modules/eslint/lib/cli-engine.js:201:27)
    at CLIEngine.executeOnText (/Users/lida.wang/code/react-server/packages/react-server/node_modules/eslint/lib/cli-engine.js:361:26)
    at verify (/Users/lida.wang/code/react-server/packages/react-server/node_modules/gulp-eslint/index.js:19:17)
    at Transform._transform (/Users/lida.wang/code/react-server/packages/react-server/node_modules/gulp-eslint/index.js:37:18)
    at Transform._read (_stream_transform.js:166:10)
    at Transform._write (_stream_transform.js:154:12)
    at doWrite (_stream_writable.js:292:12)
```

It also repros when I try to run `gulp test` in `/packages/react-server`

gigabo:
@lidawang - Sounds like your `gulp-eslint` is out of date.  Try an `npm update` in your repo root.

gigabo:
Oh, whoa, just read that again.  Looks like it's picking up an `eslint` in your `packages/react-server/node-modules`.

Might be easiest to start fresh (in your repo root):

```bash
for d in `ls -d packages/*/node_modules`; do echo $d; rm -r $d; done
npm run bootstrap
npm test
```

gigabo:
@lidawang - Given the test weirdness here, I'm tempted to say let's just ditch this and wait for #161. 😖 

That's going to obsolete it, anyway.

PR #154, opened 2016-05-13
Add debug script to react-server-test-pages

lidawang:
Let us use `npm run debug` in `react-server-test-pages`! I didn't know whether or not to pin a version on `node-inspector`, please let me know if I should and which one to pin to!

### DIFF
--- a/packages/react-server-test-pages/README.md
+++ b/packages/react-server-test-pages/README.md
@@ -19,4 +19,6 @@ Then hit http://localhost:3000/ to see what's available.
 If you make changes in `packages/react-server` you'll need to `CTRL-C` and
 re-run the last line of the setup to pull them in.
 
+If you wish to run the test server in debug mode, simply replace `npm start` with `npm run debug`.
+
 Add pages in `entrypoints.js`.  Instructions are at the top.
--- a/packages/react-server-test-pages/package.json
+++ b/packages/react-server-test-pages/package.json
@@ -6,7 +6,8 @@
   "scripts": {
     "prepublish": "gulp build",
     "start": "react-server-cli",
-    "test": "gulp test"
+    "test": "gulp test",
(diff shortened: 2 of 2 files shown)


packages/react-server-test-pages/package.json
```diff
@@ -19,5 +20,8 @@
     "react-server-cli": "^0.2.6",
     "react-server-data-bundle-cache": "^0.2.6",
     "superagent": "1.2.0"
+  },
+  "devDependencies": {
+    "node-inspector": ""
```
gigabo: No version?

gigabo:
This is awesome @lidawang!  Worth mentioning in `packages/react-server-test-pages/README.md`.

packages/react-server-test-pages/package.json
lidawang: Yea tbh I'm not sure which to pick! The latest stable version that I used for getting this to work is 0.12.8 - I wasn't sure if I should pin to that specific version or accept patches (~0.12.8?), etc. Any guidance?

gigabo: Generally the easiest way to add a dependency like this is to use:

```bash
npm install --save-dev node-inspector
```

That will install it and add it to the `devDependencies` in package.json for you, with an appropriate version.

packages/react-server-test-pages/README.md
```diff
@@ -19,4 +19,6 @@ Then hit http://localhost:3000/ to see what's available.
 If you make changes in `packages/react-server` you'll need to `CTRL-C` and
 re-run the last line of the setup to pull them in.
 
+If you wish to run the test server in debug mode, simply replace `npm start` with `npm run debug`.
+
```
gigabo: :metal:

PR #156, opened 2016-05-15
Links should transclude classes

doug-wade:
The `Link` component should pass classes attached to it the the underlying anchor tag.


doug-wade:
I got that transient `An error occurred running zombie tests TypeError: connect ECONNREFUSED 127.0.0.1:8769` again; kicking the build.

gigabo:
Good call!

PR #159, opened 2016-05-15
Add downloads badge

doug-wade:
We get a pleasantly surprisingly large number of downloads per month.  Let's advertise it

screen shot 2016-05-15 at 2 31 42 pm


gigabo:
@doug-wade - Look like this conflicts with your other recent badge PRs that just landed on master.

doug-wade:
rolled into #160

PR #168, opened 2016-05-18
Approximate above the fold paint time using synchronous script

TonyHYK:
It would be great to have a timing for when above the fold elements are parsed and painted on the client side. A proxy for this would be a synchronous script tag written to the client side right after above the fold elements are written. I'm also using `performance.mark` so that this time point shows up in WPT. In practice, this approximates "first paint" time to within 100ms!

This would really help us understand user-perceived performances on the client side and can help us approximate speed index.

The implementation of this "display" time is different from `displayElement.fromStart`, which might be confusing. There is a separate issue #169 to improve `displayElement` timings.

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -55,6 +55,7 @@ class ClientController extends EventEmitter {
 
 		window.__reactServerTimingStart = window.performance ? window.performance.timing.navigationStart : new Date;
 		var wakeTime = new Date - window.__reactServerTimingStart;
+		var aboveTheFold = window.performance ? (window.__displayAboveTheFold - window.__reactServerTimingStart) : 0;
 
 		var dehydratedState = window.__reactServerState;
 
@@ -82,6 +83,8 @@ class ClientController extends EventEmitter {
 
 		// Log this after loglevel is set.
 		logger.time('wakeFromStart', wakeTime);
+		// this is a proxy for when above the fold content gets painted (displayed) on the browser
+		logger.time('displayAboveTheFold.fromStart', aboveTheFold);
 	}
 
 	terminate() {
--- a/packages/react-server/core/renderMiddleware.js
+++ b/packages/react-server/core/renderMiddleware.js
@@ -834,6 +834,7 @@ function writeElements(res, elements) {
 
 		if (i === RLS().atfCount - 1){
 
+			logAboveTheFoldTime(res);
 			// Okay, we've sent all of our above-the-fold HTML,
 			// now we can let the client start waking nodes up.
 			bootstrapClient(res)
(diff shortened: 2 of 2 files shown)


packages/react-server/core/renderMiddleware.js
```diff
@@ -880,6 +881,13 @@ function writeElement(res, element, i){
 	}
 }
 
+function logAboveTheFoldTime(res) {
+	// write a synchronous script to record the time on the browser when above the fold content shows up
+	// this is a proxy for "first paint" when the DOM is parsed and painted
+	renderScriptsSync([{text:`__displayAboveTheFold=new Date;
+		window.performance && window.performance.mark && window.performance.mark("displayAboveTheFold.fromStart");`}], res);
```
gigabo: Whoa, this is cool!  Didn't even know about [`performance.mark`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark).

I appreciate this being broken onto two lines from a readability perspective, but because this is an es6 template string it will insert a literal newline and tabs into the output document.

This might be a good place for an old-fashioned string concatenation:
```javascript
[{text:'__displayAboveTheFold=new Date;'+
'window.performance && ...'}],

File: packages/react-server/core/ClientController.js [hunk 2]
```diff
@@ -55,6 +55,7 @@ class ClientController extends EventEmitter {
 
 		window.__reactServerTimingStart = window.performance ? window.performance.timing.navigationStart : new Date;
 		var wakeTime = new Date - window.__reactServerTimingStart;
+		var aboveTheFold = window.__displayAboveTheFold - window.__reactServerTimingStart;
```
gigabo: So, this could be negative if `!window.performance`?  Should we handle that case?

TonyHYK: yeah just as wakeTime falls back to a zero-ish value, I'll just handle the edge case as 0. Seems more sensible than some large negative number and easier to filter out.

packages/react-server/core/renderMiddleware.js
TonyHYK: Good catch!

TonyHYK:
Addressed feedbacks and squashed the changes.

packages/react-server/core/ClientController.js
```diff
@@ -55,6 +55,7 @@ class ClientController extends EventEmitter {
 
 		window.__reactServerTimingStart = window.performance ? window.performance.timing.navigationStart : new Date;
 		var wakeTime = new Date - window.__reactServerTimingStart;
+		var aboveTheFold = window.performance ? (window.__displayAboveTheFold - window.__reactServerTimingStart) : 0;
```
gigabo: So... this avoids _negative_ numbers, but it still reports _incorrect_ values.

Seems like if we don't have `window.performance` we don't have anything useful to report, so we just shouldn't report at all?  That skews results because some data points are omitted, but it may skew _by less_ than reporting _incorrect_ data points would.

TonyHYK: yeah... I feel like this is also true for any values logged in ClientController that relies on `__reactServerTimingStart` , do you think we want to change those also?

gigabo: Maybe worth filing a ticket to fix them all up and saying it's outside the scope of this PR.

gigabo: Done: #170.

gigabo:
This is awesome.  Thanks @TonyHYK!

packages/react-server/core/ClientController.js
TonyHYK: Agreed!

PR #174, opened 2016-05-18
Added work around for childProp update

bw070908:
This provides a workaround for the regression that causes a TypeError when updating childProps.

### DIFF
--- a/packages/react-server/core/components/RootElement.js
+++ b/packages/react-server/core/components/RootElement.js
@@ -27,15 +27,15 @@ class RootElement extends React.Component {
 				// Because props.childProps is set to writable/configurable:false
 				// We need to clone the entire props object and update childProps within the clone
 				var propstemp = {};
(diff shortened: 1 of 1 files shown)


packages/react-server/core/components/RootElement.js
```diff
@@ -23,8 +23,19 @@ class RootElement extends React.Component {
...
+				var propstemp = {};
+                for (var k in _this2.props) {
+                    if(k != 'childProps') {
+                        propstemp[k] = _this2.props[k];
+                    }
+                }
+                propstemp['childProps'] = childProps;
+                
+                _this2.props = propstemp;
+                _this2.forceUpdate();
```
gigabo: I think this needs to be `this`?  Looks like you may have been trying this out on the `target/`module and copied back to the source?

gigabo:
Closing this in favor of #177.

PR #177, opened 2016-05-19
Add workaround for update to childProps

bw070908:
This change provides workaround for updating childProps in RootElement.js. Previous regression has caused the logic to attempt to change childProps while it is writable/configurable: false.

### DIFF
--- a/packages/react-server/core/components/RootElement.js
+++ b/packages/react-server/core/components/RootElement.js
@@ -1,6 +1,10 @@
 var React = require('react');
 var Q = require('q');
 
+const _ = {
+	assign: require('lodash/assign'),
+};
+
 var logger = require('../logging').getLogger(__LOGGER__);
 
 class RootElement extends React.Component {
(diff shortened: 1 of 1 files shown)


packages/react-server/core/components/RootElement.js
```diff
@@ -23,7 +23,17 @@ class RootElement extends React.Component {
 
 				// Okay, now we've complained about it
 				// sufficiently, let's go ahead and update.
-				this.props.childProps = childProps;
+
+				// Because this.props.childProps is writable/configurable:false
+				// we update the entirety this.props through a clone
+				var propstemp = {};
+				for (var k in this.props) {
```
gigabo: Should probably either iterate over `Object.keys(this.props)` here or add a `this.hasOwnProperty(k)` check.

gigabo: Actually, this could all just be a one-liner with `_.assign`:

```javascript
this.props = _.assign({}, this.props, {childProps});
```

We're already using the `assign` method from lodash elsewhere, so this won't increase our bundle size.

Our pattern for including lodash methods is:

```javascript
const _ = {
        assign: require('lodash/assign'),
};
```

gigabo:
Thanks for tracking this down @bw070908!

Issue #161, opened 2016-05-16
Better API for declaring the fold: <TheFold />


bartkusa:
Instead of saying "the fold for this page is after the first N elements" (which could change based on middlewares? maybe?), it'd be nice if you could just declare it inline with everything else in JSX. Eg:

```
getElements() {
	var prefetchConfig = {prefetchUrls:this._getPrefetchURLS()};
	
	return [
		RootElements.createRootElementWhenResolved(this._emailVerificationBannerStore, <EmailVerificationBanner />),
		RootElements.createRootElementWhenResolved(this._searchAndMerchStore, <SearchAndMerch />),
		<TheFold />
		RootElements.createRootElementWhenResolved(this._listingFeeStore, <MarketingSections />),
		RootElements.createRootElementWhenResolved(this._generalInquiryStore, <GeneralInquiryForm />),
		RootElements.createRootElementWhenResolved(this._eigencitiesStore, <EigencitiesSection />),
		<Prefetch {...prefetchConfig}/>,
	];
}
```

(Obviously it doesn't need to _really_ be React or JSX. It could just be `require("react-server/TheFold")` or something.)

gigabo:
@bartkusa - That's an awesome idea!  _SO_ much better than the current API!

lidawang:
This change would make life sooo much easier for the DP!

PR #178, opened 2016-05-19
Add test case for getAboveTheFold with RootContainers

lidawang:
This is the test page I wrote to play with `getAboveTheFoldCount` - it's currently not going to show the right result, because I implemented it to do what I wanted it to do (not count `RootContainer` tags towards the `getAboveTheFoldCount`) - lemme know if I oughta change the text to have the reality match the expectations


gigabo:
Awesome, thanks @lidawang!  This will help with work on #161.

PR #179, opened 2016-05-19
Don't run tests in prepublish

gigabo:
Takes too long.  We'd already ditched this for other packages, but one snuck
in with the yeoman generator.


gigabo:
Gonna merge this so I can get `0.2.10` out the door.

PR #180, opened 2016-05-19
Generator readme

gigabo:
Clean up badges on the generator readme and add a link to it from the top-level readme.


doug-wade:
👍  thanks!

PR #181, opened 2016-05-20
Update versions in generated package.json

gigabo:
- `react-server` currently requires an _exact_ version of superagent.
- Update the `react-server-*` deps while I'm at it.


doug-wade:
👍

PR #183, opened 2016-05-20
Update Example App remote url to one which works

conorcussell:
the url http://redfin/react-server.git doesn't seem to be the correct url to clone so i've updated it.


gigabo:
Yikes!  Thanks @conorcussell!

PR #186, opened 2016-05-20
Don't run bower as part of the generator

doug-wade:
From https://www.reddit.com/r/reactjs/comments/4k5uer/react_server_blazing_fast_page_load_and_seamless/d3cpxuy

### DIFF
--- a/packages/generator-react-server/generators/app/index.js
+++ b/packages/generator-react-server/generators/app/index.js
@@ -51,6 +51,6 @@ module.exports = yeoman.Base.extend({
 	},
 
 	install: function () {
-		this.installDependencies();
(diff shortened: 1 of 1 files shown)


gigabo:
`installDependencies` and `npmInstall` are provided by `yeoman`?

What is the difference between these two?  They're interchangeable?

doug-wade:
http://yeoman.io/authoring/dependencies.html

`generator.bowerInstall()` -- install bower deps
`generator.npmInstall()` -- install npm deps
`generator.installDependencies()` -- install bower + npm deps

gigabo:
Got it.  Thanks!

Issue #189, opened 2016-05-20
Client transitions don't handle data request errors well


nampas:
Repro:
1) Go to Page A
2) Click on a link to execute a client transition to Page B. Page B should make a data request that 500's
3) Attempt a client transition back to Page A

Result:
Get the following console error, and navigation fails:
`error: [react-server.core.ClientController] onNavigate error Object {status: 404, message: "No Route!"}`

Expected Result:
Transitioning away from a page that had a data request error should succeed

Note that this repros when `bundleData` is on or off

gigabo:
Whoa... that's not good.  Thanks for reporting @nampas.

gigabo:
Hi @nampas.  I'm having trouble reproducing this.

Please have a look at #197 and see if anything jumps out at you.

nampas:
Hrm ok thanks for throwing that example together, I'll take a look and get back to you.

nampas:
Ok, so it turns out that we had a RootElement that didn't resolve in the case of a data request failing. That made react-server think that page B never finished rendering (`Navigator.finishRoute` was never invoked). Then, when we tried to execute the transition back to page A, `Navigator.startRoute` wouldn't emit `navigateStart` because it was waiting for the current page load to finish.

Closing this ticket, but it might be worth adding a timeout for finishRoute()

PR #197, opened 2016-05-20
Add a test endpoint for data errors

gigabo:
Trying to reproduce the behavior described in #189.

@nampas - I haven't been able to repro.  I tried:

1. An endpoint that returns 500
2. A response handler that throws an exception

Neither produces the behavior described in #189.  Can you see anything different about this setup and your repro environment that could help me trigger the bug here?


doug-wade:
👍

gigabo:
This turns out not to have been a bug.  The tests are still useful, though, so going to merge.

Issue #147, opened 2016-05-12
Automated tests against `react-server-test-pages` server


gigabo:
The `react-server-test-pages` server is great for testing by hand.  It's easy to build simple test pages to play with and to iterate quickly.  It would be great to take some of the test cases we exercise by _hand_ during development there and automate them for regression tests.

Would selenium work for this?  What's an easy way to set this up?

PR #198, opened 2016-05-20
Fix error object normalization in logger interceptor

gigabo:
Error messages and stack traces were missing from logs on the server.

Also added a test page that just generates an error in the logs.

### DIFF
--- a/packages/react-server/core/logging/server.js
+++ b/packages/react-server/core/logging/server.js
@@ -2,7 +2,7 @@ var winston = require('winston')
 ,   common  = require('./common')
 ,   _ = {
 	mapValues     : require("lodash/mapValues"),
-	pick          : require("lodash/pick"),
+	pickBy        : require("lodash/pickBy"),
 	isPlainObject : require("lodash/isPlainObject"),
 	isEmpty       : require("lodash/isEmpty"),
 	trimStart     : require("lodash/trimStart"),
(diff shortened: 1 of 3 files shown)


doug-wade:
👍  I dig all this test infrastructure!

packages/react-server/core/logging/server.js
```diff
@@ -68,7 +68,7 @@ function errorInterceptor (level, msg, meta) {
 // massage the error into a format suitable for logging
 function normalizeError (err) {
 	if (err instanceof Error) {
-		return _.pick({
+		return _.filter({
```
roblg: This is... confusing to me. This changes the return value from an object w/ `message` and `stack` keys to an array, right? That seems weird, given that later in this method we return an object. 

I'm also interested to know when this stopped working, because I'm _pretty_ sure (though not 100%) that it worked at some point.

gigabo: Nice catch @roblg.  Definitely want to keep this an object.

Looks like this broke during [the lodash upgrade](https://github.com/redfin/react-server/pull/137).  `_.pick` used to take a key array _or_ a predicate, but the predicate variant [got split out into `_.pickBy` in v4.0.0](https://github.com/lodash/lodash/wiki/Changelog#v400):

> Split _.omit & _.pick into _.omitBy & _.pickBy

Will update to use `_.pickBy`.  Thanks!

gigabo:
> I dig all this test infrastructure!

Yeah, but we _really_ need to [automate it](https://github.com/redfin/react-server/issues/147) to stave off regressions.

gigabo:
Gonna merge this.  Need stack traces for server-side exceptions!

PR #200, opened 2016-05-21
Add docker to yeoman generator

withinboredom:
This PR adds a simple `docker-compose.yml` file and a simple `Dockerfile` to the yeoman generator to allow people to quickly get up and running with Docker. This will be helpful for developers using Windows.

PR #206, opened 2016-05-24
Default options for cli overwrite config file options

withinboredom:
While working #200, I discovered the config files (Specifically the `host` option) were not being set. This is due to parseCliArgs.js having a "default" which comes later in the array which is passed to merge (in mergeOptions.js). This causes the cli default to always override the config file in the case of routes and host.

I've removed the defaults from parseCliArgs.js and moved the routes default to defaultOptions.js. I think that this makes the most sense for default options to be set and is always overwritten by the cli or config file. A better long term solution might be to make the "default" key in the "parseCliArgs.js" file to merely be for display.

### DIFF
--- a/packages/react-server-cli/src/parseCliArgs.js
+++ b/packages/react-server-cli/src/parseCliArgs.js
@@ -4,7 +4,6 @@ export default (args = process.argv) => {
 	var parsedArgs = yargs(args)
 		.usage('Usage: $0 [options]')
 		.option("routes", {
-			default: "./routes.js",
 			describe: "The routes file to load.",
 		})
 		.option("p", {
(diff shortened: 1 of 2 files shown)


packages/react-server-cli/src/parseCliArgs.js
```diff
@@ -13,7 +12,6 @@ export default (args = process.argv) => {
 			type: "number",
 		})
 		.option("host", {
-			default: "localhost",
```
gigabo: This is safe to just omit because it's defaulted somewhere deeper than `defaultOptions.js`?

withinboredom: It's already the default in defaultOptions.js

gigabo: image

gigabo:
Nice fix @withinboredom!

PR #207, opened 2016-05-24
Better Generator Tests

doug-wade:
Previously, we were checking if the created files were created, but not in the temp directory we were creating them in.  Now, we check that they exist where they actually are!  Unblocks #200


gigabo:
Thanks for fixing that up @doug-wade! 👍

PR #208, opened 2016-05-24
Compress pngs

doug-wade:
Our pngs were looking a little chunky; maybe we should use svgs @careylin ?

Now about a third the size they were before
```
dougwade code/react-server ‹compress-pngs*› » git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
dougwade code/react-server ‹master*› » du -sh images
104K	images
dougwade code/react-server ‹master*› » git checkout compress-pngs
Switched to branch 'compress-pngs'
dougwade code/react-server ‹compress-pngs*› » du -sh images
36K	images
```


gigabo:
:squint: 👍

PR #211, opened 2016-05-25
Fix #103: log NODE_ENV on startup

### LINKED ISSUE #103: Request: Startup log prints current NODE_ENV
When starting the server (particularly for the first time) We're not really explicit as to what NODE_ENV was found.  We mention in the docs that it defaults to `staging` but we don't actually log out something along the lines of `NODE_ENV defaulting to 'staging'` 

Just a little friendlier than `2016-04-19T22:00:45.054Z - warning: [react-server-cli.src.index] -- NODE_ENV is not set to "production".`

doug-wade:
Log `NODE_ENV` during server startup.  Fixes #103

### DIFF
--- a/packages/react-server-cli/src/startServer.js
+++ b/packages/react-server-cli/src/startServer.js
@@ -309,10 +309,7 @@ const logProductionWarnings = ({hot, minify, jsUrl, longTermCaching}) => {
 		if (!longTermCaching) {
 			logger.warning("-- Long-term caching is disabled. To enable, set longTermCaching to true (--long-term-caching at the command-line) or set NODE_ENV=production to turn on.");
 		}
-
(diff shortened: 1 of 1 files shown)


packages/react-server-cli/src/startServer.js
```diff
@@ -313,6 +313,7 @@ const logProductionWarnings = ({hot, minify, jsUrl, longTermCaching}) => {
 		if (process.env.NODE_ENV !== "production") { //eslint-disable-line no-process-env
 			logger.warning("-- NODE_ENV is not set to \"production\".");
 		}
+		logger.info(`NODE_ENV is set to ${process.env.NODE_ENV}`);
```
roblg: Does it make sense to keep the `not production` warning above if we're adding this?

roblg:
Change lgtm

PR #212, opened 2016-05-25
Fix #37: Add README for react-server-gulp-module-tagger

### LINKED ISSUE #37: Write README for react-server-gulp-module-tagger
currently, on [npmjs.com](https://www.npmjs.com/package/react-server-gulp-module-tagger), there is no information about what react-server-gulp-module-tagger is

doug-wade:
Fix #37

### DIFF
--- a/packages/react-server-gulp-module-tagger/README.md
+++ b/packages/react-server-gulp-module-tagger/README.md
@@ -1,0 +1,64 @@
+# react-server-gulp-module-tagger
+
+A [gulp](http://gulpjs.com) plugin for tagging [react-server](https://www.npmjs.com/package/react-server) logger instances with information about the module they're being used in.
+
+To transpile your source for use with [React Server](https://www.npmjs.com/package/react-server), install gulp and the plugin
+
+```shell
+npm i -D gulp react-server-gulp-module-tagger
+```
+
+Then add the task to your gulpfile
+
+```javascript
+const gulp = require('gulp');
+const tagger = require('react-server-gulp-module-tagger');
+gulp.task('compile', () => {
+    gulp.src('src')
+        .pipe(tagger())
+        .pipe(gulp.dest('dist'));
+});
+```
+
+A compile task might also use [Babel](https://babeljs.io) with the [React Server Babel preset](https://www.npmjs.com/package/babel-preset-react-server) to transpile jsx and
+es 7 for the browser and the server
+
+```javascript
+const gulp = require('gulp');
+const babel = require('gulp-babel');
+const tagger = require('react-server-gulp-module-tagger');
+
+gulp.task('compile', () => {
+    gulp.src('src')
+        .pipe(tagger(
+            trim: 'src.'
+        ))
+        .pipe(babel({ presets: ['react-server'] }))
+        .pipe(gulp.dest('dist'));
+});
+```
+
+Given a [`getLogger`](http://redfin.github.io/react-server/annotated-src/logging) call,
+adds the correct arguments to keep the server and the browser in sync.
+
+For example, given a module in `src/components/my-feature/foo.js`, and using the options
+`{ trim: 'src.' }`
+
+```javascript
+let logger = require("react-server").logging.getLogger(__LOGGER__);
+```
+
+returns a logger instance that will have consistent coloring on the server and
+the client, and that has a human-friendly, readable name that easily maps to
+the file tree (in this example `components.my-feature.foo`).
+
+If you need more than one logger in your module, you can distinguish them
+with labels
+
+```javascript
+var fooLogger = logging.getLogger(__LOGGER__({ label: "foo" }));
+var barLogger = logging.getLogger(__LOGGER__({ label: "bar" }));
+```
+
+Two other tokens, `__CHANNEL__` and `__CACHE__`, are reserved for future use,
+and will also be replaced with a module context.


packages/react-server-gulp-module-tagger/README.md
```diff
@@ -0,0 +1,45 @@
+# react-server-gulp-module-tagger
+
+A [gulp](http://gulpjs.com) plugin for tagging [react-server](https://www.npmjs.com/package/react-server) logger instances with information about the module they're being used in.
+
+To transpile your React for use with [React Server](https://www.npmjs.com/package/react-server), install gulp and the plugin
```
roblg: "transpile your React" confused me here for a sec. Maybe "transpile your JS" instead?

packages/react-server-gulp-module-tagger/README.md
```diff
@@ -0,0 +1,45 @@
...
+For example, given the following three files
+
+#### src/foo.js
+    const logger = require("react-server").logging.getLogger(__CACHE__);
+
+#### src/bar.js
+    let logger = require("react-server").logging.getLogger(__LOGGER__);
+
+#### src/baz.js
+    var logger = require("react-server").logging.getLogger(__CHANNEL__);
```
roblg: `__CACHE__` and `__CHANNEL__` seem like funny examples to use here w/ a logger. They plop in the same info, but they're typically used in different contexts. (Which leads me to believe maybe our replacement string should be `__MODULE_CTX__` or something)

doug-wade: As far as I can tell, we support `__CACHE__` and `__CHANNEL__` for historical, internal uses, like our ipc for our cluster.  I'm not sure if we plan to release them, but I gather we want to support arbitrary tokens, per #12.  

I would prefer to document `__CACHE` and `__CHANNEL__` though, in the meantime, since it would be pretty mystifying if someone were to stumble across them; I think my question is how.  My first intuition was to give an example using `__CACHE__` and `__CHANNEL__` with a destructuring assignment

```
let {name, color} = __CACHE__;
let {name, color: {server, client}} = __CHANNEL__;
```

but I'm not sure if we want to make the structure of the tags part of our public api.  The other option (if we decide to keep the api small by only documenting that "a module context is substituted") would be to document `__CACHE__` and `__CHANNEL__` as "reserved for future functionality", with the assumption that caching and clustering are forthcoming.  Which do you think is a better approach?

roblg: I think I like the "reserved for future functionality" approach for now. It keeps the documentation for the common case (`__LOGGER__`) cleaner, and we don't have to expose what that module context is yet. Eventually, I can imagine documenting the exposed module context and allowing extensions via the tagger, particularly if we're allowing arbitrary tags -- the cache may not need `color`, but the logger certainly does.

packages/react-server-gulp-module-tagger/README.md
```diff
@@ -0,0 +1,45 @@
...
+        gulp.src('src')
+            .pipe(tagger())
+            .pipe(babel({ presets: ['react-server'] }))
+            .pipe(gulp.dest('dist'));
+    });
+
+Given a [React Server](https://www.npmjs.com/package/react-server) getLogger call,
+adds the correct arguments to keep the server and the browser in sync.
+
+For example, given a module in `src/components/my-feature/foo.js`
```
gigabo: Link to the source here?

packages/react-server-gulp-module-tagger/README.md
```diff
@@ -0,0 +1,45 @@
...
+    const gulp = require('gulp');
+    const babel = require('gulp-babel');
+    const tagger = require('react-server-gulp-module-tagger');
+
+    gulp.task('compile', () => {
+        gulp.src('src')
+            .pipe(tagger())
+            .pipe(babel({ presets: ['react-server'] }))
+            .pipe(gulp.dest('dist'));
+    });
```
gigabo: Would benefit from syntax highlighting.

```javascript
const gulp = require('gulp');
const tagger = require('react-server-gulp-module-tagger');
gulp.task('compile', () => {
    gulp.src('src')
        .pipe(tagger())
        .pipe(gulp.dest('dist'));
});
```

packages/react-server-gulp-module-tagger/README.md
```diff
@@ -0,0 +1,45 @@
...
+Given a [React Server](https://www.npmjs.com/package/react-server) getLogger call,
+adds the correct arguments to keep the server and the browser in sync.
+
+For example, given a module in `src/components/my-feature/foo.js`
+
+    let logger = require("react-server").logging.getLogger(__LOGGER__);
+
+returns a logger instance that will have consistent coloring on the server and
+the client, and that has a human-friendly, readable name that easily maps to
+the file tree (`src.components.my-feature.foo`).
```
gigabo: Would be nice to also mention the `trim` option as a way to strip that `src.` from the module name.

packages/react-server-gulp-module-tagger/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```javascript
+const gulp = require('gulp');
+const babel = require('gulp-babel');
+const tagger = require('react-server-gulp-module-tagger');
+
+gulp.task('compile', () => {
+    gulp.src('src')
+        .pipe(tagger(
+            trim: 'src'
```
gigabo: I think this would actually need to be 'src.' to trim the dot, too.

gigabo:
This looks good.  Thanks for adding it @doug-wade! 👍

Issue #209, opened 2016-05-24
Add docs for the logger


doug-wade:
The logger is not particularly simple to use, and afaict, there isn't anywhere its really documented.  The best guide we have is the [annotated source](http://redfin.github.io/react-server/annotated-src/logging), but that [isn't discoverable as-is](https://github.com/redfin/react-server/issues/195).  We should add a doc to the docs folder on logging, add a link to it from the writing a page guide, and link to the (better formatted) annotated source code.

PR #213, opened 2016-05-25
Add logger to yeoman generator

doug-wade:
Demonstrate how to use the logger.  Addresses #209

### DIFF
--- a/packages/generator-react-server/generators/app/templates/hello-world.js
+++ b/packages/generator-react-server/generators/app/templates/hello-world.js
@@ -1,10 +1,14 @@
 import React from 'react';
+import {logging} from 'react-server';
+
+const logger = logging.getLogger(__LOGGER__);
 
 export default class HelloWorld extends React.Component {
 	constructor(props) {
 		super(props);
 		this.state = {exclamationCount: 0};
 		this.handleClick = () => {
+			logger.info(`Getting more excited! previously ${this.state.exclamationCount} excitements.`);
 			this.setState({exclamationCount: this.state.exclamationCount + 1});
 		};
 	}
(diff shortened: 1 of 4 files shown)


packages/generator-react-server/generators/app/templates/hello-world.js
```diff
@@ -1,10 +1,14 @@
 import React from 'react';
+import rs from 'react-server';
```
gigabo: I've been leaning toward this convention:

```javascript
import {logging} from 'react-server';
```

Thoughts?

packages/generator-react-server/package.json
```diff
@@ -44,6 +44,7 @@
     ],
     "envs": [
       "node"
-    ]
+    ],
+    "globals": ["__LOGGER__", "__CACHE", "__CHANNEL__"]
```
gigabo: Why the other globals?  They're not used, right?  Just `__LOGGER__`?

packages/generator-react-server/generators/app/templates/package.json
```diff
@@ -28,6 +28,7 @@
   "xo": {
     "esnext": true,
     "extends": "xo-react",
-    "ignores": ["__clientTemp/**/*"]
+    "ignores": ["__clientTemp/**/*"],
+    "globals": ["__LOGGER__", "__CACHE", "__CHANNEL__"]
```
gigabo: If this were alphabetized it would be pure addition. 👹

gigabo:
Looks good.  Thanks @doug-wade. 👍

doug-wade:
Here's what it looks like
screen shot 2016-05-25 at 5 42 25 pm

Issue #127, opened 2016-05-04
reuseDom Client Transition error


nampas:
I ran into an issue today with reuseDom, when the destination page has fewer root elements than the source page.

Let's say Page A has 3 root elements, Page B has 1 root element.
When transitioning from from page A to page B via a Client Transition with reuseDom, the 1st root element will get correctly replaced by Page B, but the 2nd and 3rd root elements will remain unchanged (minus CSS) from Page A.

reuseDom needs to wipe out extra root elements when transitioning to a page with fewer Root Elements. 

Unfortunately I don't have a good, non-sensitive test case for this.

gigabo:
Nice catch @nampas!  Thanks for reporting!

PR #219, opened 2016-05-25
Trim dangling root nodes with `reuseDom`

gigabo:
A `reuseDom` transition to a page with fewer root elements leaves root elements from the previous page hanging out in the DOM.  This patch cleans them up.

Addresses #127.


doug-wade:
lgtm 👍

PR #221, opened 2016-05-26
Add new component: `<TheFold />`

### LINKED ISSUE #161: Better API for declaring the fold: <TheFold />

gigabo:
This replaces the `getAboveTheFoldCount()` page method, which is awful:

- It's separate from the elements themselves, so even remembering to update it
  can be a challenge.
- If the number of root elements above the fold varies, then duplicative logic
  may be required to determine the count.
- Middleware that inserts elements must remember to _increment_ it.
- It has non-obvious behavior when root containers are used above the fold (#144)

This closes #161 where @bartkusa proposed this _much_ nicer interface.

This is a breaking change.  The _default_ was previously to wake up the client
controller after the _first_ element.  With this change the client controller
isn't bootstrapped until either `<TheFold />` is seen, or we're out of
elements.  So, if unspecified, the entire page is assumed to be above the
fold.

### DIFF
--- a/docs/understanding-rendering.md
+++ b/docs/understanding-rendering.md
@@ -28,6 +28,7 @@ getElements() {
 				<Sidebar  />
 			</RootElement>
 		</RootContainer>
+		<TheFold />
 		<Footer />
 	</RootContainer>
 }
@@ -44,6 +45,10 @@ In the example above:
   the all objects from `bodyEmitter` and the resolution of `sidebarPromise`.
   Client-side it will re-render with updated props from `bodyEmitter` if it
   fires after the initial render.
+- `<TheFold />` will cause an inline `<script>` tag to be sent that kicks off
+  the client-side render, making elements above the fold interactive.  This
+  forces a browser paint, so it's important to put it after above-the-fold
+  elements.
 - `<Footer />` Will render immediately, and will be sent to the browser as
   soon as all elements before it have rendered and been sent.  It won't
   receive any props.
@@ -55,7 +60,7 @@ In the example above:
 - If an element is blocking already-rendered elements after it, when it
   renders the entire block of elements will be sent in a single `write` to the
   response socket.
-- After the above-the-fold elements (as specified by `getAboveTheFoldCount()`)
+- After the above-the-fold elements (followed by `<TheFold/>`)
   have been sent an inline `<script>` is sent that instantiates the
   `ClientController` in the browser and gives it the data bundle for all
   requests that have resolved.  It then renders all elements that have already
--- a/packages/react-server-test-pages/pages/data/delay.js
+++ b/packages/react-server-test-pages/pages/data/delay.js
@@ -1,4 +1,7 @@
 import Q from "q";
+import _ from "lodash";
+
+const BIG = n => _.range(+n).reduce((m, v) => (m['element_'+v] = v, m), {})
 
 export default class DelayDataPage {
 	setConfigValues() {
@@ -8,8 +11,8 @@ export default class DelayDataPage {
 		return 'application/json';
 	}
 	getResponseData() {
-		const { ms, val } = this.getRequest().getQuery();
+		const { ms, val, big } = this.getRequest().getQuery();
 		return Q.delay(ms||0)
-			.then(() => val||JSON.stringify({'ok':true}));
+			.then(() => val||JSON.stringify(big?BIG(big):{'ok':true}));
 	}
 }
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -497,11 +497,6 @@ class ClientController extends EventEmitter {
 		// each request so we can keep track of that overhead.
 		var totalRenderTime = 0;
 
-		// We'll consider the page "interactive" when we've rendered
-		// all elements that we expect to be above the fold.
-		// Need this to be an integer value greater than zero.
-		var aboveTheFoldCount = Math.max(page.getAboveTheFoldCount()|0, 1)
-
 		// These resolve with React elements when their data
 		// dependencies are fulfilled.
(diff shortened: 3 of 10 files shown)


packages/react-server-test-pages/pages/data/delay.js
```diff
@@ -8,8 +11,8 @@ export default class DelayDataPage {
 		return 'application/json';
 	}
 	getResponseData() {
-		const { ms, val } = this.getRequest().getQuery();
+		const { ms, val, big } = this.getRequest().getQuery();
```
gigabo: Added a `big=<size>` param to the test data endpoint that just tells it to return a JSON object with `size` keys.  It's easiest to see the fold when there's a bunch of JS to for the browser to slurp up right after.

packages/react-server/core/ClientController.js
```diff
@@ -613,6 +608,11 @@ class ClientController extends EventEmitter {
 
 			if (element.containerOpen || element.containerClose){
 				return; // Nothing left to do.
+			} else if (element.isTheFold) {
+				logger.time(`renderAboveTheFold.fromStart`, new Date - tStart);
+				logger.time(`renderAboveTheFold.individual`, totalRenderTime);
+				logger.time(`renderAboveTheFold.elementCount`, index + 1);
```
gigabo: Another breaking change: Unless a page includes `<TheFold />` there are now no `renderAboveTheFold` stats.

packages/react-server/core/common.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
 	RootContainer: require("./components/RootContainer"),
 	RootElement: require("./components/RootElement"),
 	Link: require('./components/Link'),
+	TheFold: require('./components/TheFold').default,
```
gigabo: Awkward to pull out `.default` here, but IMO worth it for the cleaner implementation in `TheFold.js`.

packages/react-server/core/util/PageUtil.js
```diff
@@ -61,7 +62,6 @@ var PAGE_METHODS = {
 	getBase            : [() => null, Q],
 	getBodyClasses     : [() => [], Q],
 	getElements        : [() => [], standardizeElements],
-	getAboveTheFoldCount : [() => 1, _ => _],
```
gigabo: Glad I never updated the tabling around this.  It's almost like I knew it was an ugly hack that should be replaced... 👹

packages/react-server/core/util/PageUtil.js
```diff
@@ -174,6 +174,7 @@ function standardizeElements(elements) {
 		.makeArray(elements)
 		.map(e => isRootContainer(e)?flattenForRender(e):e)
 		.reduce((m, e) => m.concat(Array.isArray(e)?e:[e]), [])
+		.map(e => isTheFold(e)?markTheFold():e)
```
roblg: If I understand correctly, this is just converting a React element to an object that looks like `{isTheFold: true}` for the purposes of switching on it during render, right?

roblg:
Looks reasonable to me. This is _way_ nicer than `getAboveTheFoldCount()`. Three cheers for @bartkusa!

packages/react-server/core/util/PageUtil.js
gigabo: Yep, exactly.  The normalized elements array contains only root elements and control objects.

This introduces a new control object: `{isTheFold: true}`.

packages/react-server/core/components/TheFold.js
```diff
@@ -0,0 +1,19 @@
+import {Component} from "react";
+
+export default class TheFold extends Component {
+	render() {
+		throw new Error("Something went wrong.  Trying to render the fold...");
+	}
+}
+
+TheFold.defaultProps = {
```
doug-wade:
```
const isTheFold = Symbol();
TheFold.defaultProps = { [isTheFold]: true };
export function isTheFold(e) => e && e.props && e[isTheFold];
```
is, I think, privater

gigabo: Yeah, I actually thought about that.  AFAIK the reason `React` does that is to avoid injection attacks via JSON.  I don't see that as a concern here, so I opted to just keep it simple.

doug-wade:
👍

gigabo:
Note that this is merging into `next`.  Once I merge this I'll rebase `next` onto master and look at publishing a _prerelease_ of `0.3.0`.

docs/understanding-rendering.md
```diff
@@ -28,6 +28,7 @@ getElements() {
 				<Sidebar  />
 			</RootElement>
 		</RootContainer>
+		<TheFold />
```
bartkusa: What happens if some fool tries to stick `<TheFold />` in a `RootContainer`, or use it in a plain React component?

PR #222, opened 2016-05-26
Add bike share example

### LINKED ISSUE #191: Include examples of fetching data
As-is, `ReactServerAgent` is mostly undocumented.  We may want to borrow @aickin's [example](https://github.com/aickin/react-server-examples/blob/master/simple/AjaxPage.js) and stick it somewhere in the docs.

doug-wade:
Add an example application to highlight frameback, using [city bikes](http://api.citybik.es/v2/).  Fixes #191

### DIFF
--- a/packages/react-server-examples/bike-share/README.md
+++ b/packages/react-server-examples/bike-share/README.md
@@ -1,0 +1,82 @@
+# react-server-examples/bike-share
+
+An example project for `react-server` which demos server rendering, interactivity
+on the client side, frameback, ReactServerAgent, url parameters and logging.
+Uses [api.citybik.es](http://api.citybik.es/v2/) to get data about bike shares
+and their availability around the world.
+
+To start in development mode:
+
+```shell
+npm start
+```
+
+Then go to [localhost:3000](http://localhost:3000/). You will see an index page
+that shows the covered bike share networks around the world.  Each bike share
+network is a link to a details page that, when clicked, loads an iframe in front
+of the index page containing a network page for that bike share network,
+including information about each station in that network, and the number of
+available bikes the last time that data is available for.
+
+If you want to optimize the client code at the expense of startup time, type
+`NODE_ENV=production npm start`. You can also use
+[any react-server-cli arguments](../../react-server-cli#setting-options-manually)
+after `--`. For example:
+
+```shell
+# start in dev mode on port 4000
+npm start -- --port=4000
+```
+
+# Developing using Docker and Docker Compose
+
+These steps assume you are familiar with docker and already have it installed.
+Some basics:
+
+1. Download [Docker Toolbox](https://www.docker.com/products/docker-toolbox) and
+    install it.
+2. Start `docker quick start shell`
+3. Navigate to where you generated the project
+4. Add a configuration to set the `host` option to the ip given by
+    `docker-machine ip`. An example configuration might be like:
+```json
+{
+  "port": "3000",
+  "env": {
+    "docker": {
+      "host": "Your ip from `docker-machine ip` here"
+    },
+    "staging": {
+      "port": "3000"
+    },
+    "production": {
+      "port": "80"
+    }
+  }
+}
+```
+5. Now that your system is ready to go, start the containers:
+```shell
+docker-compose build --pull
+docker-compose up
+```
+
+The containers will now be running. At any time, press ctrl+c to stop them.
+
+To clean up, run the following commands:
+
+```shell
+docker-compose stop
+docker-compose rm --all
+docker volume ls # and get the name of the volume ending in react_server_node_modules
+# this name will be different depending on the name of the project
+docker volume rm _react_server_node_modules
+```
+
+The configuration included stores the node_modules directory in a "named volume".
+This is a special persistent data-store that Docker uses to keep around the
+node_modules directory so that they don't have to be built on each run of the
+container. If you need to get into the container in order to investigate what
+is in the volume, you can run `docker-compose exec react_server bash` which will
+open a shell in the container. Be aware that the exec functionality doesn't
+exist in Windows (as of this writing).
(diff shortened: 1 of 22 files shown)


packages/react-server-examples/bike-share/README.md
```diff
@@ -0,0 +1,65 @@
+# react-server-examples/hello-world
+
+An example project for `react-server` which only server rendering, interactivity on the client side, frameback, ReactServerAgent, url parameters and logging.  Uses [api.citybik.es](http://api.citybik.es/v2/) to get data about bike shares and their availability around the world.
```
gigabo: Looks like there may be a word missing here? "which only server rendering".

Also, please break these long lines up.  Long lines are problematic for [a number of reasons](https://github.com/redfin/react-server/pull/108).

packages/react-server-examples/bike-share/README.md
```diff
@@ -0,0 +1,65 @@
+# react-server-examples/hello-world
```
gigabo: s/hello-world/bike-share

packages/react-server-examples/bike-share/pages/index.js
```diff
@@ -0,0 +1,29 @@
...
+import Footer from '../components/footer';
+
+export default class IndexPage {
+	handleRoute(next) {
+		this.data = ReactServerAgent.get('http://api.citybik.es/v2/networks');
+		return next();
+	}
+	getElements() {
+		return [
+			<RootContainer key={0}>
```
gigabo: What's this key?

Also, doesn't seem like there's any need for a `<RootContainer>` here...?

packages/react-server-examples/bike-share/pages/index.js
```diff
@@ -0,0 +1,29 @@
...
+			</RootElement>,
+			<RootContainer key={2}>
+				<Footer/>
+			</RootContainer>
+		];
+	}
+
+	getAboveTheFoldCount() {
+		return 2;
+	}
```
gigabo: Soon to be out-of-date. 😛

gigabo:
@doug-wade - This is awesome!  What a great example!

packages/react-server-examples/bike-share/components/network-card.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import {logging, Link} from 'react-server';
+
+const logger = logging.getLogger(__LOGGER__);
+
+const NetworkCard = ({network}) => {
+	logger.info(`rendering card for network ${JSON.stringify(network)}`);
+	return (
+		<div><Link path={`/network?network=${network.id}`} frameback reuseFrame>{network.name}</Link> in {network.location.city}</div>
```
gigabo: Does this `reuseFrame` actually trigger?  I suspect it might need a truthy value.

gigabo:
So, this is cool.  It's a good example of the stated use-case for `frameback`.  _But_... in order to actually demonstrate the benefit the list page needs to be _slow_.  I haven't pulled this to try it out, yet, but something to keep in mind.  If the list page isn't _slow_ then you're better off with normal client transitions.

gigabo:
@doug-wade - Hit an error trying to get this running:

image

packages/react-server-examples/bike-share/pages/index.js
doug-wade: I added the key because of [this eslint rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md) that comes from [xo-react](https://github.com/sindresorhus/eslint-config-xo-react).  If I understand correctly, its to address [this  jsx gotcha](https://facebook.github.io/react/docs/jsx-gotchas.html#html-entities).  I believe it addresses the `Each child in an array should have a unique "key" prop.` message I see so often in our React Server implementation.

packages/react-server-examples/bike-share/README.md
doug-wade: See #230

packages/react-server-examples/bike-share/components/network-card.js
doug-wade: [jsx-boolean-value](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md) errors when I provide a truthy value; iiuc, `={true}` is no different than providing attribute alone.

doug-wade:
@gigabo The page loads in between 2.17s and 2.24s in my testing, which is slower than the recommended [2 second page load time](https://blog.kissmetrics.com/loading-time/)

doug-wade:
Listing page
screen shot 2016-05-29 at 11 55 42 pm

Details page
screen shot 2016-05-29 at 11 56 19 pm

doug-wade:
@gigabo the revision I just pushed should resolve your `npm i` problem.

gigabo:
@doug-wade - Just tried this out.  Looks like it's blocked by #228?  Details pages are failing to load for me, and that error's appearing in the log.

gigabo:
Oh, no, I see.  It's actually the _second_ details page that fails, due a CORS violation.

image

We'll need to either pass those requests through a data endpoint or disable `reuseFrame`.

packages/react-server-examples/bike-share/components/network-card.js
gigabo: Good to know!  Thanks. :)

packages/react-server-examples/bike-share/pages/index.js
gigabo: I don't _think_ those warnings should arise from the root level, since we don't actually let React render root containers or the outer array.  Would be nice to keep the example JSX clean of extraneous attributes.

doug-wade:
I did the quick 'n' dirty ™️ thing and pulled `reuseFrame` -- I'm not ready to recommend `lerna` (or some other js monorepo tool) as part of our examples.

doug-wade:
I added the `rawResponse` endpoint, migrated to react-server 0.3.1, and cleaned up the props.  I think this should be good to merge.

gigabo:
Awesome!  Looks great!

PR #224, opened 2016-05-26
Expose new convenience function: `navigateTo`

### LINKED ISSUE #114: Simpler API for client transitions
We've got to make this easier.  I want to be able to initiate a client transition with a single import.

Something as simple as:

```javascript
const {navigateTo} = require("react-server");

navigateTo(path, {options});
```

Currently looks like:

```javascript
const {
	ClientRequest
	History,
	getCurrentRequestContext,
} = require("react-server");

getCurrentRequestContext().navigate(
	new ClientRequest(this.props.path),
	History.events.PUSHSTATE
);
```

gigabo:
This is a much nicer interface for kicking off a client transition.

This closes #114.


doug-wade:
🔥 🔥 🔥 
👍

PR #226, opened 2016-05-27
Standardize routes file configuration (and rename `--routes` => `--routes-file`)

gigabo:
Previously the routes file took a special path through function calls that
prevented it from receiving values from the normal configuration sieve
(defaults => config file => cli).

When the (redundant) CLI default was removed in #206 this meant that a routes
file _must be specified on the command line_.

This patch standardizes handling of routes file configuration.  It also
changes the name of the `--routes` option to `--routes-file` to more
accurately reflect what it is (and to avoid a name conflict with an existing
`const` in `startServer.js`).  As such this is a breaking change.

### DIFF
--- a/packages/react-server-cli/src/cli.js
+++ b/packages/react-server-cli/src/cli.js
@@ -15,4 +15,4 @@ if (argv.httpsKey || argv.httpsCert || argv.httpsCa || argv.httpsPfx || argv.htt
 	}
 }
 
-start(argv.routes, argv);
+start(argv);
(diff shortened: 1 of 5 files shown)


gigabo:
@aickin, @withinboredom - I think you two are probably the most familiar with this code.  Have a look?

withinboredom:
lol ... woops :smiley: LGTM

doug-wade:
👍

aickin:
This is a good change! 👍

PR #230, opened 2016-05-30
Break up generated readme lines

doug-wade:
Based on [this comment](https://github.com/redfin/react-server/pull/222#discussion_r64776122) from @gigabo


gigabo:
Thanks @doug-wade. :)

Issue #203, opened 2016-05-23
Generate `.reactserverrc` with `yo react-server`


gigabo:
I think the generator predates the config file.  Seems like it should make one.

withinboredom:
What specifically should be generated in the default? Def want to have a default docker env setup ... Would make sense for a generic staging/prod, maybe a local env too?

gigabo:
Yeah, that sounds good.

Would it make sense just to populate with all the defaults so it's clear how to override?

PR #231, opened 2016-05-30
Add .reactserverrc to generator

doug-wade:
Fix [203](https://github.com/redfin/react-server/issues/203)


packages/generator-react-server/generators/app/templates/_reactserverrc
```diff
@@ -0,0 +1,25 @@
+{
```
gigabo: I guess the only down-side of having all of the config defaults listed here is we need to make sure it stays in sync with the actual defaults. 😛

gigabo:
👍

Issue #233, opened 2016-05-31
Handle web font files


gigabo:
Webpack chokes on them without a loader, and the asset servers don't know what to do with them.

Can they be handled in user space with smarter asset servers and [custom loader support](https://github.com/redfin/react-server/issues/199)?

Do we need to bake in handling for webfonts (and all future asset types)?

Need to do _something_.

Born from #227.

gigabo:
@withinboredom - You interested in looking at this?

aickin:
FYI, I'm working on the quick and dirty solution in a branch. Hopefully a PR within a few days, depending on other commitments.

gigabo:
Awesome.  Thanks @aickin!

aickin:
I believe this should be closed now that #251 is merged. 

Might be worth opening a bug for more generically applicable file handling, but the use case for this bug is fixed, I think.

gigabo:
Oh, yep!  Thanks @aickin.

PR #251, opened 2016-06-01
Handle image and font files linked from CSS

aickin:
This is the "quick and dirty" fix for #233. It hard codes a bunch of image and font file extensions and handles them with `file-loader`. It is not configurable.

`file-loader` copies the included files to the target directory and renames them as a file hash, so they can be cached forever. I think this should unblock @withinboredom.

Licensing notes for the assets:

- cat.png is CSS0-licensed, according to https://pixabay.com/en/animal-cat-leopard-snow-leopard-162020/
- the font files are SIL-licensed, according to https://github.com/theleagueof/prociono . I've included the license and copyright file in the directory, although I presume (but have not checked) that the font files also have the license in metadata.


packages/react-server-cli/bin/react-server-cli
```diff
@@ -1,8 +1,5 @@
 #!/usr/bin/env node
 
 require("babel-core/register");
-require.extensions['.css'] = function(module, filename) {
-  return module._compile("", filename);
-};
```
aickin: In order for require("foo.css") to work on server-side in tests, I had to move this to `index.js`.

packages/react-server-cli/src/compileClient.js
```diff
@@ -135,10 +135,15 @@ const packageCodeForBrowser = (entrypoints, outputDir, outputUrl, hot, minify, l
 					exclude: /node_modules/,
 				},
 				{
-					test: /.css$/,
+					test: /\.css$/,
```
aickin: This is an unrelated bug fix that I noticed.

packages/react-server-cli/src/index.js
```diff
@@ -1,3 +1,7 @@
-import startServer from "./startServer"
+require.extensions['.css'] = function(module, filename) {
+  return module._compile("", filename);
+};
```
aickin: This is still a bit of a hack, but probably the best thing if we don't want to do server-side webpack.

Note that we only include CSS here, which means that this PR will not support JS files that directly `require` a font or image file. This PR only supports fonts and images when they are included in CSS using the `url(foo.png)` syntax.

packages/react-server-integration-tests/gulpfile.babel.js
```diff
@@ -33,8 +34,12 @@ function getSpecGlob (prefix) {
...
 }
 gulp.task("compile", () => {
-	return gulp.src("src/**/*.js")
-		.pipe(babel())
+	const jsFilter = filter("**/*.js");
+
+	return gulp.src("src/**")
+		.pipe(jsFilter)
+			.pipe(babel())
+		.pipe(jsFilter.restore())
```
aickin: The tests were only copying over JS files in the build, but they need to copy over CSS/images/fonts too.

packages/react-server-cli/src/compileClient.js
```diff
@@ -135,10 +135,15 @@ const packageCodeForBrowser = (entrypoints, outputDir, outputUrl, hot, minify, l
...
 					loader: extractTextLoader,
 					exclude: /node_modules/,
 				},
+				{
+					test: /\.(eot|woff|woff2|ttf|ttc|png|svg|jpg|jpeg|gif|cgm|tiff|webp|bmp|ico)$/,
```
gigabo: Wow, that's a lot of file types!

Should this be `/.../i` to pick up upper-case extensions?

aickin: I did a bit of searching on the Internets for different font and image types. I started including audio file types and then decided that was stupid.

> Should this be /.../i to pick up upper-case extensions?

Good catch! I'll patch it up.

packages/react-server-integration-tests/src/__tests__/css/CssSpec.js
```diff
@@ -0,0 +1,35 @@
...
+		"./CssWithAssetsPage",
+	]);
+
+	helper.stopServerAfterAll();
+
+	describe("can apply a css class", () => {
+		helper.testWithDocument("/cssWithAssets", (document) => {
+      // this test is really just to make sure that react-server-cli doesn't explode
+      // when compiling CSS that includes images and fonts.
+			expect(document.querySelector("head link[rel=stylesheet]").href).toMatch(/route0\.css/);
```
gigabo: Any easy way to test that the secondary assets are served and referenced correctly?

gigabo:
This is a surprisingly little amount of code to make all of those assets work.  Magic.

Related: @aickin's back, baby! 🎉

packages/react-server-integration-tests/src/__tests__/css/CssSpec.js
aickin: Sigh. There's a way but not easy. Also, I'm lazy.

I first wrote this test to check the computed styles in the HTML to see if the CSS was applied, but it turns out that ZombieJS doesn't download CSS 😢. So, I think the good way to do this would be to download the CSS with node's `http` module and insert the text into a `<style>` node, then test the computed style. Then, for styles that are of the form `url(foo.bar)`, try to download the asset referenced.

However, the test as written was failing and showing the bug that @withinboredom found, and I was too lazy to do all of that in this PR. Can we leave that for down the road? 😬

gigabo: Yikes.  Yeah, sounds worth punting.  Maybe it'll be easier when tests are converted to selenium... ;)

aickin:
Fixed the case sensitivity issue in f15fc83. Thanks, @gigabo!

> This is a surprisingly little amount of code to make all of those assets work. Magic.

I think I spent about 90% of the time of this PR on getting the tests running and finding a font file with a permissive license. The code was like 10 minutes, most of which was looking up file extensions. 😄 

> Related: @aickin's back, baby! 🎉

Aw shucks. For the record, I'm probably not going to be as intensely involved as I was earlier in the year, but it sure feels great to be so welcomed. Thanks, @gigabo!

packages/react-server-cli/src/compileClient.js
```diff
@@ -135,10 +135,15 @@ const packageCodeForBrowser = (entrypoints, outputDir, outputUrl, hot, minify, l
 					exclude: /node_modules/,
 				},
 				{
-					test: /.css$/,
+					test: /\.css$/,
 					loader: extractTextLoader,
 					exclude: /node_modules/,
 				},
+				{
+					test: /\.(eot|woff|woff2|ttf|ttc|png|svg|jpg|jpeg|gif|cgm|tiff|webp|bmp|ico)$/i,
```
doug-wade: Should we use the file loader for other kinds of files, like csv and txt?

packages/react-server-cli/src/index.js
```diff
@@ -1,3 +1,7 @@
-import startServer from "./startServer"
+require.extensions['.css'] = function(module, filename) {
+	return module._compile("", filename);
+};
 
-export {startServer as start}
```
doug-wade: why the switch from es6 modules to commonjs?

aickin: I wanted to put the require extension before the call to `require` to be 100% sure that no CSS gets loaded. There's no way to do that with `import` 😢

doug-wade:
> For the record, I'm probably not going to be as intensely involved as I was earlier in the year

😿

packages/react-server-cli/src/compileClient.js
aickin: If folks want to, it's fine by me. Could also consider other media files (i.e. video, audio, and 3D/VR).

However, remember that with this PR that `require(foo.txt)` will still throw an error server-side, because the only `require.extension` we set is ".css", so adding txt or csv here will only help for txt and csv files _that are referenced by CSS_. That seems pretty unusual to me, and probably not worth it?

doug-wade: got it -- you're right, it doesn't make sense.

packages/react-server-cli/src/index.js
doug-wade: 🐑

doug-wade:
lgtm 👍

Issue #255, opened 2016-06-03
.reactserverrc is malformed json


DanielFGray:
It would seem `undefined` is not a valid symbol in json, and the debug log-level needed quoting

Seems strange that what one would assume to be generated JSON isn't actually valid
 
```diff
.reactserverrc
 {
-  "routes": "./routes.js"
+  "routes": "./routes.js",
   "host": "localhost",
   "port": 3000,
   "js-port": 3001,
   "hot": false,
   "minify": false,
   "long-term-caching": false,
   "compile-only": false,
-  "js-url": undefined,
   "https": false,
-  "https-key": undefined,
-  "https-cert": undefined,
-  "https-ca": undefined,
-  "https-pfx": undefined,
-  "https-passphrase": undefined,
-  "log-level": debug,
+  "log-level": "debug",
   "env": {
     "staging": {
       "port": "4000"
     },
     "production": {
        "port": "80"
     }
+  }
}
```

Changing this file and copying the `.gulpfile` from the repo as noted in #254 seems make the server work for me.

gigabo:
Ouch.  Thanks for reporting @DanielFGray.

Issue #254, opened 2016-06-02
Generated server failing with "No gulpfile found"


gigabo:
Looks like `gulp` was made part of `npm start` in #213.

@doug-wade - Ideas?  Looks like the gulpfile is _there_ in that PR...

![image](https://cloud.githubusercontent.com/assets/115908/15761971/13c95ad8-28d0-11e6-852b-6d1993a2f6cb.png)

PR #258, opened 2016-06-03
Add end-to-end test to generator-react-server

doug-wade:
Prevent bugs like #254 and #255 from sneaking in again


gigabo:
image

😬 

That's ~10 minutes added to our (already painfully slow) travis build.

These are high value tests, but... ouch.

doug-wade:
@gigabo I think we want to generate the tests regardless; I think the question here is whether we should run the generator's tests as part of the monorepo build.  Turns out we're not actually getting a lot of benefit from keeping the generator in the monorepo -- given that we don't keep the versions in sync for the [generated server](https://github.com/lerna/lerna/issues/156) any way, we may want to consider moving this to its own repo so it doesn't affect the build times for other React Server prs.

gigabo:
Yeah, that's worth considering.  I think it's valuable to have it in the main repo with everything else just for visibility, even if [lerna's not helping us](https://github.com/lerna/lerna/issues/156) keep versions in sync.

Our latest master build took [almost 10 minutes](https://travis-ci.org/redfin/react-server/builds/135095180) even though the build for the PR that merged to kick it off [only took 7](https://travis-ci.org/redfin/react-server/builds/135092445).  I've seen _complete_ tests hang for a while after finishing locally on my laptop.  I wonder if our test servers are somehow not exiting cleanly or if we could speed that up somehow...

I guess my point is I'd rather try to make the tests faster than to start spinning packages out of the repo to speed the build up.

aickin:
> I've seen complete tests hang for a while after finishing locally on my laptop.

This always happens for me, and I'm convinced (without any other data) that something in my test setup code isn't cleaning up servers correctly. I spent a while trying to find out what it could be and failed. 😢

gigabo:
Filed: #262.

doug-wade:
Looks like I'm blocked on #265

doug-wade:
Used `npm run nuke && npm run bootstrap` to get around #265, but now I seem to be stuck on #262.  I'm getting 

```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

from travis, which is strange, because local testing passes much faster than that

```
dougwade code/react-server ‹start-server-tests*› » time npm test
...
npm test  98.81s user 19.06s system 117% cpu 1:40.27 total
```

gigabo:
Ugh, that's frustrating.  Guess we'd better figure out what's going wrong there...

packages/generator-react-server/generators/app/templates/test.js
```diff
@@ -0,0 +1,54 @@
+import cp from 'child_process';
+import http from 'http';
+import test from 'ava';
+
+test.before('start the server', async () => {
+	let stdout = await exec('npm run compile');
+	console.log(stdout);
+	stdout = await exec('node node_modules/.bin/pm2 start npm -i 1 --name "hello-world" -- start');
```
gigabo: What is `pm2` giving us here?  Seems easier (and lower overhead) to just exec the server directly...

packages/generator-react-server/generators/app/templates/test.js
```diff
@@ -0,0 +1,54 @@
...
+	console.log(stdout);
+	await sleep(10000);
+});
+
+test('server is running', async t => {
+	t.is(200, await getResponseCode('/'));
```
gigabo: I like this async/await style!

packages/generator-react-server/generators/app/templates/test.js
doug-wade: I'm not sure how I would bring the server down when the tests are done, or how to make it possible to tail the logs in real time in a separate window.  We may very well be able to overcome both of those when `exec`ing the server directly; `pm2` is not a small dep, and I'd be happy to get rid of it.

gigabo: > bring the server down

[Kill](https://nodejs.org/api/child_process.html#child_process_child_kill_signal) it!

> tail the logs in real time in a separate window

Is that a requirement for automated tests?

doug-wade:
Final build time
screen shot 2016-06-10 at 11 36 32 am
seems about in line with [recent build times](http://scribu.github.io/travis-stats/#redfin/react-server)

packages/generator-react-server/generators/app/templates/README.md
```diff
@@ -23,6 +24,35 @@ after `--`. For example:
...
+npm i -g xo ava pm2 nsp
+```
+
+The first test is a linter, which checks for common bugs and code style; you can
+run it with `xo`.
+
+The second test is a security auditing test, which checks for known security
+issues with the installed dependencies.  You can run it with `nsp check`.
+
+The last test is an integration test, written with `ava` and using `pm2` to
```
gigabo: No longer using `pm2`, it looks like?  Docs out of date?

gigabo:
Nice!  This is looking slick.  Will be good to have real tests on the generated server.

I noticed one issue with the docs, but other than that this looks good to me! 👍

PR #260, opened 2016-06-03
Better getting-started section of react-server README

gigabo:
Taken from the top-level README, with absolute links pointing to npmjs.com
(and github for the docs) instead of relative links within the repo.

Also changed an instance of `react-server` to React Server.


doug-wade:
👍

PR #268, opened 2016-06-03
Include URL fragment in history navigation stack pathnames

### LINKED ISSUE #267: React server clobbers URL fragment on page load
The client controller does a `history.replaceState()` on page load to write request options into the history navigation stack.  It only uses [path and query](https://github.com/redfin/react-server/blob/master/packages/react-server/core/ClientController.js#L47-L49), though, and not fragment.

For example `/pages/foo#bar` gets rewritten to `pages/foo`.

gigabo:
This fixes #267.


doug-wade:
👍

Issue #262, opened 2016-06-03
Test servers hang during cleanup


gigabo:
Repro:

```bash
$ cd packages/react-server-integration-tests
$ npm test
```

After the tests complete the process doesn't exit immediately.  Maybe something in test server teardown?

![image](https://cloud.githubusercontent.com/assets/115908/15789116/792c360c-297f-11e6-9e74-f572059d07a0.png)

Seems to [cause travis build times to fluctuate](https://github.com/redfin/react-server/pull/258#issuecomment-223653737).

aickin:
To be clear for whoever works on this: the tests usually pause for a ton of time (minutes!) after the specs are all finished but before the gulp task exits, but to my knowledge they don't just hang _forever_.

gigabo:
Oops, the screen shot in the description left out some key info. 😆 

![image](https://cloud.githubusercontent.com/assets/115908/15789252/09d7e05c-2980-11e6-8812-93314385050d.png)

doug-wade:
Not sure if this is the right place to add this, but we can track our overall build time with [this here graph](http://scribu.github.io/travis-stats/#redfin/react-server)

PR #273, opened 2016-06-07
Faster integration test teardown

gigabo:
The integration test process was hanging around after completion, sometimes
for a long time.  Turns out zombie was holding a keep-alive connection open to
the static asset server.  This patch destroys the socket on teardown which
lets the server shut down.

This addresses #262.

### DIFF
--- a/packages/react-server-cli/src/startServer.js
+++ b/packages/react-server-cli/src/startServer.js
@@ -267,8 +267,22 @@ const handleCompilationErrors = (err, stats) => {
 // returns a method that can be used to stop the server. the returned method
 // returns a promise to indicate when the server is actually stopped.
 const serverToStopPromise = (server) => {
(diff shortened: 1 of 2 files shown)


gigabo:
Note that even _with_ this patch the integration tests hang around for ~20 seconds after completion.  Looked at `process._getActiveHandles()`, and there are a bunch of 20 second timers.  Some timeout on something?  Would be nice if those were unref'd.

roblg:
:tada:

re: 20s timers -- `setTimeout(navigateDfd.reject, FAILSAFE_ROUTER_TIMEOUT);` in renderMiddleware?

gigabo:
@roblg - Yes!  Nice find!

doug-wade:
lgtm 👍

PR #274, opened 2016-06-07
Clear failsafe timeouts on route/render success

### LINKED ISSUE #262: Test servers hang during cleanup

gigabo:
These were unnecessarily hanging around for 20s.  Nice catch @roblg!

This eliminates the 20s wait after integration test completion.

Fixes #262.

### DIFF
--- a/packages/react-server/core/renderMiddleware.js
+++ b/packages/react-server/core/renderMiddleware.js
@@ -138,7 +138,10 @@ module.exports = function(server, routes) {
 
 		var navigateDfd = Q.defer();
 
-		setTimeout(navigateDfd.reject, FAILSAFE_ROUTER_TIMEOUT);
+		const timeout = setTimeout(navigateDfd.reject, FAILSAFE_ROUTER_TIMEOUT);
+
+		// Don't leave dead timers hanging around.
+		navigateDfd.promise.then(() => clearTimeout(timeout));
 
 		// If we fail to navigate, we'll throw a 500 and move on.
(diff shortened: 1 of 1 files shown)


packages/react-server/core/renderMiddleware.js
```diff
@@ -138,7 +138,10 @@ module.exports = function(server, routes) {
 
 		var navigateDfd = Q.defer();
 
-		setTimeout(navigateDfd.reject, FAILSAFE_ROUTER_TIMEOUT);
+		const timeout = setTimeout(navigateDfd.reject, FAILSAFE_ROUTER_TIMEOUT);
```
roblg: This we should unref these `setTimeout`s too? Or does it even matter?

roblg:
LGTM. Curious about maybe adding `.unref()`, but changes look good.

packages/react-server/core/renderMiddleware.js
gigabo: Shouldn't matter.  These are intra-request timers, and we're going to wait for requests to wrap up on a graceful shutdown, anyway.

roblg: 👍

PR #275, opened 2016-06-07
Use ava tap output for generator tests

gigabo:
This is kind of hilarious, actually. The default ava reporter was filling up
the `lerna run` input buffer by erasing and re-drawing its spinner while it
waited for the test to complete.

This patch just switches to the tap reporter, which doesn't do that.


gigabo:
I had to [patch lerna](https://github.com/lerna/lerna/pull/181) to track this down.

doug-wade:
👍  cool stuff.  Thanks @gigabo

PR #278, opened 2016-06-08
Set the repository correctly in the generator's package.json

gigabo:
I think (guess) this is preventing stars from showing up on http://yeoman.io/generators/

image


doug-wade:
Oops
:+1:

PR #276, opened 2016-06-08
[DISCUSSION] Add babel plugin

doug-wade:
Hey gang, this one isn't quite ready, but I'm working on a fix for #247 .  Instead of running the gulp plugin, I'm running the transform in the babel webpack loader, and wanted to get some early feedback on the general approach.

Issue #247, opened 2016-05-31
react-server-cli doesn't hot reload or watch components that require react-server-gulp-module-tagger


doug-wade:
When working on #222, I had to take down the server and restart it every time I changed a component because I used the logger, and that requires tagging, which react-server-cli doesn't do.  We should either switch from using gulp to using a [babel plugin](https://babeljs.io/docs/plugins/) for our source transforms, or we could update react-server-cli to use react-server-gulp-module-tagger as part of its compile step.

gigabo:
> update react-server-cli to use react-server-gulp-module-tagger as part of its compile step

I totally thought it did that.

PR #280, opened 2016-06-09
Add babel plugin for module tagging

### LINKED ISSUE #12: Pass in the token to be replaced in the module tagger

doug-wade:
Add babel-plugin-react-server to the react-server-cli webpack loader and
babel-preset-react-server.  Factor out react-server-module-tagger from
react-server-gulp-module-tagger to share module tagging format between
gulp and babel, and add tests to ensure prevent regressions.  As discussed in #276; fixes #12 and #247


doug-wade:
This has gotten unwieldy, so I'm breaking it into multiple prs (see #300, #294, and #297, with more to follow)

PR #282, opened 2016-06-10
Run generator end-to-end test on unique port

doug-wade:
Fix broken master build


gigabo:
👍 

Maybe we should keep track of the ports we're using for tests somewhere?

PR #283, opened 2016-06-10
Add test port manifest

doug-wade:
As suggested in https://github.com/redfin/react-server/pull/282#issuecomment-225273000
Note: the manifest may be incomplete or out-and-out wrong, as my test target is currently throwing


doug-wade:
@gigabo this was your suggestion; do you have any thoughts?

gigabo:
Yeah, great. 👍 

I thought we had one more, but I don't remember where it is... so, :shipit:

Issue #250, opened 2016-06-01
Bootstrap and wake nodes in a single script tag


gigabo:
Even better would be a single LAB script.

This is ugly, and inefficient:

![image](https://cloud.githubusercontent.com/assets/115908/15718387/ed7f0422-27df-11e6-9f4c-94f149f84848.png)

gigabo:
This was closed by #288.

PR #288, opened 2016-06-15
Reduce nodeArrival calls by taking in range of node indices

SteveVitali:
Previously ClientController's nodeArrival function just took in the node index,
but this made it such that each arriving node required its own script tag and
LABjs callback; now nodeArrival can take in a range of indices to minimize
overhead.

### DIFF
--- a/packages/react-server/core/renderMiddleware.js
+++ b/packages/react-server/core/renderMiddleware.js
@@ -838,7 +838,7 @@ function writeElements(res, elements) {
 
 			// We've already bootstrapped, so we can immediately tell the
 			// client controller to wake the new element we just sent.
-			wakeElement(res, i);
+			wakeElementRange(res, i, i);
 		} else if (i === elements.length - 1) {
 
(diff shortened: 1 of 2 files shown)


packages/react-server/core/renderMiddleware.js
```diff
@@ -915,15 +915,16 @@ function bootstrapClient(res, lastElementSent) {
 	// function to avoid letting responses slip in between.
 	setupLateArrivals(res);
 
-	for (var i = 0; i <= lastElementSent; i++) {
-		wakeElement(res, i);
-	}
+	wakeElement(res, 0, lastElementSent);
```
gigabo: `wakeElement` seems a little mis-named now that it may wake _multiple_ elements. 😕

gigabo:
This is awesome!  This could provide a big improvement on pages with lots of root elements! 🎉

gigabo:
This addresses #250.

doug-wade:
Amazing!  Thanks so much!

lgtm 👍 🚀

PR #289, opened 2016-06-15
Added info on how to start a yeoman-generated project

aickin:
Really simple edit to the two READMEs adding info on how to run a yeoman-generated project. I followed the instructions and got confused why it wouldn't run; I figured others probably would, too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,13 @@ npm install -g yo
 # install the react-server generator
 npm install -g generator-react-server
 
-# make a new react-server project
+# make a new react-server project in the CURRENT directory
 yo react-server
+
+# compile and run the new app
+npm run compile
+npm run start
+# go to http://localhost:3010
 ```
 
 That hooks you up with [`react-server-cli`](packages/react-server-cli), which
--- a/packages/react-server/README.md
+++ b/packages/react-server/README.md
@@ -30,8 +30,13 @@ npm install -g yo
 # install the react-server generator
 npm install -g generator-react-server
 
-# make a new react-server project
+# make a new react-server project in the CURRENT directory
 yo react-server
+
+# compile and run the new app
+npm run compile
+npm run start
+# go to http://localhost:3010
 ```
 
 That hooks you up with [`react-server-cli`](packages/react-server-cli), which


gigabo:
Whoa.  @doug-wade - Is that new?  Didn't it used to just start with `npm start`?

README.md
```diff
@@ -23,8 +23,13 @@ npm install -g yo
 # install the react-server generator
 npm install -g generator-react-server
 
-# make a new react-server project
+# make a new react-server project in the CURRENT directory
```
gigabo: Yeah, that's kinda weird.  I guess it's just the convention with `yo`? 😕

aickin:
Actually, wait. This may be a mistake on my part. Gimme a sec to see if I screwed up.

aickin:
> Actually, wait. This may be a mistake on my part. Gimme a sec to see if I screwed up.

No, I was right the first time; got confused looking at the wrong package.json. 

Looks like what the yeoman project does is: `npm run compile` calls `gulp`, which runs the JS files through `react-server-gulp-module-tagger` and babel with react-server preset, dumping the output in the `build` directory. Then `npm run start` runs `react-server-cli` with `build/routes.js`.

I think that after #280 is merged, the compile step can be skipped altogether. This will enable hot reloading (which currently doesn't work because of the build step) and CSS and other assets (which currently don't work because the gulp script doesn't copy them to `build`.

README.md
aickin: I guess so, but I don't really know.

I also ran into the problem that it defaulted to calling the project by my current directory name, replacing `-` with a space, but when it tried to npm install, that failed. I should file that.

gigabo: image

gigabo: Maybe we should change the instructions to include:

```bash
$ mkdir myServer
$ cd myServer
```

But... ugh.

aickin: Filed as https://github.com/redfin/react-server/issues/290

aickin: FWIW, I just tried `yo generator-webapp`, which is the example on Yeoman's README, and it drops all its files in the current directory, so I guess that's the standard. Interestingly, they don't tell you to make a sub-directory in their example.

doug-wade: yup, that's just how yeoman works.  I guess I didn't think to add it to the README because its also what I expect; change lgtm

doug-wade:
@gigabo no, you have to run compile then start because module tagging.  @aickin's summary is 100% spot-on

gigabo:
I see.  In that case let's get #280 out (and revert that portion of this PR when we do)!

Thanks for clearing this up @aickin.

PR #294, opened 2016-06-16
Add gulp module tagger tests

doug-wade:
#280 is getting too big for me to handle, so I'm breaking it up into smaller prs.  Add regression tests to prevent errors while refactoring gulp-module-tagger.  Includes #297

PR #297, opened 2016-06-20
Cat npm-debug.logs on failure

doug-wade:
#280 and #294 are plagued by failures that I can't repro locally.  Dump every `npm-debug.log` in the monorepo to stdout so they're available in the travis logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,6 @@ addons:
 
 before_script:
   - npm run bootstrap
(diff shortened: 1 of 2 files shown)


doug-wade:
Here's a build that includes the result https://travis-ci.org/redfin/react-server/builds/139049381

gigabo:
Awesome.

PR #298, opened 2016-06-21
Validate project name in generator to ensure it's a valid npm package name

### LINKED ISSUE #290: Yeoman generator fails with defaults if directory has a dash
To repro, assuming you have `yo` and `generator-react-server` globally installed:

1. `mkdir my-dir`
1. `cd my-dir`
1. `yo react-server`
1. Accept all defaults

The name that it suggests for the project as a default will be "my dir". But when yo attempts to call `npm install`, that will fail, as you can't have spaces in a project name.

Expected: it uses the `my-dir` as the default project name.

latentflip:
Fixes #290 

Ran into this myself so figured I'd fix it :)

### DIFF
--- a/packages/generator-react-server/generators/app/index.js
+++ b/packages/generator-react-server/generators/app/index.js
@@ -2,6 +2,7 @@
 var yeoman = require('yeoman-generator');
 var chalk = require('chalk');
 var yosay = require('yosay');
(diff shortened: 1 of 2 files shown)


doug-wade:
Huge thanks!  lgtm 👍

PR #300, opened 2016-06-21
Add react server module tagger

doug-wade:
Continue breaking up #280 (see also #294 and #297)

### DIFF
--- a/packages/react-server-module-tagger/README.md
+++ b/packages/react-server-module-tagger/README.md
@@ -1,0 +1,25 @@
+# react-server-module-tagger
+
+A function for tagging [react-server](https://www.npmjs.com/package/react-server)
+logger instances with information about the module they're being used in.
+
+To transpile your source for use with
+[React Server](https://www.npmjs.com/package/react-server), install gulp and the plugin
+
+```shell
+npm i -D gulp react-server-module-tagger
+```
+
+Then require and call the function.  The tagger expects to have config and file
+data on its prototype, so use `.bind`.
+
+```javascript
+const tagger = require('react-server-module-tagger');
+const filepath = 'path/to/my/output.js';
+const optString = '({label:"foo"})'
+const moduleTag = tagger.bind({ file: { path: filepath }, config: { trim: 'path/to'} })(filepath, optString));
+```
+
+returns a logger instance that will have consistent coloring on the server and
+the client, and that has a human-friendly, readable name that easily maps to
+the file tree (in this example `components.my-feature.foo`).
--- a/packages/react-server-module-tagger/index.js
+++ b/packages/react-server-module-tagger/index.js
@@ -1,0 +1,129 @@
+var isWindows = ('win32' === process.platform);
+
+/**
+ * A util function for tagging modules.  Expects to find data on its prototype
+ * as follows
+ *     {
+ *         file: {
+ *             path: the path to the file to tag
+ *         }
+ *         config: {
+ *             trim: the string to trim off the front of the logger name
+ *         }
+ *     }
+ * @example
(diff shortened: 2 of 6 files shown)


packages/react-server-integration-tests/package.json
```diff
@@ -13,6 +13,7 @@
     "react-dom": "~0.14.2",
     "react-server": "../react-server",
     "react-server-cli": "../react-server-cli",
+    "react-server-module-tagger": "../react-server-module-tagger",
```
doug-wade: I make this transitive dependency a direct dependency so that `react-server-module-tagger`'s `prepublish` script is called; otherwise, we get module not found errors from `react-server-gulp-module-tagger` and `babel-plugin-react-server`

packages/react-server-module-tagger/index.js
```diff
@@ -0,0 +1,129 @@
+var isWindows = ('win32' === process.platform);
+
+/**
+ * A util function for tagging modules.  Expects to find data on its prototype
```
gigabo: It expects to find data in its context, right?  Doesn't need to be on the prototype.

packages/react-server-module-tagger/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+To transpile your source for use with
+[React Server](https://www.npmjs.com/package/react-server), install gulp and the plugin
+
+```shell
+npm i -D gulp react-server-module-tagger
+```
+
+Then require and call the function.  The tagger expects to have config and file
+data on its prototype, so use `.bind`.
```
gigabo: ... in its context...

packages/react-server-module-tagger/index.js
```diff
@@ -0,0 +1,129 @@
...
+		// The slash replacement here is so we don't choke on example
+		// loggers in comments.
+		opts = new Function("return "+optString.replace(/^\/\//mg,''))(); // eslint-disable-line no-new-func
+	}
+
+	opts.name  = getName  (fn, opts, trim, basePath);
+	opts.color = getColor (opts);
+
+	return JSON.stringify(opts);
+}
```
gigabo: Really, a nicer interface would be to accept the context object as an argument and return `opts` as an object.  The binding and serialization seems more relevant to the calling environment in the gulp module tagger.

packages/react-server-module-tagger/index.js
```diff
@@ -0,0 +1,129 @@
...
+ * @example
+ *     const file = '/path/to/my/file.js';
+ *     loggerSpec.bind({file, { trim: 'path.to.' }})(file, '({label: "foo"})')
+ *     // returns '{\"label\":\"foo\",\"name\":\"my.file\",\"color\":{\"server\":87,\"client\":\"rgb(42,212,212)\"}}'
+ * @param  {String} fullMatch May also be provided as this.file.path, the path to the file
+ * @param  {String} optString The label to add to the module tag, in the form '({label:"$label"})'
+ * @return {String}           A json object containing a module identifier
+ */
+module.exports = function(fullMatch, optString){
+	var fn   = this.file && this.file.path ? this.file.path : fullMatch
```
gigabo: `file` isn't used for anything other than the `path` in here, right?  Might be nicer to just take the path rather than the file object, which is particular to gulp.

roblg:
Looks reasonable to me, though I didn't go line by line through the stuff that was moved from `react-server-gulp-module-tagger` to `react-server-module-tagger`. 👍 to more documentation there.

gigabo:
Yeah, LGTM, too.  👍 

I've got some interface shuffling itches, but nothing prohibitive.

PR #302, opened 2016-06-22
Use babel preset in cli

doug-wade:
Transpile the cli (but not its clients) with babel-preset-react-server

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,9 +10,11 @@
   },
   "devDependencies": {
     "babel-plugin-react-require": "^2.1.0",
+    "babel-plugin-react-server": "^0.3.3",
     "babel-plugin-transform-runtime": "~6.5.0",
     "babel-preset-es2015": "~6.5.0",
     "babel-preset-react": "^6.5.0",
+    "babel-preset-react-server": "^0.3.3",
     "babel-preset-stage-1": "^6.5.0",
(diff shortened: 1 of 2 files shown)


doug-wade:
This is blocking on 0.4.0 to publish

gigabo:
@doug-wade - Blocking on 0.4.0 of what?

doug-wade:
@gigabo React Server!

gigabo:
It requires breaking changes?

doug-wade:
0.3.3, 0.4.0, or 1.0.0, whichever comes first

package.json
```diff
@@ -38,7 +38,7 @@
     "jasmine": "2.3.1",
     "jsdom": "5.6.1",
     "jsx-loader": "0.12.2",
-    "lerna": "2.0.0-beta.23",
+    "lerna": "git://github.com/lerna/lerna.git#master",
```
doug-wade: build is passing locally with my lerna built off master (which runs `prepublish` as part of bootstrap), but not on ci so I tried pulling from master.  However, lerna's master branch isn't a functioning node module, so this isn't working either.  Next, I'll try a `lerna run prepublish` on ci, but if that doesn't work, we may just need to wait for lerna-2.0.0-beta.24

PR #306, opened 2016-06-24
Remove unused fn variable

doug-wade:
Resolve https://github.com/redfin/react-server/pull/294#discussion-diff-68335654


gigabo:
👍

PR #309, opened 2016-06-24
Dependency refresh

doug-wade:
update all the deps


doug-wade:
oh wait snap updated too many deps

doug-wade:
I apparently haven't learned my lesson; Ima do this per-dependency

Issue #314, opened 2016-06-29
Improve react-server-module-tagger interface


doug-wade:
In #301 and #302, I factored out a common module from react-server-gulp-module tagger that has some pretty rough remnants of its past as a gulp module.  We should clean up the interface to be more appropriate to the new api.

doug-wade:
This was resolved in https://github.com/redfin/react-server/commit/98a95771f8dcccb644987cdce01316e7b47a34d8

Issue #277, opened 2016-06-08
Eliminate dependencies


gigabo:
`npm run bootstrap` takes _forever_.  Do we really need the kitchen sink?

Eliminate a dependency, and be celebrated for all time! 🎉

gigabo:
Some candidates:

`xo`:

```
 22M    node_modules/eslint
 82M    node_modules/xo
```

`ava`:

```
256M    node_modules/ava
820K    node_modules/jasmine
```

These are modules that have equivalents _already in our repo_.  Standardizing on one can reduce our dependencies without sacrificing functionality.  Just dropping `xo` and `ava` would eliminate _300MB!_

doug-wade:
Bonus points if they are [slow dependencies](https://www.npmjs.com/package/slow-deps)

Saifu0:
@gigabo I can take this, is this still an issue?

dhuang612:
Hi, 
is this available to work on?

mbhole1857:
Hi there,
Is the issue is still open??

PR #320, opened 2016-06-30
Lint react-server-module-tagger

doug-wade:
Get ready for #314 by adding linting, cleaning up linting errors, and adding a watch mode.  There is nothing of substance in this pr

### DIFF
--- a/packages/react-server-module-tagger/index.js
+++ b/packages/react-server-module-tagger/index.js
@@ -1,4 +1,91 @@
-var isWindows = ('win32' === process.platform);
+var isWindows = (process.platform === 'win32');
+
+/**
+ * Gets the name of a logger from its filepath.
+ * @example
+ *     getName(
+ *         'my-component',
+ *         { label: 'sub' },
+ *         'my-project.src',
+ *         'my-project/src/components/my-component.js'
+ *     ) // returns 'components.my-component'
+ * @param  {String}   fn       filename
+ * @param  {Object}   opts     { label: 'Optional logger label' }
+ * @param  {String}   trim     The leading portion of the name to remove.
+ * @param  {String}   basePath The path to the file
+ * @return {String}            The logger name, e.g. my-project.components.my-component
+ */
+var getName = function (fn, opts, trim, basePath) {
+	var slashPattern = isWindows ? /\\/g : /\//g;
+	var name = fn.substring(basePath.length + trim.length, fn.length)
+		.replace(/\.jsx?$/, '')
+		.replace(slashPattern, '.');
+	if (opts.label) {
+		name += '.' + opts.label;
+	}
+	return name;
+};
+
+/**
+ * Gets isomorphic color objects from filenames.
+ * @param  {String} filename
+ * @return {Object} An isomorphic color object in the form {'color':{'server':147,'client':'rgb(127,127,212)'}}
+ */
+var getColor = (function () {
+	// ANSI escape sequence on the server.
+	// CSS rgb(...) color in the browser.
+	var makeColor = function (r, g, b) {
+		return {
+			server: 16 + (r * 36) + (g * 6) + b,
+			client: 'rgb(' + [
+				(r * 42.5) | 0,
+				(g * 42.5) | 0,
+				(b * 42.5) | 0
+			].join(',') + ')'
+		};
+	};
+
+	// This produces a list of 24 colors that are distant enough from each
+	// other to be visually distinct.  It's also, conveniently, the same
+	// palette client side and server side.
+	var colors = [];
+	for (var r = 1; r < 6; r += 2) {
+		for (var g = 1; g < 6; g += 2) {
+			for (var b = 1; b < 6; b += 2) {
+				if (r !== g || g !== b) { // No gray.
+					colors.push(makeColor(r, g, b));
+				}
+			}
+		}
+	}
+
+	// Just want a fairly well distributed deterministic mapping.
+	//
+	// Adapted from:
+	// http://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript-jquery
+	var hash = function (str) {
+		var hash = 0;
+		var chr;
+		if (str.length === 0) {
+			return hash;
+		}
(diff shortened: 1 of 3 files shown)


packages/react-server-module-tagger/package.json
```diff
@@ -4,8 +4,10 @@
   "description": "calculates module level config",
   "main": "index.js",
   "scripts": {
-    "test": "ava test.js",
-    "clean": "del index.js npm-debug.log*"
+    "lint": "xo",
+    "test": "ava test.js && xo",
+    "clean": "del index.js npm-debug.log*",
+    "watch": "chokidar index.js -c 'npm test'"
```
doug-wade: "But Doug", you might say, "surely setting up a watch mode can't be this simple!", and yet it is.

packages/react-server-module-tagger/index.js
```diff
@@ -49,81 +8,122 @@ module.exports = function(fullMatch, optString){
...
 
 /**
  * Gets isomorphic color objects from filenames.
  * @param  {String} filename
- * @return {Object} An isomorphic color object in the form {"color":{"server":147,"client":"rgb(127,127,212)"}}
+ * @return {Object} An isomorphic color object in the form {'color':{'server':147,'client':'rgb(127,127,212)'}}
  */
-var getColor = (function(){
-
+var getColor = (function () {
```
doug-wade: So, given [var hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting), this isn't actually necessary.  The linting rule would allow this to be at the bottom of the file were it a function declaration, but the simple transformation breaks the closure we use and fails the tests, so I opted for the simplest fix (manually hoisting it), and I'll make this and getName function declarations when I fix #314

packages/react-server-module-tagger/index.js
```diff
@@ -49,81 +8,122 @@ module.exports = function(fullMatch, optString){
...
 		return {
-			server: 16 + r*36 + g*6 + b,
-			client: "rgb("+[
-				(r*42.5)|0,
-				(g*42.5)|0,
-				(b*42.5)|0,
-			].join(',')+")",
-		}
-	}
+			server: 16 + r * 36 + g * 6 + b,
```
doug-wade: Maybe I should add parens because expecting devs to know about order of operations is too much?

packages/react-server-module-tagger/index.js
```diff
@@ -49,81 +8,122 @@ module.exports = function(fullMatch, optString){
...
-		for (i = 0, len = str.length; i < len; i++) {
-			chr   = str.charCodeAt(i);
-			hash  = ((hash << 5) - hash) + chr;
+	var hash = function (str) {
+		var hash = 0;
+		var chr;
+		if (str.length === 0) {
+			return hash;
+		}
+		var len = str.length;
```
doug-wade: this technically move `len` out of the scope in which it is used, but :yolo:

packages/react-server-module-tagger/test.js
```diff
@@ -2,31 +2,31 @@ import test from 'ava';
 import loggerSpec from '.';
 
 test('creates a module tag', t => {
-  const expected = '{\"name\":\"foo.bar\",\"color\":{\"server\":73,\"client\":\"rgb(42,127,127)\"}}';
```
doug-wade: Turns out just copy-pasting from the node terminal is _not_ the best way to get regression test values 😆

packages/react-server-module-tagger/index.js
gigabo: Arithmetic order of operations is not language-specific.  Seems reasonable to expect devs to be familiar.

packages/react-server-module-tagger/index.js
```diff
@@ -49,81 +8,122 @@ module.exports = function(fullMatch, optString){
...
-			server: 16 + r*36 + g*6 + b,
-			client: "rgb("+[
-				(r*42.5)|0,
-				(g*42.5)|0,
-				(b*42.5)|0,
-			].join(',')+")",
-		}
-	}
+			server: 16 + r * 36 + g * 6 + b,
+			client: 'rgb(' + [
```
gigabo: What rule is requiring single-quotes?  Kinda bugs me that this is the [_opposite_ of lerna](https://github.com/lerna/lerna/blob/master/.eslintrc#L5-L9).

packages/react-server-module-tagger/index.js
```diff
@@ -49,81 +8,122 @@ module.exports = function(fullMatch, optString){
...
-		if (str.length === 0) return hash;
-		for (i = 0, len = str.length; i < len; i++) {
-			chr   = str.charCodeAt(i);
-			hash  = ((hash << 5) - hash) + chr;
+	var hash = function (str) {
+		var hash = 0;
+		var chr;
+		if (str.length === 0) {
+			return hash;
+		}
```
gigabo: What rule prohibits single-line if statements?  They're useful for this sort of thing.

doug-wade: I don't actually know -- `xo` comes with a reasonable set of defaults and I just took them as-is and called it a day.

gigabo: Wait... why is this using `xo` instead of `eslint`?

doug-wade: Its what we use for [the generator](https://github.com/redfin/react-server/blob/master/packages/generator-react-server/package.json#L31) and [the bike share example](https://github.com/redfin/react-server/blob/master/packages/react-server-examples/bike-share/package.json#L12) so I figured it wouldn't be an issue.  Mostly I like it because its really easy to set up.

gigabo: We already have a linter config for the project.

It is confusing to have different parts of the repo require conflicting conventions.

Also:

```sh
$ npm install xo
$ npm install eslint
$ du -sh node_modules/*
 22M    node_modules/eslint
 82M    node_modules/xo
```

Seems like the perfect candidate for #277.

PR #330, opened 2016-07-01
Wait for CSS to load on client transition

gigabo:
Don't start writing content into the DOM until external CSS has loaded.


roblg:
Previously, I think we'd get a flash of unstyled (or broken-styled) content as stuff was being added into the dom and stylesheets were being swapped. Does this fix that?

gigabo:
> Does this fix that?

Yep.

roblg:
👍

PR #348, opened 2016-07-14
Add json loader to webpack

doug-wade:
See our gitter chat starting at https://gitter.im/redfin/react-server?at=5786ecb59f79ee4f2b9e1666 for context

### DIFF
--- a/packages/react-server-cli/package.json
+++ b/packages/react-server-cli/package.json
@@ -23,6 +23,7 @@
     "express": "~4.14.0",
     "extract-text-webpack-plugin": "~0.8.2",
     "file-loader": "~0.8.5",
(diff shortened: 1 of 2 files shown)


gigabo:
Nice. 👍

gigabo:
Yeah, it would be nice to have this pluggable somehow.  JSON is a pretty good stopgap, though. ;)

doug-wade:
> it may be worth noting in docs somewhere that loading files that we don't have a loader coded in for will cause failures/bugs

I find this to be a very compelling suggestion, and I'd kinda like to get it into this pr, but I'm not really sure where in the docs it belongs.  react-server-cli's README, maybe?

PR #350, opened 2016-07-15
Fix the build

doug-wade:
The number of prs that are piling up on our broken ci build is out of control.  Let's get this under control.


packages/react-server/core/logging/stats.js
```diff
@@ -73,7 +73,7 @@ var wrapLogger = function(getLoggerForConfig, opts){
 
 			// The `stop` method logs the total elapsed time since
 			// timer creation.
-			stop: () => (mainLogger.time(token, ct = new Date - t0, opts), ct),
+			stop: () => { mainLogger.time(token, ct = new Date - t0, opts); return ct; },
```
gigabo: ?

doug-wade:
```
/Users/perry.shuman/Code/react-server/packages/react-server/core/logging/stats.js
76:10 error Arrow function should not return assignment no-return-assign
✖ 1 problem (1 error, 0 warnings)
[13:22:52] 'eslint' errored after 2.11 s
[13:22:52] ESLintError in plugin 'gulp-eslint'
Message:
Failed with 1 error
```

gigabo:
Weird. 👍

Issue #317, opened 2016-06-29
Add engines to generator-react-server


doug-wade:
Check out the conversation [in gitter](https://gitter.im/redfin/react-server?at=5770334e8c9263ba30f8470f), but in general, we wanna warn when a user is using the wrong node version

Noirbot:
What would be the correct version though? `>4.2.1`?

doug-wade:
@Noirbot seems like a reasonable place to start.  If you were really feeling ambitious, you could test it using something like [nave](https://github.com/isaacs/nave) or [n](https://github.com/tj/n)

PR #354, opened 2016-07-18
Issue #317: Add node version to Package.

Noirbot:
While this won't actually enforce that people use a specific Node version, this will at least serve as a reference for users to know what versions of node this was intended to run with.

Original issue: #317.


doug-wade:
Thank you @Noirbot.  lgtm!

PR #356, opened 2016-07-19
Making react-server-cli a relative dependency

davidalber:
Issue reproduction:

1. In repo root `npm run bootstrap`.
1. `cd packages/react-server`
1. `npm start`

This results in the following:
```
$ npm start

> example@0.0.1 start /Users/david.alber/dev/react-server/packages/react-server-website
> react-server-cli --port 3010 --js-port 3011

sh: react-server-cli: command not found
.
.
.
```

This is fixed by using a relative dependency in react-server-website.

### DIFF
--- a/packages/react-server-website/package.json
+++ b/packages/react-server-website/package.json
@@ -15,7 +15,7 @@
     "react": "~0.14.2",
     "react-dom": "~0.14.2",
     "react-server": "^0.3.4",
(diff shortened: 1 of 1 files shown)


gigabo:
I _think_ we need to do this with `react-server`, too...?

davidalber:
@gigabo: Not to make the react-server-website app launch. I'm not sure if something else is broken that needs this.

PR #363, opened 2016-07-19
Added common page components and added content to the homepage.

decoy31:


gigabo:
Hi @decoy31, please rebase this again to pick up #365, which just landed.

PR #367, opened 2016-07-20
Switch gears on Markdown/syntax highlighting

egid:
I've ripped prism and react-markdown back out in favor of highlight.js and remarkable. By combining the two we have a component that handles markdown and syntax highlighting in one shot.


gigabo:
Hi @egid, please rebase this to pick up #365, which just landed.

PR #373, opened 2016-07-20
Ensure that if `common` exists in the manifest, we include it.

### LINKED ISSUE #155: css can't be shared between sites
When you use the same css file between two sites, it breaks the css altogether.  I pulled together a [quick repro](https://github.com/doug-wade/react-server-repro); [this pr](https://github.com/doug-wade/react-server-repro/pull/1) breaks the styles for both / and /other

egid:
This fixes #155.

I've also removed a comment in the `[routeName]` section that no longer provides context.


gigabo:
Nice!  Thanks @egid! :metal:

PR #382, opened 2016-07-21
Setting up containerization and composition

davidalber:
- Creating Dockerfiles for the Node service and NGINX.
- Set up service composition.
- Static assets are shared from the Node container to the NGINX container as a volume.
- Added build targets for the container.


packages/react-server-website/Dockerfile.nginx
```diff
@@ -0,0 +1,13 @@
+FROM ubuntu
```
davidalber: I will pin the version on the base image.

PR #404, opened 2016-07-21
Incrementing Generator default versions.

Noirbot:
The Generator is defaulting people to a fairly old version of `React-Server` and the CLI, so I bumped those to the most recent one.

Is there a more effective way to do this? Where it will auto-update?


doug-wade:
This is a duplicate of #361

PR #444, opened 2016-07-25
Hackathon website

gigabo:
Nice work everyone!  I think what we have here is safe to merge into master.  We can continue to improve with PRs pointed there.

The team:
- @briannathanhartwell
- @davidalber 
- @decoy31 
- @doug-wade 
- @egid 
- @gigabo


doug-wade:
👍

egid:
👍

PR #445, opened 2016-07-25
Trying out travis caching

gigabo:
https://docs.travis-ci.com/user/caching/

Going to stage this from the redfin org instead of my user account, since only
normal pushes update the cache (not PR builds).


gigabo:
This nearly cuts build time in half (~8 1/2 min => ~4 1/2 min).

image

gigabo:
With one final tweak (avoid unnecessary cache store):

image

_More_ than twice as fast. 🚀

PR #456, opened 2016-07-27
Stab in the dark at dealing with spurious integration test failures

gigabo:
🔪


gigabo:
Fiiinnaalllyyyyyyy!!!!! Figured it out.

Turns out jasmine's default timeout applies to asynchronous `beforeAll` functions, too.  Just need to pass an override to increase it for server start up. :feelsgood:

PR #457, opened 2016-07-28
Improvements to provisioning and deployment and fixing Slackin credentials

davidalber:
- Fixing Slackin credentials (hopefully), and storing new credentials encrypted in the repository.
- Improvements to provisioning and deployment:
  - Adding EC2 dynamic inventory script and configuration.
  - Adding Ansible deployment configuration to allow redeploying into existing instances.
  - Updating provisioning configuration to cycle instances by bringing up new instances then removing existing instances.

_My branch is currently deployed to the production site, so we can test the Slack feature._

### DIFF
--- a/packages/react-server-website/provisioning/decrypt_credentials.py
+++ b/packages/react-server-website/provisioning/decrypt_credentials.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+
+import argparse
+from base64 import b64decode
+import boto3
+import json
+
+"""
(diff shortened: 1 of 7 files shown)


davidalber:
We are going to want a different email for Slackin. Before setting a name in Slack on the account, this is what the email from Slack looks like after submitting the form:
image

After setting a name, the email was still present.

gigabo:
> sinkhole@gmail.com

😬

davidalber:
Better!
image

PR #460, opened 2016-07-28
Just don't mess with `resolve.root` at all

gigabo:
Don't seem to need anything but loaders from cli's node_modules?


aickin:
Between this, #458, and #459, um, thanks for picking up after me! 😟😟😟

gigabo:
Together we take on webpack and we shall prevail! 🎏

gigabo:
(Please let me know if I'm doing anything dumb here... 😁 )

PR #463, opened 2016-07-29
Get react-server-cli link on the homepage working

gigabo:
Link was broken.

Also, can't `reuseDom` between homepage and docs section.


gigabo:
Also pushed a text change to avoid this weird wrap:

image

Now looks like:

image

PR #465, opened 2016-07-29
Website deployment improvements

davidalber:
- Modifying deployment commands.
- Deploying Slack token.
- Updating to insert instances into new load balancer.

### DIFF
--- a/packages/react-server-website/provisioning/deploy.yml
+++ b/packages/react-server-website/provisioning/deploy.yml
@@ -2,6 +2,9 @@
   hosts: tag_Name_react_server_box
   user: ubuntu
   tasks:
+    - name: Install Slack API token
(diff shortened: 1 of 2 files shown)


packages/react-server-website/provisioning/deploy.yml
```diff
@@ -2,16 +2,19 @@
   hosts: tag_Name_react_server_box
   user: ubuntu
   tasks:
+    - name: Install Slack API token
+      script: decrypt_credentials.py -k slack.api.token > slack-api-token
```
gigabo: Is `decrypt_credentials.py` something we can check in?

davidalber: It is checked in: https://github.com/redfin/react-server/blob/master/packages/react-server-website/provisioning/decrypt_credentials.py.

PR #471, opened 2016-07-29
Bundle caching for the website

gigabo:
Each page ensures sections (docs, source) are pre-loaded.

Each section ensures all pages in section are pre-loaded.

:rocket:

### DIFF
--- a/packages/react-server-website/pages/source-api.js
+++ b/packages/react-server-website/pages/source-api.js
@@ -8,7 +8,7 @@ export default class SourceApi {
 	}
 
 	getContentType() {
-		return 'text/html';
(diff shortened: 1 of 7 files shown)


packages/react-server-website/pages/source-api.js
```diff
@@ -8,7 +8,7 @@ export default class SourceApi {
 	}
 
 	getContentType() {
-		return 'text/html';
+		return 'application/json';
```
gigabo: `ReactServerAgent` actually won't transfer `text/html` responses.

Requests to this endpoint weren't getting bundled.

PR #474, opened 2016-07-29
Using NGINX proxy cache for documentation site

davidalber:
- Using NGINX caching on responses from the documentation site.
- Changing `default` to `default_server`, which is just an update to the current correct directive name.


gigabo:
Yeah! 🚀

PR #476, opened 2016-07-29
Redirecting HTTP to HTTPS

davidalber:
- Modifying the NGINX configuration to redirect requests arriving on port 80 to HTTPS. Note the load balancer has been configured to send HTTPS requests to port 443 in the services using the HTTP.
- Opening port 443 in the NGINX service section of the Docker Compose file.


gigabo:
Awesome.  Thanks @davidalber!

PR #481, opened 2016-08-01
Add debug params for log level adjustment in browser

gigabo:
Trying to debug a flash of unstyled content on the website.  Doesn't repro
locally, so want to get some debug output in production.


doug-wade:
🆒 
👍

PR #482, opened 2016-08-01
Make react a singleton client side with webpack resolve.alias

gigabo:
Getting closer to `react-server-cli` working as a global install.


gigabo:
Superseded by #484.

PR #484, opened 2016-08-01
Make `react-server-cli` work as a global install

gigabo:
Use the _caller's_ `react-server`.  Also, rename the binary to `react-server`.

```bash
$ npm i -g react-server-cli
$ react-server --port 3010 --js-port 3011
```

This supersedes #482.

### DIFF
--- a/packages/generator-react-server/test/app.js
+++ b/packages/generator-react-server/test/app.js
@@ -81,7 +81,15 @@ function installDeps() {
 			if (error) {
 				reject(error);
 			} else {
-				resolve();
+				const localDeps = ['react-server-cli', 'react-server']
+					.map(dep => path.resolve(path.join(__dirname, '../..', dep)));
+				cp.exec('npm install ' + localDeps.join(' '), (error) => {
+					if (error) {
+						reject(error);
+					} else {
+						resolve();
+					}
+				});
 			}
 		});
 	});
--- a/packages/react-server-cli/package.json
+++ b/packages/react-server-cli/package.json
@@ -40,9 +40,6 @@
     "webpack-stats-plugin": "^0.1.3",
     "yargs": "~3.15.0"
   },
-  "peerDependencies": {
-    "react-server": "^0.3.4"
-  },
   "devDependencies": {
     "nsp": "^2.4.0",
     "rimraf": "^2.5.2"
--- a/packages/react-server-cli/package.json
+++ b/packages/react-server-cli/package.json
@@ -10,7 +10,6 @@
   },
   "author": "Sasha Aickin",
   "bin": {
-    "react-server-cli": "./bin/react-server-cli",
     "react-server": "./bin/react-server-cli"
   },
   "license": "Apache-2.0",
--- a/packages/react-server-cli/package.json
+++ b/packages/react-server-cli/package.json
@@ -10,7 +10,7 @@
   },
   "author": "Sasha Aickin",
   "bin": {
-    "react-server-cli": "./bin/react-server-cli"
+    "react-server": "./bin/react-server-cli"
   },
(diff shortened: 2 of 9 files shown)


packages/generator-react-server/test/app.js
```diff
@@ -81,7 +81,15 @@ function installDeps() {
 			if (error) {
 				reject(error);
 			} else {
-				resolve();
+				const localDeps = ['react-server-cli', 'react-server']
```
doug-wade: 👍  this is way better

packages/react-server-cli/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "author": "Sasha Aickin",
   "bin": {
-    "react-server-cli": "./bin/react-server-cli"
+    "react-server": "./bin/react-server-cli"
```
doug-wade: `react-server startServer` is kinda cumbersome, though easily discoverable.  Iirc, you can have as many bin aliases as you want -- would it make sense to have an `rs` alias by default, or should we leave users to create their own aliases in `.bashrc` or `.zshrc`?

packages/react-server-cli/src/callerDependency.js
```diff
@@ -0,0 +1,7 @@
+import path from "path";
+
+export default function callerDependency(dep) {
+	// TODO: We should grab stuff based on what the routes file would get out
+	// of `require.resolve(dep)`.  Using `process.cwd()` instead for now.
+	return path.resolve(path.join(process.cwd(), "node_modules/" + dep));
```
doug-wade: Maybe use https://www.npmjs.com/package/look-up to look up the file system until you find `.reactserverrc` and treat that as the root of the project?

packages/react-server-cli/src/compileClient.js
```diff
@@ -165,6 +166,13 @@ const packageCodeForBrowser = (entrypoints, outputDir, outputUrl, hot, minify, l
 				},
 			],
 		},
+		resolve: {
+			alias: {
+				// These need to be singletons.
+				"react"        : callerDependency("react"),
```
doug-wade: Iiuc, if you want to break all your react components into modules that export the component, actions, stores &c, this means you'll all agree on the same react and react-server.  Should we also resolve `q`? `reflux`? `react-dom`?

packages/react-server-cli/package.json
gigabo: Unfortunately `rs` generally already exists:

```
$ man rs

RS(1)                     BSD General Commands Manual                    RS(1)

NAME
     rs -- reshape a data array
...
```

FWIW - I'm targeting `react-server start` rather than `react-server startServer`.  Doesn't require the shift key, at least. 😆

packages/react-server-cli/src/callerDependency.js
gigabo: Thank you!  That's exactly the wheel I was about to re-invent!

packages/react-server-cli/src/compileClient.js
gigabo: Hmm... yeah, with npm 2 that may wind up being necessary.  In npm 3 land, though, those _should_ be accounted for once these top level deps are handled.

I'd say let's only add others on an as-needed basis.

doug-wade: Not if you're doing local dev with `npm link` -- npm can't/doesn't/won't (one of those, not sure which) flatten the dependency tree across multiple modules.

gigabo: > `npm link`

image

doug-wade:
coupla things left outstanding, but nothing we need to hold up this pr over.

lgtm ⛵

PR #493, opened 2016-08-03
Making development compose file more similar to production

davidalber:
This change makes it possible to use the Docker Compose file in development again. I will document how to use this in a future PR.

### DIFF
--- a/packages/react-server-website/docker-compose.yml
+++ b/packages/react-server-website/docker-compose.yml
@@ -1,18 +1,27 @@
 version: '2'
 services:
-  node:
+  docs:
     build:
       context: .
       dockerfile: Dockerfile
+  slackin:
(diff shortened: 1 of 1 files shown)


gigabo:
How does one use this in development?

davidalber:
That's funny. I was positive I had addressed this in the commit message, but it's not there.

When I committed this I planned to call this out and say that it is on my list for a subsequent commit. It's a little more detailed than just "run this command".

gigabo:
Okay, uh... LGTM, then? 👍 

/me has no idea what this does.

davidalber:
Well, you don't have to merge it. I'm the only one impacted by this right now anyway.

I'll create this again when I have a chance to document it.

PR #494, opened 2016-08-03
Compressing static assets

### LINKED ISSUE #491: Need compression on the website
Can we just enable the nginx equivalent of `mod_deflate`?

Google page speed tool is complaining about uncompressed JS bundles.

image

davidalber:
Enabling ngx_http_gzip_module to compress assets on the way out.

This fixes #491.


gigabo:
Thanks @davidalber!  How about compressing the HTML output, too? 😁

PR #495, opened 2016-08-03
Eliminate the top-level react deps

gigabo:
These don't seem to be used, and GreenKeeper's foaming at the mouth over them.


doug-wade:
👍

PR #497, opened 2016-08-03
Changing port back to 3010

davidalber:
We were previously running the service on port 3010. We either needed to update the container and compose files or just change it back. We opted for changing it back.

### DIFF
--- a/packages/react-server-website/.reactserverrc
+++ b/packages/react-server-website/.reactserverrc
@@ -12,8 +12,7 @@
     "production": {
        "jsUrl": "/assets/",
        "logLevel": "error",
-       "minify": true,
-       "port": "80"
+       "minify": true
     }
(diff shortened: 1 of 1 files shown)


gigabo:
Whoops!  Thanks @davidalber!

doug-wade:
Can you update the [port testing manifest](https://github.com/redfin/react-server/blob/master/docs/testing-ports.md) to include that [this test](https://github.com/redfin/react-server/blob/master/packages/react-server-website/test.js) is using 3010/3011?

gigabo:
Oh, wait... the env-specific config just needs to be removed.  The _outer_ port is 3010.

davidalber:
@doug-wade: I created #499 to update the port testing manifest with the website's ports.

PR #501, opened 2016-08-04
Prevent another type of flash of unstyled content

gigabo:
Handles the case of a site is configured with a fully-qualified URL as a
static asset prefix.


doug-wade:
lgtm 👍

PR #503, opened 2016-08-04
Remove compile from the generated readme

doug-wade:
From @Noirbot bug report https://react-server.slack.com/archives/discussion/p1470273944000053


doug-wade:
The github inline editor is pretty cool

gigabo:
Thanks for reporting @Noirbot.

(And thanks for fixing @doug-wade 👍 )

PR #509, opened 2016-08-06
Redirecting requests to www.react-server.io to react-server.io

davidalber:
Additionally, I added some comments to the NGINX configuration to help organize it.

The PR diff isn't too clear. The commit that changes NGINX behavior is 	9d3d5ec.

### DIFF
--- a/packages/react-server-website/nginx.conf
+++ b/packages/react-server-website/nginx.conf
@@ -1,11 +1,17 @@
 proxy_cache_path /tmp/nginx levels=1:2 keys_zone=react_server_zone:10m inactive=60m;
 
-# Redirect requests to port 80 to HTTPS.
 server {
     listen 80 default_server;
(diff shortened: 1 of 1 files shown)


packages/react-server-website/nginx.conf
```diff
@@ -1,14 +1,30 @@
...
 }
 
 server {
+    # Redirect requests to https://*.react-server.io to
+    # https://react-server.io, except slack.react-server.io, which is handled
+    # separately below.
+    listen 443;
+    server_name *.react-server.io;
+
+    return 301 https://react-server.io$request_uri;
```
gigabo: Need to remember to add a server entry for a new domain here before we set up DNS to avoid accidental 301.

gigabo:
Thanks @davidalber. 👍

PR #513, opened 2016-08-09
add body-parser middleware for express

mattiasewers:
Adds the body-parser middleware to make POST requests to `react-server`.

### DIFF
--- a/packages/react-server-cli/src/commands/start.js
+++ b/packages/react-server-cli/src/commands/start.js
@@ -3,6 +3,7 @@ import https from "https"
 import express from "express"
 import pem from "pem"
 import compression from "compression"
+import bodyParser from "body-parser"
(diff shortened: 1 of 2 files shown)


packages/react-server-cli/src/commands/start.js
```diff
@@ -83,6 +84,8 @@ const startHtmlServer = (serverRoutes, port, httpsOptions) => {
 			logger.info("Starting HTML server...");
 
 			server.use(compression());
+			server.use(bodyParser.urlencoded({ extended: false }))
+			server.use(bodyParser.json())
```
gigabo: Nice.  We'll probably need to provide some way to pass configuration to bodyParser, eventually, but this will work great for now.

doug-wade:
woo!  thanks @mattiasewers 🚢 👍

PR #516, opened 2016-08-10
Removing the old container and volume during deployment

### LINKED ISSUE #511: Deployment does not get new static files
# Overview
The Docker volume being used to share static assets from the React Server Website service with NGINX is persisting between deployments. This results in new assets not actually being deployed to NGINX when you do an "in-place" deployment (i.e., a deployment that does not deploy to a new instance using deploy.yml).

# Reproduction
For good measure, you might want to remove containers, volumes, and images you already have:

1. `docker rm -v $(docker ps -aq)`
1. `docker volume rm $(docker volume ls -q) # in case you have any unassociated volumes`
1. `docker rmi $(docker images -q)`

Now the reproduction:

1. Check out a fresh copy of the repository.
1. `cd react-server`
1. `docker build -t reactserver/react-server .`
1. `cd packages/react-server-website`
1. `SLACK_API_TOKEN=YOUR_SLACK_TOKEN docker-compose up -d`

At this point we can look at the static assets created by the website container and what's in the volume shared with NGINX.

1. `docker exec -i -t reactserverwebsite_docs_1 "/bin/bash"`
1. `shasum __clientTemp/build/common.bundle.js` returns as hash of `b92b2ae8ae8573c5913f3e46b142ecbeffca11a1`.
1. `shasum /www/data/assets/common.bundle.js` returns a hash of `b92b2ae8ae8573c5913f3e46b142ecbeffca11a1`.

Finally, we can make sure NGINX sees the same thing.

1. `docker exec -i -t reactserverwebsite_nginx_1 "/bin/bash"`
1. `shasum /www/data/assets/common.bundle.js` returns a hash of `b92b2ae8ae8573c5913f3e46b142ecbeffca11a1`.

Looks good (i.e., all of the hashes were the same)!

Next, let's change something that will modify common.bundle.js. For this example, I am changing "https://slack.react-server.io/" in components/Header.js to "https://slackack.react-server.io/". Save the changes. Now restart this service.

1. `SLACK_API_TOKEN=YOUR_SLACK_TOKEN docker-compose up --build -d`

# Expected
[http://localhost:8443/assets/common.bundle.js](http://localhost:8443/assets/common.bundle.js) will now contain the string "https://slackack.react-server.io/".

# Actual
[http://localhost:8443/assets/common.bundle.js](http://localhost:8443/assets/common.bundle.js) still has "https://slack.react-server.io/". Let's look in the documentation service again:

1. `docker exec -i -t reactserverwebsite_docs_1 "/bin/bash"`
1. `shasum __clientTemp/build/common.bundle.js` returns as hash of `3c4f9c49a4566a71358458009f717ef3b988c4c9`. Additionally, it contains "https://slackack.react-server.io/".
1. `shasum /www/data/assets/common.bundle.js` returns a hash of `b92b2ae8ae8573c5913f3e46b142ecbeffca11a1`. _Note: that's not the same hash as in the step above._

Thus, NGINX is serving the old common.bundle.js. This can be remedied by doing the following.

1. `docker-compose stop docs`
1. `docker rm -v reactserverwebsite_docs_1`
1. `SLACK_API_TOKEN=YOUR_SLACK_TOKEN docker-compose up --build -d`

Now we find "https://slackack.react-server.io/" in [http://localhost:8443/assets/common.bundle.js](http://localhost:8443/assets/common.bundle.js). Furthermore, the hashes between the build common.bundle.js and the common.bundle.js file in the volume used by NGINX are the same.

1. `docker exec -i -t reactserverwebsite_docs_1 "/bin/bash"`
1. `shasum __clientTemp/build/common.bundle.js` returns as hash of `3c4f9c49a4566a71358458009f717ef3b988c4c9`. Additionally, it contains "https://slackack.react-server.io/".
1. `shasum /www/data/assets/common.bundle.js` returns a hash of `3c4f9c49a4566a71358458009f717ef3b988c4c9`.

The deploy.yml file needs to be updated to use the approach above or some other way to cycle the volume.

davidalber:
- Removing the old docs service container and volume during deployment. This fixes #511.
- Using Ansible's `command` module, instead of `shell`, when possible.

### DIFF
--- a/packages/react-server-website/deployment/deploy.yml
+++ b/packages/react-server-website/deployment/deploy.yml
@@ -12,15 +12,15 @@
       copy: src=../nginx.conf dest=/home/ubuntu/nginx.conf mode=0644
 
     - name: Pull images
-      shell: docker-compose pull
+      command: docker-compose pull
       become: true
 
     - name: Stop services
-      shell: docker-compose stop
+      command: docker-compose stop
       become: true
 
     - name: Remove docs container and volume
-      shell: docker-compose rm -vf docs
+      command: docker-compose rm -vf docs
       become: true
 
     - name: Restart services
--- a/packages/react-server-website/deployment/deploy.yml
+++ b/packages/react-server-website/deployment/deploy.yml
@@ -11,10 +11,18 @@
     - name: Copy NGINX configuration
       copy: src=../nginx.conf dest=/home/ubuntu/nginx.conf mode=0644
 
-    - name: Pull containers
-      shell: docker-compose pull && docker-compose build
+    - name: Pull images
+      shell: docker-compose pull
+      become: true
+
+    - name: Stop services
+      shell: docker-compose stop
+      become: true
+
+    - name: Remove docs container and volume
+      shell: docker-compose rm -vf docs
       become: true
 
     - name: Restart services
-      shell: docker-compose stop && SLACK_API_TOKEN=`cat slack-api-token | tr -d '\n'` docker-compose up -d
+      shell: SLACK_API_TOKEN=`cat slack-api-token | tr -d '\n'` docker-compose up -d
       become: true


packages/react-server-website/deployment/deploy.yml
```diff
@@ -11,10 +11,18 @@
     - name: Copy NGINX configuration
       copy: src=../nginx.conf dest=/home/ubuntu/nginx.conf mode=0644
 
-    - name: Pull containers
-      shell: docker-compose pull && docker-compose build
+    - name: Pull images
+      command: docker-compose pull
```
doug-wade: previously, we docker-compose pulled AND docker-compose built, but now it seems like we don't build anymore?  Is that expected?

davidalber: Yes, the `build` step was an attempt to address #511 before I had reproduced it or understood what was going on. In fact, doing a `build` there has no effect because we are deploying pre-built images.

doug-wade:
lgtm 👍

PR #518, opened 2016-08-10
Eliminate a frameback navigation initialization race

gigabo:
During the first frameback navigation forward with `reuseFrame` set, if the
client controller within the frame completes before the frame's load event
fires we were previously failing to wire up our listener to handle subsequent
navigations.

### DIFF
--- a/packages/react-server/core/FramebackController.js
+++ b/packages/react-server/core/FramebackController.js
@@ -289,12 +289,15 @@ class FramebackController extends EventEmitter {
 		// If the frame has a client controller we'll listen to when
 		// _it_ tells us it's done loading.
 		//
+		if (clientController && !clientController.__parentIsListening) {
+			clientController.__parentIsListening = true;
+			clientController.context.onLoadComplete(this._handleFrameLoad.bind(this, frame));
+		}
+
 		// It may have finished _before_ we get here, so we'll also check
 		// whether it has already set its `_previouslyRendered` flag.
 		//
-		if (clientController && !clientController.__parentIsListening && !clientController._previouslyRendered) {
-			clientController.__parentIsListening = true;
-			clientController.context.onLoadComplete(this._handleFrameLoad.bind(this, frame));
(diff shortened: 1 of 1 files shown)


packages/react-server/core/FramebackController.js
```diff
@@ -289,12 +289,15 @@ class FramebackController extends EventEmitter {
...
+			clientController.context.onLoadComplete(this._handleFrameLoad.bind(this, frame));
+		}
+
 		// It may have finished _before_ we get here, so we'll also check
 		// whether it has already set its `_previouslyRendered` flag.
 		//
-		if (clientController && !clientController.__parentIsListening && !clientController._previouslyRendered) {
-			clientController.__parentIsListening = true;
-			clientController.context.onLoadComplete(this._handleFrameLoad.bind(this, frame));
+		if (clientController && !clientController._previouslyRendered) {
```
doug-wade: Can this be an `else if` block?  or even an `else` block?  `clientController.__parentIsListening` is set on the first load, as is `clientController._previouslyRendered` -- are the two separate checks because there's asynchrony and users may perform navigations before we've finished loading?

gigabo: > Can this be an else if block? or even an else block?

It can't, actually.  You'll notice that _previously_ the bodies of the two cases were _together_, and the conditionals were _anded_.  In the case where the frame's load event fires before the client controller finishes rendering we still need both to happen.

gigabo: Basically: We _always_ need to wire up our listener (_once_).  And, in the case where the event we're listening to hasn't fired yet, we'll bail out early and wait for it.

gigabo: > there's asynchrony and users may perform navigations before we've finished loading?

It's actually just a race between the browser's `load` event and the client controller finishing its render.  No user interaction needed.

doug-wade:
lgtm 👍

PR #522, opened 2016-08-11
gh-pages: npm run compile does not exist

rwoody:
Looks like this was not removed after some changes


doug-wade:
Jeez, how many places did we write the same command 🐑 
Thanks @rwood, lgtm

PR #523, opened 2016-08-11
Declare chalk dependency

doug-wade:
There's an [undeclared dependency on chalk](https://github.com/redfin/react-server/blob/c3533db335279b4c4ace585fb1bfdd6745afd7fa/packages/react-server-cli/src/run.js#L2) in react-server-cli.  Add it to package.json.


gigabo:
Oops... 😇 

Thanks @doug-wade!

PR #527, opened 2016-08-11
Spell née correctly

nattofriends:
"neé" is actually written née [1]. 

[1] https://en.wikipedia.org/wiki/Given_name#Name_at_birth


gigabo:
Whoa, nice catch @nattofriends!

PR #528, opened 2016-08-11
Don't require node 6 to run react-server-cli

lidawang:
cc @gigabo

### DIFF
--- a/packages/react-server-cli/src/cli.js
+++ b/packages/react-server-cli/src/cli.js
@@ -1,5 +1,5 @@
 require("babel-core/register");
 
-const {run, parseCliArgs} = require(".");
+const cli = require(".");
 
-run(parseCliArgs());
+cli.run(cli.parseCliArgs());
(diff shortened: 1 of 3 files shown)


gigabo:
Thanks @lidawang!  Will be nice to pick up https://github.com/lerna/lerna/pull/192 when it lands so we can drop the extra two copies of `cli.js` here. 😛

doug-wade:
This seems like its going backwards to me.  Node 4 supports destructuring, but its behind a flag; can you try adding the `--harmony_destructuring` flag to https://github.com/redfin/react-server/blob/master/packages/react-server-cli/bin/react-server-cli#L1, rather than removing the destructuring?  I don't think we want to try to support node < 4.

gigabo:
@doug-wade - This is actually just the top-most entry point for `react-server-cli`.  We have a couple of extra copies laying around until the next version of Lerna lands, but after that it will just be this one tiny file that needs this treatment. :)

doug-wade:
@gigabo oh, I see.  sorry @lidawang, lgtm 👍

PR #529, opened 2016-08-11
Force Linguist to exclude `.md` and `.py` files from our stats.

egid:
After changes, here's what Linguist spits out locally:

```
48.06%  JavaScript  
44.40%  CSS  
5.33%   Nginx  
2.22%   Shell
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 * text=auto
+*.py linguist-vendored
+*.md linguist-documentation


gigabo:
> 44.40%  CSS

image


Eh, close enough. 😁 

Thanks @egid!

doug-wade:
Js is the plurality? good enough for government work 🏛

PR #531, opened 2016-08-11
Upgrade to lerna-2.0.0-beta.25

gigabo:
This picks up https://github.com/lerna/lerna/pull/192 which lets us simplify our use of the CLI within the repo.


doug-wade:
🚢

PR #532, opened 2016-08-11
Allow RootElements and RootContainers to specify their node type

kcjonson:
- Implements `tagName` attribute on `RootElement` and `RootContainer` that specifies what tag the element or container will render with.  
- Changes `ClientController` to search for any node type with matching data attribute (not just div)

Example usage:

```
getElements() {
    return <RootContainer tagName="section">
        <h2>Hello</h2>
        <RootElement className="content">
            <HelloWorld />
        </RootElement>
    </RootContainer>;
}
```


packages/react-server/core/ClientController.js
```diff
@@ -650,7 +650,11 @@ class ClientController extends EventEmitter {
 
 			_.forEach(
 				getRootElementAttributes(element),
-				(v, k) => root.setAttribute(k, v)
+				(v, k) => {
+					if (k !== 'tagName') {
+						root.setAttribute(k, v)
+					}
+				}
```
gigabo: This bit actually isn't necessary.  The attributes we pass through are [white listed in RootElement.js](https://github.com/redfin/react-server/blob/master/packages/react-server/core/components/RootElement.js#L102-L105).

gigabo: Oh, I hadn't gotten far enough... looks like it's being added as an attr there now.  Hmm...

kcjonson: Yea, its a kinda ugly bit of code.  No harm arises by putting the attribute on the node, but its redundant information because the node with be of the type specified.

packages/react-server/core/components/RootElement.js
```diff
@@ -103,7 +103,7 @@ RootElement.getRootElementAttributes = function(element) {
 		'id',
 		'style',
 	].forEach(k => props[k] && (attrs[k] = props[k]));
-
+	attrs.tagName = props.tagName || 'div';
```
gigabo: I wouldn't bundle this in with `attrs`.  That's really meant to just be HTML attributes.

The tag name is a different beast.

packages/react-server/core/components/RootContainer.js
```diff
@@ -23,9 +23,11 @@ RootContainer.defaultProps = {
 }
 
 RootContainer.flattenForRender = function(element) {
-	return [{containerOpen: getRootElementAttributes(element)}]
+	let attrs = getRootElementAttributes(element) || {};
+	attrs.tagName = attrs.tagName || 'div';
+	return [{containerOpen: true, attrs: attrs}]
```
gigabo: Something like `{containerOpen: tagName, attrs}` would be nicer here, to keep `attrs` and `tagName` separate.

packages/react-server/core/components/RootContainer.js
```diff
@@ -23,9 +23,11 @@ RootContainer.defaultProps = {
...
 		.concat(prepChildren(element))
-		.concat([{containerClose: true}])
+		.concat([{containerClose: true, attrs: attrs}])
```
gigabo: Then this could just be `{containerClose: tagName}`, and the (irrelevant for a close tag) `attrs` can be omitted.

packages/react-server/core/components/RootElement.js
kcjonson: Yea, I was waffling a bit on that one, flipped a coin and came out this way.  I think I agree with you that the other way is easer to read.

gigabo:
I like this.  It's going to be a cool improvement!

doug-wade:
🚀 i dig it

gigabo:
Hi @kcjonson, I just took another pass through here and realized that the root elements created during client transitions will also need to respect `tagName`.  Take a look at [`_createReactServerRootNode`](https://github.com/redfin/react-server/blob/master/packages/react-server/core/ClientController.js#L758) and [`_createContainerNode`](https://github.com/redfin/react-server/blob/master/packages/react-server/core/ClientController.js#L765) in ClientController.js.

It would be good to have a test page set up to play around with this.  You could either create a new page, or modify an existing page like the [navigation playground](https://github.com/redfin/react-server/blob/master/packages/react-server-test-pages/pages/navigation/playground.js) to use non-div root containers/elements.  That could actually really simplify the playground, which is table-like, but uses CSS on divs to lay it out.  The other benefit there is you can use it to test all of the navigation variants (client transitions with and without DOM reuse, etc).

So, I think the remaining things I'd like to see here are:

- [ ] Separate `tagName` from `attrs`
- [ ] Support client transitions
- [ ] Add a test page that uses `tagName` (or modify an existing page)

BTW - You'll probably want to pull and bootstrap to pick up #533 before you work on test pages.

kcjonson:
Good catch on the transition bug, will address that and moving the tagName out of attrs which was a dubiously good decision to put them there in the first place.

Will put some time into the playground too, as you know CSS/HTML is my jam.

doug-wade:
@kcjonson looks like you may have committed as a different user than your github user -- maybe as your work user?  Its possible adding another email address to your gh profile would resolve this, or `git commit --amend --author`

drewpc:
Closing this PR since there is no movement.

PR #533, opened 2016-08-12
Don't require CLI users to install webpack locally

gigabo:
We already have everything we need as deps of `react-server-cli`.

Don't make _users_ of the CLI _also_ install that stuff.


doug-wade:
Awesome! 👍

Issue #540, opened 2016-08-12
In generator-react-server  v0.4.2  dependencies  "react-server": "^0.3.4" !


AlexMistVrn:
After updating to v0.4.2, package.json remained in the old version.

doug-wade:
closed in #541

PR #541, opened 2016-08-12
#540 update react-server in generator-react-server

doug-wade:
Generate a package that depends on the most recent react server


gigabo:
Thanks @doug-wade. 👍

PR #542, opened 2016-08-12
Add linting to babel plugin

doug-wade:
The point of linting is that everything should look and read the same, so it kinda defeats the purpose when we don't do it for some packages.

### DIFF
--- a/packages/babel-plugin-react-server/package.json
+++ b/packages/babel-plugin-react-server/package.json
@@ -18,7 +18,7 @@
   "scripts": {
     "clean": "rm -rf lib npm-debug.log*",
     "build": "babel src -d lib",
-    "test": "mocha --compilers js:babel-register",
+    "test": "mocha --compilers js:babel-register && eslint src/",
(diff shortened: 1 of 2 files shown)


gigabo:
Thanks @doug-wade.  Dropping the `enhancement` label here.  Doesn't improve the functionality of the plugin.  Does clean it up nicely, though! 👍 

Do we need a `devDependency` on `eslint` here, since it's using the binary?

doug-wade:
@gigabo it uses the eslint at the monorepo root, which we prefer because it keeps bootstrap times and disk utilisation low

gigabo:
Yeah, it's preferable for dev dependencies _unless_ we use them in an npm script. 😁 

Did some testing the other day and discovered that `lerna run ...` doesn't actually have any magic for making binaries from the repo root available in packages' npm scripts.  Guess it works _sometimes_, though?  Pretty confused by the whole situation.

I guess as long as it's passing in travis... 👍

PR #546, opened 2016-08-13
Updating .dockerignore and changing Slack interval

davidalber:
- Updating .dockerignore with deployment directory and ignoring docker-compose.yml files.
- Increasing Slack fetch interval from five seconds to thirty seconds.


davidalber:
This is cleanup and modification to the deployment. Merging immediately.

PR #560, opened 2016-08-15
#314 Pass parameters to module tagger explicitly, rather than in function context

doug-wade:
Smooth out the rough edges of the module tagger interface


doug-wade:
builds on #558

gigabo:
Nice, this is a saner interface. ;)

👍

PR #561, opened 2016-08-15
Fix #12 Support configurable tokens for the gulp module tagger

### LINKED ISSUE #12: Pass in the token to be replaced in the module tagger

doug-wade:
Support configurable tokens for the gulp module tagger.  Fixes #12

### DIFF
--- a/packages/react-server-gulp-module-tagger/index.js
+++ b/packages/react-server-gulp-module-tagger/index.js
@@ -6,13 +6,19 @@ var replace = require("gulp-replace")
 //   - "__LOGGER__"
 //   - "__LOGGER__({ /* options */ })"
 var isWindows = ('win32' === process.platform)
-,   REPLACE_TOKEN = /(?:__LOGGER__|__CHANNEL__|__CACHE__)(?:\(\s*(\{[\s\S]*?\})\s*\))?/g
+,   DEFAULT_REPLACE_TOKEN = tokenToRegExp("__LOGGER__|__CHANNEL__|__CACHE__")
 ,   THIS_MODULE   = isWindows
 	? /(?:[^\\]+\\node_modules\\)?react-server-gulp-module-tagger\\index\.js$/
 	: /(?:[^\/]+\/node_modules\/)?react-server-gulp-module-tagger\/index\.js$/
 
 module.exports = function(config) {
 	config || (config = {});
+	var REPLACE_TOKEN;
+	if (config.token) {
+		REPLACE_TOKEN = tokenToRegExp(config.token);
+	} else {
+		REPLACE_TOKEN = DEFAULT_REPLACE_TOKEN;
+	}
 	config.basePath = module.filename.replace(THIS_MODULE,'');
 	return forEach(function(stream, file){
 		return stream.pipe(replace(REPLACE_TOKEN, (match, optString) => {
@@ -23,4 +29,8 @@ module.exports = function(config) {
 			});
 		}));
 	});
+}
+
+function tokenToRegExp(token) {
+	return new RegExp("(?:" + token + ")(?:\\(\\s*(\\{[\\s\\S]*?\\})\\s*\\))?", 'g');
 }
--- a/packages/react-server-gulp-module-tagger/test/fixtures/custom-token/gulpfile.js
+++ b/packages/react-server-gulp-module-tagger/test/fixtures/custom-token/gulpfile.js
@@ -1,0 +1,10 @@
+const gulp = require('gulp');
+const tagger = require('../../..');
+
+gulp.task('default', () => {
+	gulp.src('actual.js')
+		.pipe(tagger({
+			token: "__OZZIE_ALONSO__",
+		}))
+		.pipe(gulp.dest('build'));
+});
--- a/packages/react-server-gulp-module-tagger/test/fixtures/custom-token/actual.js
+++ b/packages/react-server-gulp-module-tagger/test/fixtures/custom-token/actual.js
@@ -1,0 +1,3 @@
+var logger = require('react-server').logging.getLogger(__OZZIE_ALONSO__);
+var fooLogger = logging.getLogger(__OZZIE_ALONSO__({ label: "foo" }));
+var barLogger = logging.getLogger(__OZZIE_ALONSO__({ label: "bar" }));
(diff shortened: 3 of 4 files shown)


doug-wade:
builds on #560

gigabo:
The stack's getting too tall here.  Will review when #560 lands.

doug-wade:
@gigabo #560 merged and I rebased onto latest master

packages/react-server-gulp-module-tagger/index.js
```diff
@@ -6,13 +6,19 @@ var replace = require("gulp-replace")
...
+,   DEFAULT_REPLACE_TOKEN = /(?:__LOGGER__|__CHANNEL__|__CACHE__)(?:\(\s*(\{[\s\S]*?\})\s*\))?/g
 ,   THIS_MODULE   = isWindows
 	? /(?:[^\\]+\\node_modules\\)?react-server-gulp-module-tagger\\index\.js$/
 	: /(?:[^\/]+\/node_modules\/)?react-server-gulp-module-tagger\/index\.js$/
 
 module.exports = function(config) {
 	config || (config = {});
+	var REPLACE_TOKEN;
+	if (config.token) {
+		REPLACE_TOKEN =  new RegExp("(?:" + config.token + ")(?:\\(\\s*(\\{[\\s\\S]*?\\})\\s*\\))?", 'g');
```
gigabo: Don't like having the regexp machinery duplicated...

Maybe this could be in a function, and the default could be defined using it?

Like:

```javascript
DEFAULT_REPLACE_TOKEN = tokenToRegexp("__LOGGER__|__CHANNEL__|__CACHE__")
```

packages/react-server-gulp-module-tagger/test/fixtures/custom-token/gulpfile.js
```diff
@@ -0,0 +1,10 @@
+const gulp = require('gulp');
+const tagger = require('../../..');
+
+gulp.task('default', () => {
+	gulp.src('actual.js')
+		.pipe(tagger({
+			token: "__OZZIE_ALONSO__",
```
gigabo: image

packages/react-server-gulp-module-tagger/index.js
doug-wade: I didn't reuse the second capture group based on this advice from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp)

>The literal notation provides compilation of the regular expression when the expression is evaluated. Use literal notation when the regular expression will remain constant. For example, if you use literal notation to construct a regular expression used in a loop, the regular expression won't be recompiled on each iteration.

>The constructor of the regular expression object, for example, new RegExp('ab+c'), provides runtime compilation of the regular expression. Use the constructor function when you know the regular expression pattern will be changing, or you don't know the pattern and are getting it from another source, such as user input.

The benefit of using a pre-compiled regex [is ~10%](http://www.levihackwith.com/javascript-performance-tip-precompile-your-regular-expressions/) -- if the code reuse tradeoff seems worth the lost performance in the default case, I'm more than happy to make the change.

gigabo: Yeah, that's a good consideration.  I think in this case we can have our cake and eat it, too, though.  So long as `DEFAULT_REPLACE_TOKEN` stays at the package level where it is, the penalty for running it through the `RegExp` constructor will only be paid once.  After that it's just a compiled RegExp.  What we want to avoid for performance reasons is `new RegExp(...)` at _run time_.

packages/react-server-gulp-module-tagger/index.js
```diff
@@ -6,13 +6,17 @@ var replace = require("gulp-replace")
 //   - "__LOGGER__"
 //   - "__LOGGER__({ /* options */ })"
 var isWindows = ('win32' === process.platform)
-,   REPLACE_TOKEN = /(?:__LOGGER__|__CHANNEL__|__CACHE__)(?:\(\s*(\{[\s\S]*?\})\s*\))?/g
```
doug-wade: @gigabo like this?

packages/react-server-gulp-module-tagger/index.js
```diff
@@ -6,13 +6,17 @@ var replace = require("gulp-replace")
...
 ,   THIS_MODULE   = isWindows
 	? /(?:[^\\]+\\node_modules\\)?react-server-gulp-module-tagger\\index\.js$/
 	: /(?:[^\/]+\/node_modules\/)?react-server-gulp-module-tagger\/index\.js$/
 
 module.exports = function(config) {
 	config || (config = {});
+	var REPLACE_TOKEN;
+	if (config.token) {
+		REPLACE_TOKEN = tokenToRegExp(config.token);
+	}
```
gigabo: This looks like it _masks_ the outer `REPLACE_TOKEN`?

Needs a different name inside this function, right?

```javascript
var token = REPLACE_TOKEN;
if (config.token) {
    token = rokenToRegExp(config.token);
}
```

doug-wade: we want it to _replace_ the outer `REPLACE_TOKEN`.

gigabo: But... don't we still want the value of the outer `REPLACE_TOKEN` as a default if `config.token` isn't passed in?

doug-wade: Yeah, which is why the `if (config.token) { ... }` check guards against replacing it

gigabo: But the `var REPLACE_TOKEN` declares a _new variable_ inside the function that masks the outer variable, making it no longer accessible...?

gigabo: Oh, it's been updated again since this comment thread started. :)

packages/react-server-gulp-module-tagger/index.js
```diff
@@ -6,13 +6,16 @@ var replace = require("gulp-replace")
...
-,   REPLACE_TOKEN = /(?:__LOGGER__|__CHANNEL__|__CACHE__)(?:\(\s*(\{[\s\S]*?\})\s*\))?/g
+,   REPLACE_TOKEN = tokenToRegExp("__LOGGER__|__CHANNEL__|__CACHE__")
 ,   THIS_MODULE   = isWindows
 	? /(?:[^\\]+\\node_modules\\)?react-server-gulp-module-tagger\\index\.js$/
 	: /(?:[^\/]+\/node_modules\/)?react-server-gulp-module-tagger\/index\.js$/
 
 module.exports = function(config) {
 	config || (config = {});
+	if (config.token) {
+		REPLACE_TOKEN = tokenToRegExp(config.token);
```
gigabo: So... it looks like this replaces the default.  If I use the module tagger once with `config.token` and then later once without I'll get the previous invocation's `config.token` value instead of the default?

doug-wade: new diff

gigabo:
Looks awesome to me! :metal: 

Thanks for humoring my obsession with factoring out the regexp line noise @doug-wade. 😅

PR #568, opened 2016-08-16
Fixed bug in https handling

CreepGin:
The default `return startServers()` needs to be wrapped in an `else` block because the first `return` statement was done in a callback (inside `pem.createCertificate`).


gigabo:
Oooh, nice catch @CreepGin!

Better still would be to have a consistent return value in the `if (httpsOptions === true)` case, but that may not be straightforward to get patched through...

This at least avoids trying to start everything twice. 😆

gigabo:
^ File #569 to track that.

Thanks @CreepGin!

PR #573, opened 2016-08-17
Add support to Link for href

### LINKED ISSUE #380: Link should accept `href` as a prop
Right now, `<Link/>` expects a `path`. But it's emulating `<a/>`, which uses `href`. 

It'd be a useful change to allow Link to accept either.

doug-wade:
Fixes #380


gigabo:
Nice. 👍

PR #583, opened 2016-08-18
Fix react-server-cli module tags

doug-wade:
Previously

screen shot 2016-08-17 at 6 15 28 pm

Now

screen shot 2016-08-17 at 6 16 14 pm

I don't dig that we're tagging the modules with their _source_ -- it seems like it would be much more elegant to tag them with their _destination_ (in this case, `react-server-cli.commands.start` instead of `react-server-cli.src.commands.start`).  I think the easiest way will be to trim the `react-server-cli.src` with the existing `trim` option, and add a `prefix` option, but there may be a more elegant way.

### DIFF
--- a/packages/react-server-cli/gulpfile.babel.js
+++ b/packages/react-server-cli/gulpfile.babel.js
@@ -5,7 +5,7 @@ import logging from "react-server-gulp-module-tagger";
 
 gulp.task("default", () => {
 	return gulp.src("src/**/*.js")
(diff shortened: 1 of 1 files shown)


doug-wade:
Oh, and I'm hoping its a coincidence that `react-server-cli.src.commands.start` and `react-server-cli.src.logProductionWarnings` are apparently using the same color 😓

gigabo:
👍 

Also 👍  to the trim/prefix idea.

PR #587, opened 2016-08-18
Fire a "pageview" event from the client controller

### LINKED ISSUE #502: Navigation event for Analytics
Create a navigation event that fires as soon as page title is set.

gigabo:
Analytics virtual pageviews need path and page title, so fire an event with
those as soon as they're available on client transition.

Note that on frameback navigation the event will fire _in the frame_.

This closes #502.

Also updated the website's analytics middleware to use the new event.

Next step: Pull the website's analytics middleware out into a dedicated
package and make it public. :rocket:

### DIFF
--- a/packages/react-server-website/middleware/Analytics.js
+++ b/packages/react-server-website/middleware/Analytics.js
@@ -8,19 +8,22 @@ ga('create', 'UA-81160060-1', 'auto');
 ga('send', 'pageview');
 `.replace(/\s*\n\s*/g, '');
 
+let alreadyListening = 0;
+
 export default class AnalyticsMiddleware {
 
-	// This only gets called after client transitions in the browser.  It's
-	// not called for the initial page view (since the title was set by the
-	// server).
-	getTitle(next) {
-		return next().then(title => {
-			if (typeof window !== "undefined") {
-				const page = location.pathname;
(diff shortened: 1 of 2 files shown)


doug-wade:
I don't think we mention adding event listeners to the client controller anywhere in the docs, let alone which events might be emitted.  We should probably do that.

gigabo:
Hmm... yeah, I think this is the first event that's actually worth listening to from the client controller.

A doc on analytics would be good.  Makes sense to ship with the analytics middleware to serve as an example, I think.

doug-wade:
Dear future intrepid soul searching for docs on events emitted by React Server,

We emit the following events: render, showFrame, loadComplete, page, navigateDone, pageview.

Good luck,
<3 The React Server Team

gigabo:
> adding event listeners to the client controller
> ...
> We emit the following events: render, showFrame, loadComplete, page, navigateDone, pageview.

Only `render` is emitted by the client controller...?

doug-wade:
yeah, that's all of them from everywhere -- frameback controller, navigator, the whole lot

gigabo:
Gotcha.  Yeah, maybe a doc on events in general would be good.

You know... to go along with our thorough API docs. 👹

PR #591, opened 2016-08-19
Use sass loader in the bike share example

doug-wade:
Yay! 🎉

### DIFF
--- a/packages/react-server-examples/bike-share/package.json
+++ b/packages/react-server-examples/bike-share/package.json
@@ -7,13 +7,10 @@
   "author": "Doug Wade <doug.wade@redfin.com>",
   "scripts": {
     "clean": "rm -rf build __clientTemp",
-    "start": "npm run clean && npm run styles && react-server start",
-    "styles": "node-sass styles/index.scss build/styles/index.css && node-sass styles/network.scss build/styles/network.css",
-    "test": "xo && nsp check"
+    "start": "npm run clean && react-server start --hot"
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "babel-runtime": "^6.6.1",
     "react": "~0.14.2",
     "react-dom": "~0.14.2",
     "react-server": "^0.4.5",
@@ -21,21 +18,6 @@
     "superagent": "1.2.0"
   },
   "devDependencies": {
-    "babel-preset-react-server": "^0.4.4",
-    "eslint-config-xo-react": "^0.7.0",
-    "eslint-plugin-react": "^5.1.1",
-    "node-sass": "^3.7.0",
-    "nsp": "^2.3.3",
-    "xo": "^0.15.1"
-  },
-  "xo": {
-    "esnext": true,
-    "extends": "xo-react",
-    "globals": [
-      "__LOGGER__"
-    ],
-    "ignores": [
-      "__clientTemp/**/*"
-    ]
+    "babel-preset-react-server": "^0.4.4"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/react-server-examples/bike-share/package.json
```diff
@@ -7,35 +7,21 @@
...
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "babel-runtime": "^6.6.1",
+    "babel-runtime": "^6.11.6",
     "react": "~0.14.2",
     "react-dom": "~0.14.2",
     "react-server": "^0.4.5",
     "react-server-cli": "^0.4.5",
+    "sass-loader": "^4.0.0",
```
doug-wade: this feels like a bug in react-server-cli, tbqh

packages/react-server-examples/bike-share/package.json
```diff
@@ -7,35 +7,21 @@
...
   },
   "devDependencies": {
     "babel-preset-react-server": "^0.4.4",
-    "eslint-config-xo-react": "^0.7.0",
-    "eslint-plugin-react": "^5.1.1",
-    "node-sass": "^3.7.0",
-    "nsp": "^2.3.3",
-    "xo": "^0.15.1"
-  },
-  "xo": {
```
doug-wade: This is a react-server example, keep it as simple as possible.

packages/react-server-examples/bike-share/package.json
```diff
@@ -7,35 +7,21 @@
   "author": "Doug Wade <doug.wade@redfin.com>",
   "scripts": {
     "clean": "rm -rf build __clientTemp",
-    "start": "npm run clean && npm run styles && react-server start",
-    "styles": "node-sass styles/index.scss build/styles/index.css && node-sass styles/network.scss build/styles/network.css",
-    "test": "xo && nsp check"
+    "start": "npm run clean && react-server start --hot",
```
gigabo: Wow, so much cleaner!

Should we move the `hot` into a `.reactserverrc`?

packages/react-server-examples/bike-share/package.json
```diff
@@ -7,35 +7,21 @@
...
+    "test": "nsp check"
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "babel-runtime": "^6.6.1",
+    "babel-runtime": "^6.11.6",
```
gigabo: This _shouldn't_ even be necessary?

Should only need the packages that are installed by [`react-server init`](https://github.com/redfin/react-server/blob/master/packages/react-server-cli/src/commands/init.js#L8-L16).

packages/react-server-examples/bike-share/package.json
gigabo: Yeah, this shouldn't be necessary either. 😕

gigabo: Have you run `npm i` in `packages/react-server-cli` since the dep was added there?

gigabo:
This LGTM, though I'm confused by the babel/sass-loader deps.

packages/react-server-examples/bike-share/package.json
doug-wade: It seems like a lot of overhead to add a whole new file to eliminate one flag...

doug-wade:
@gigabo pulled the sass and babel deps.  I had to run a relative install (`npm i ../react-server-cli`) to get it to pass locally -- `lerna bootstrap` didn't seem to fix it -- and I'm not sure what that means about whether we should merge this.

gigabo:
Yeah, `lerna bootstrap` can't help us with the examples ([yet](https://github.com/lerna/lerna/issues/126)) because they're not direct children of `packages/`.  But... the install from `^0.4.5` should work. 😕

PR #602, opened 2016-08-22
Fix module tagging for react server core

doug-wade:
Still working on a repro to make sure this resolves the issue


doug-wade:
Repro'd -- it's fixed, I promise.

gigabo:
Thanks @doug-wade!

PR #603, opened 2016-08-22
Fix react-server-data-bundle-cache logging

doug-wade:
We trimmed this string in a lot of places...


doug-wade:
Please don't kick the build; I'm filing an issue with nsp

gigabo:
image

doug-wade:
Got a response from nsp:

>>> Apologies for the trouble. We've been experiencing some instabilities related to infrastructure lately. We're working diligently to address these issues and scale up so we can meet demand. Thank you for your patience.

Guess #605 is more important than I thought.  Restarting the build.

PR #609, opened 2016-08-23
Add auto-incrementing keys to RootElements

### LINKED ISSUE #588: Handle auto-incrementing keys at the root level automatically
Get rid of the tedious `key={2}`, `key = {3}` https://github.com/redfin/react-server/blob/master/packages/react-server-examples/bike-share/pages/index.js#L29

doug-wade:
Fixes #588

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -764,6 +764,9 @@ class ClientController extends EventEmitter {
 	_createReactServerRootNode(mountNode, index) {
 		var root = document.createElement("div");
 		root.setAttribute(REACT_SERVER_DATA_ATTRIBUTE, index);
+		if (!root.getAttribute("key")) {
+			root.setAttribute("key", index);
+		}
 		mountNode.appendChild(root);
 		return root;
 	}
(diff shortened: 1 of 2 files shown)


packages/react-server/core/ClientController.js
```diff
@@ -764,6 +764,9 @@ class ClientController extends EventEmitter {
 	_createReactServerRootNode(mountNode, index) {
 		var root = document.createElement("div");
 		root.setAttribute(REACT_SERVER_DATA_ATTRIBUTE, index);
+		if (!root.getAttribute("key")) {
+			root.setAttribute("key", index);
+		}
```
gigabo: I think the guard here isn't necessary?  We've just created `root`, so it shouldn't already have a "key" attribute.

doug-wade:
Talked to @gigabo irl about this -- turns out this pr doesn't actually resolve the issue.  Ima switch the bike share example to use a `RootContainer` to get a better repro, then come at this again.

drewpc:
@doug-wade can we close this PR if it doesn't resolve the issue?

PR #616, opened 2016-08-23
Emit showMaster event when the back navigation is triggered in Frameb…

vinsewah:
…ackController


gigabo:
Nice.  Should there be a corresponding `hideMaster`?

gigabo:
Thanks @vinsewah!

PR #617, opened 2016-08-23
Updating the dependency on request-local-storage from 1.0.0 to 1.1.0

vinsewah:


gigabo:
This picks up the new [`isActive()` method](https://github.com/redfin/request-local-storage#what-is-error-rls-access-outside-of-request).

doug-wade:
🚢 

Mostly unrelated: I don't think our docs mention getCurrentRequestContext at all?

gigabo:
Yeah... I keep hoping we'll come up with something better. 😖

PR #625, opened 2016-08-25
Ensure generated .reactserverrc option keys match cli option keys

jbenesch:
Looks like the cli started using camelCasing for configuration options, whereas the generator generates a config file that uses dashes to separate words.

So when following the getting started instructions with the generator, and trying to change the js-port option, results in the port remaining 3001. Updating the option to jsPort, corrects the issue.

Added a unit test to future proof changes to the defaultOptions in the cli package and the generator.


packages/generator-react-server/test/app.js
```diff
@@ -24,6 +24,46 @@ test('generator-react-server:app creates default files', async t => {
...
+		// Using an if statement with pass/fail instead of t.true because the linter wanted:
+		// The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype; guard-for-in
+		if (defaultOptionsArray.indexOf(key) > -1) {
+			t.pass(`${key} exists in the cli default options`);
+		} else {
+			t.fail(`${key} does not exist in the cli default options. Please fix this.`);
+		}
+	}
+});
+
```
gigabo: Whoa.  This is sweet!

packages/generator-react-server/test/app.js
```diff
@@ -24,6 +24,46 @@ test('generator-react-server:app creates default files', async t => {
...
+	let testDir;
+	await helpers.run(path.join(__dirname, '../generators/app'))
+		.inTmpDir(dir => {
+			testDir = dir;
+		})
+		.withPrompts({name: 'foo', dockerCfg: false})
+		.toPromise();
+
+	// Reading files here instead of requiring because we don't understand export default here.
+	const defaultOptions = await readFile('defaultOptions.js', path.join(__dirname, '../../react-server-cli/src'));
```
gigabo: Seems easier to just `require(path.resolve(...))` here... what's the benefit of reading and parsing like this?

jbenesch: I completely agree but when I try and require it I get this error:

```
# generator-react-server:app .reactserverrc matches react-server-cli default option keys
not ok 1 - failed with "Unexpected token export"
  ---
    operator: undefined
    expected: undefined
    actual: undefined
    at: SyntaxError: Unexpected token export
  ...
```

Thoughts? There's gotta be a babel plugin different between react-server-cli and the generator but I couldn't quite figure it out and didn't really want to add a package for a unit test.

jbenesch: I get this error when I try and import the file at the top:

`import defaultOptions from '../../react-server-cli/src/defaultOptions.js';`

```
TAP version 13
/Users/jason/Projects/react-server/packages/react-server-cli/src/defaultOptions.js:14
export default {
^^^^^^
SyntaxError: Unexpected token export
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:513:28)
```

gigabo: Oh, I see.  Yeah, that's the source file, which hasn't gone through babel yet.

I think the nicest here would be to export `defaultOptions` from `react-server-cli`?

Then you could just do:

```javascript
import {defaultOptions} from "react-server-cli";
```

If you're up for that, the CLI exports are set up [here](https://github.com/redfin/react-server/blob/master/packages/react-server-cli/src/index.js#L20-L23).  Because the CLI has a build step that does babel transpilation you'll need to cd in there and `npm run prepublish` for the changes to become available.

doug-wade:
omg this is the best and you are the best and I <3 everything about this forever ✨ 🚀 🚢 :shipit:

PR #627, opened 2016-08-26
Upgrade superagent to 1.8.4

doug-wade:
Pull in bug fixes from superagent, most notably the fix to [250](https://github.com/visionmedia/superagent/issues/250), which was released in [1.5.0](https://github.com/visionmedia/superagent/releases/tag/v1.5.0). I could upgrade to 2.2.0, but I'll save the major version bump for another day.

### DIFF
--- a/packages/generator-react-server/generators/app/templates/package.json
+++ b/packages/generator-react-server/generators/app/templates/package.json
@@ -16,7 +16,7 @@
     "react-dom": "~0.14.2",
     "react-server": "^0.4.6",
     "react-server-cli": "^0.4.6",
-    "superagent": "1.2.0"
+    "superagent": "1.8.4"
   },
   "devDependencies": {
     "ava": "^0.15.1",
--- a/packages/react-server-cli/src/commands/init.js
+++ b/packages/react-server-cli/src/commands/init.js
@@ -10,7 +10,7 @@ const DEPENDENCIES = [
 	"babel-preset-react-server",
 
 	// TODO: Modernize our peer deps and remove versions here.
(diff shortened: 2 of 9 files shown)


gigabo:
Remind me why superagent is a peer dependency instead of a normal dependency?

roblg:
> Remind me why superagent is a peer dependency instead of a normal dependency?

I think because we wanted to make sure it's a singleton. Internally, we're installing custom response parsers. If we added the ability to install custom response parsers via ReactServerAgent (do we have that? We have response plugins..?) then it maybe wouldn't matter.

gigabo:
Yeah, I think we do everything through RSA... not dealing with superagent at all from outside?

Would be nice to eliminate the need for another explicit dep for node 6 users.

doug-wade:
@roblg we can use `plugRequest` and `plugResponse` for that, yeah? https://github.com/redfin/react-server/blob/master/docs/guides/using-react-server-agent.md#extending-superagent

gigabo:
Make https://github.com/redfin/react-server/issues/628.  Doesn't need to hold this up, I guess.

Though users will need to update their superagent deps when they upgrade to the next version of React Server...

doug-wade:
@gigabo we'll see how far we get today -- I also opened #630

roblg:
> we can use plugRequest and plugResponse for that, yeah?

I think `plugResponse` may actually be too late, but I found this: https://visionmedia.github.io/superagent/#parsing-response-bodies, which definitely wan't around when we were hacking things up initially, and should make our lives easier. :)

PR #630, opened 2016-08-26
Upgrade superagent harder

### LINKED ISSUE #628: Make superagent a normal dependency
Would be nice to eliminate a required dependency for node 6 users.

See discussion [here](https://github.com/redfin/react-server/pull/627).

doug-wade:
Fix #628


gigabo:
:metal: This is a headline improvement.

(If it works) 😁

doug-wade:
> If it works

current status: tests fail, but the examples work as anticipated with `react-server-cli` and `react-server` linked (even after pruning).

PR #631, opened 2016-08-26
Export default options so other dependent packages can test against changing properties

jbenesch:
See: https://github.com/redfin/react-server/pull/625

I also had questions around dependencies. When I tried to run gulp from react-server-cli package I was getting an error like so:

```
JASONs-Mac-Pro:react-server-cli jason$ npm run prepublish
path.js:7
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received null
    at assertPath (path.js:7:11)
```

What I realized was that because gulp wasn't a part of this package, it was trying to use a globally installed version I had which was out of date. I have always been under the belief that packages should force using local packages by using npm scripts. By adding gulp as a devDep, running npm run prepublish will always work.

Would love to hear your thoughts?

Thanks!

### DIFF
--- a/packages/react-server-cli/package.json
+++ b/packages/react-server-cli/package.json
@@ -48,6 +48,9 @@
   },
   "devDependencies": {
     "babel-preset-react-server": "^0.4.4",
+    "gulp": "^3.9.1",
+    "gulp-babel": "^6.1.2",
+    "gulp-eslint": "^3.0.1",
     "nsp": "^2.4.0",
     "react-server-gulp-module-tagger": "^0.4.5",
     "rimraf": "^2.5.2"
--- a/packages/react-server-cli/src/index.js
+++ b/packages/react-server-cli/src/index.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
 const run = require("./run").default;
 const parseCliArgs = require("./parseCliArgs").default;
+const defaultOptions = require("./defaultOptions").default;
 
 require.extensions['.css'] =
 require.extensions['.less'] =
--- a/packages/react-server-cli/src/index.js
+++ b/packages/react-server-cli/src/index.js
@@ -20,4 +20,5 @@ require.extensions['.md'] =
 module.exports = {
 	parseCliArgs,
 	run,
+	defaultOptions,
 };


packages/react-server-cli/package.json
```diff
@@ -48,6 +48,9 @@
   },
   "devDependencies": {
     "babel-preset-react-server": "^0.4.4",
+    "gulp": "^3.9.1",
+    "gulp-babel": "^6.1.2",
+    "gulp-eslint": "^3.0.1",
```
gigabo: Yeah, this is great.

We've gone back and forth here.  It's nice to factor out devDependencies into the repo root to cut down on bootstrap time and storage requirements.  But for modules where we use the "bin" script that breaks things.  In this case I think adding it back into the package is the right move.

Eventually I'd like to have a `lerna bootstrap --hoist-external` or something that will let us define deps where they're actually used but then install them at the root and link bin scripts in where needed.  Some day. :)

packages/react-server-cli/src/index.js
```diff
@@ -20,4 +20,5 @@ require.extensions['.md'] =
 module.exports = {
 	parseCliArgs,
 	run,
+	defaultOptions,
```
gigabo: I think you'll need to import this first?

Like:

```javascript
const defaultOptions = require("./defaultOptions").default;
```

Up at the top?

gigabo: Actually this should fail travis with a linter violation.

doug-wade:
Re: installing gulp as a dev dependency -- I dig it.  We shouldn't rely on globally-installed tools at all, and personally I'd like to see us go one step further and use https://www.npmjs.com/package/npm-run

packages/react-server-cli/src/index.js
gigabo: Right on cue. 👹 
image

packages/react-server-cli/package.json
jbenesch: Maybe I'm missing the correct dev entry point... but when I cloned and then ran `npm run bootstrap` I hit the same gulp error.

packages/react-server-cli/src/index.js
jbenesch: I was just testing the CI... Making sure it was still working correctly :)

packages/react-server-cli/package.json
gigabo: Yeah, we probably have other packages that will need similar treatment. 😬

gigabo:
Your original title was actually good.  We use [lerna-changelog](https://github.com/lerna/lerna-changelog) to generate our [release notes](https://github.com/redfin/react-server/releases).  It automatically groups changes by package.

image

gigabo:
Thanks @jbenesch!

PR #521, opened 2016-08-11
Remove the missing/unused command

dharamgollapudi:
Per the *Getting Started with React Server* steps at [https://react-server.io/](url) when I run `npm run compile`, getting the following error:

```
npm ERR! Darwin 15.6.0
npm ERR! argv "/usr/local/Cellar/node/6.3.1/bin/node" "/usr/local/bin/npm" "run" "compile"
npm ERR! node v6.3.1
npm ERR! npm  v3.10.6

npm ERR! missing script: compile
npm ERR!
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>
```

When I proceeded with the next command, `npm run start`, site seems to be up and running.

Looks like the step `npm run compile` is not needed. Documentation at [https://react-server.io/](url) needs to be updated.

PR #634, opened 2016-08-26
Update website docs with correct getting started path

jbenesch:
Small update, but it's not a great experience when you get an error walking through the getting started.


gigabo:
Man, [how](https://github.com/redfin/react-server/pull/521/files) [many](https://github.com/redfin/react-server/pull/522/files) of [these](https://github.com/redfin/react-server/pull/538/files) did we _have_?!

Thanks @jbenesch. 😛

PR #642, opened 2016-08-29
Adding http response header hook to page lifecyle

nampas:
This PR provides another lifecycle method to allow page chains to specify their own http headers.

Some things:
- Subsequent lifecycle methods that set http headers will override those that the user has specified (e.g. Transfer-Encoding in writeHeader() )
- This doesn't really work for client transitions, because in that case no page request is issued.

### DIFF
--- a/packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
+++ b/packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
@@ -1,0 +1,12 @@
+var Q = require("Q");
+
+class AsyncHttpHeadersPage {
+
+	getHeaders() {
+		return Q([
+			["Content-Security-Policy", "example.com"],
+		]);
+	}
+}
+
+module.exports = AsyncHttpHeadersPage;
--- a/packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
+++ b/packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
@@ -7,10 +7,6 @@ class AsyncHttpHeadersPage {
 			["Content-Security-Policy", "example.com"],
 		]);
 	}
-
-	getContentType() {
-		return "application/example";
-	}
 }
 
 module.exports = AsyncHttpHeadersPage;
--- a/packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
+++ b/packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
@@ -7,6 +7,10 @@ class AsyncHttpHeadersPage {
 			["Content-Security-Policy", "example.com"],
 		]);
 	}
+
+	getContentType() {
+		return Q("application/example");
+	}
 }
 
 module.exports = AsyncHttpHeadersPage;
--- a/packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
+++ b/packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
@@ -9,7 +9,7 @@ class AsyncHttpHeadersPage {
 	}
 
 	getContentType() {
-		return Q("application/example");
+		return "application/example";
 	}
 }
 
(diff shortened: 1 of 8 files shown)


packages/react-server-integration-tests/src/__tests__/httpHeaders/HttpHeadersSpec.js
```diff
@@ -0,0 +1,35 @@
...
+describe('A page with custom http headers', () => {
+
+	helper.startServerBeforeAll(__filename, [
+		'./HttpHeadersPage',
+		'./AsyncHttpHeadersPage',
+	]);
+
+	helper.stopServerAfterAll();
+
+	fdescribe('has sync rendered headers', () => {
```
decoy31: `fdescribe` -> `describe`

nampas: Ha, whoops

packages/react-server-integration-tests/src/__tests__/httpHeaders/AsyncHttpHeadersPage.js
```diff
@@ -0,0 +1,12 @@
+var Q = require("Q");
```
doug-wade: The module is named [q](https://www.npmjs.com/package/q)

doug-wade:
I'm going to have to think about this one for a while.  All of the lifecycle methods on pages so far have to do with the document that is returned by the request, and nothing to do with the http request/response lifecycle that underpins it.  While I understand and agree that we'll need to have some support for http headers, I wonder if this is the way to do it, or if we should expose some abstraction or api for the underlying request object to keep the page api small and coherent.

nampas:
> ll of the lifecycle methods on pages so far have to do with the document that is returned by the request, and nothing to do with the http request/response lifecycle that underpins it. 

That's not quite true, we already have `getContentType()`. Also several other methods in the render flow set their own headers, e.g. `writeHeader()`

packages/react-server/core/renderMiddleware.js
```diff
@@ -295,6 +296,13 @@ function setDataBundleContentType(req, res) {
 	res.set('Content-Type', 'application/json');
 }
 
+function setHeaders(req, res, context, start, pageObject) {
```
vinsewah: Nitty: Would probably rename this to setResponseHeaders to be extra explicit?

roblg: Yeah, I think `setResponseHeaders` or `setHttpHeaders` would be clearer, since we already have `writeHeader`, referring to `<head>`

roblg:
I think this changeset is in line with the overall philosophy of the page class: all things response-related are represented by methods on the Page. There are pros and cons to that philosophy, but it's mostly been OK so far.

I'm tempted to say we should also allow `getHeaders` in `rawResponseLifecycle`, since that's in line with `getContentType`.

nampas:
Because I'll be adding `setHttpHeaders()` to `rawResponseLifecycle`, I'll just roll `setContentType` into the function too.

nampas:
After consolidating `setContentType` and `setHttpHeaders`, my tests are failing. `setHttpHeaders` is correctly setting the content-type header, but then somehow the default implementation kicks in. Will revisit soon.

nampas:
So yeah, turns out that `getContentType()` does nothing in the page/fragment lifecycle, likely due to the `res.type('html');` call in `writeHeader()`.
I'm now preparing a final commit which will split out getContentType again from setHttpHeaders, and remove it from the page/fragment lifecycles. Also adding an integration test for `rawResponseLifecycle`

gigabo:
So, the `setContentType` call for the page and fragment lifecycles were basically noops?

Is express just giving us a `charset=utf-8` itself?  And [this one](https://github.com/redfin/react-server/blob/master/packages/react-server/core/util/PageUtil.js#L53) is just not used?

Just want to make sure the charset will still be set explicitly in the content-type header...

packages/react-server-integration-tests/src/__tests__/httpHeaders/HttpHeadersSpec.js
```diff
@@ -0,0 +1,34 @@
+var helper = require('../../specRuntime/testHelper');
```
gigabo: Thanks for adding these new tests!

nampas:
Yes, getContentType() on pages and fragments were basically noops.

See here for a test to support this claim:
https://github.com/nampas/react-server/pull/1
I also tested by setting content type on a Corvair page (Redfin's react-server implementation), and got the same result.

gigabo:
Awesome.  Thanks for verifying @nampas.

nampas:
No prob. Let me know if you want me to squash these commits down

gigabo:
No need.  We'll squash on merge. 🍆

PR #643, opened 2016-08-29
Fix broken link to lerna docs

alanbares:
Noticed the link to the lerna docs in the contributing file was broken so fixed the link.


gigabo:
Oh, whoops!  Thanks @alanbares!

Issue #569, opened 2016-08-16
Inconsistent return value from cli.start if `https === true`


gigabo:
Follow on to #568.

The other two cases return the output of `startServers`.

The `https === true` case doesn't return anything.

PR #644, opened 2016-08-29
Normalize httpsOptions from parseCli

jbenesch:
See: https://github.com/redfin/react-server/issues/569

Most checks in the codebase, check for httpsOptions and not the https flag. This PR ensures httpsOptions is always defined and set from parseCli.

- I also added a babel plugin called `add-module-exports` to make things easier when importing default exports.
- I added a package called ava for tests (following the generator package lead).
- I moved the pem package from the start command over to parseCli, so we can get a standard httpsOptions object (or boolean) in cli.start.
- Removed the camelize function from parseCli, as the yargs package already does that for us.
- Wrote 24 unit tests. It's still not complete and will work on adding more tests around all cli flags.
- Linter passes.

### DIFF
--- a/packages/react-server-cli/src/cli.js
+++ b/packages/react-server-cli/src/cli.js
@@ -1,7 +1,4 @@
 require("babel-core/register");
 
 const cli = require(".");
-
-cli.parseCliArgs().then(args => {
-	cli.run(args);
-});
+cli.parseCliArgs().then(cli.run);
--- a/packages/react-server-cli/src/cli.js
+++ b/packages/react-server-cli/src/cli.js
@@ -2,4 +2,6 @@ require("babel-core/register");
 
 const cli = require(".");
 
-cli.run(cli.parseCliArgs());
+cli.parseCliArgs().then(args => {
+	cli.run(args);
+});
--- a/packages/react-server-cli/src/index.js
+++ b/packages/react-server-cli/src/index.js
@@ -1,7 +1,7 @@
 const fs = require('fs');
-const run = require("./run").default;
-const parseCliArgs = require("./parseCliArgs").default;
-const defaultOptions = require("./defaultOptions").default;
+const run = require("./run");
+const parseCliArgs = require("./parseCliArgs");
+const defaultOptions = require("./defaultOptions");
 
 require.extensions['.css'] =
 require.extensions['.less'] =
--- a/packages/react-server-cli/src/parseCliArgs.js
+++ b/packages/react-server-cli/src/parseCliArgs.js
@@ -153,11 +153,11 @@ const sslize = async argv => {
 			};
 		} else {
 			argv.httpsOptions = await new Promise((resolve, reject) => {
-				pem.createCertificate({ days:1, selfSigned:true }, (err, keys) => {
+				pem.createCertificate({ days: 1, selfSigned: true }, (err, keys) => {
 					if (err) {
 						reject(err);
 					}
-					resolve({key: keys.serviceKey, cert:keys.certificate});
+					resolve({ key: keys.serviceKey, cert: keys.certificate });
 				});
 			});
 		}
--- a/packages/react-server-cli/src/parseCliArgs.js
+++ b/packages/react-server-cli/src/parseCliArgs.js
@@ -1,10 +1,13 @@
 import yargs from "yargs"
 import fs from "fs"
+import pem from "pem"
 
 export default (args = process.argv) => {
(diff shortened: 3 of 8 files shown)


packages/react-server-cli/src/index.js
```diff
@@ -1,7 +1,7 @@
 const fs = require('fs');
-const run = require("./run").default;
-const parseCliArgs = require("./parseCliArgs").default;
-const defaultOptions = require("./defaultOptions").default;
+const run = require("./run");
+const parseCliArgs = require("./parseCliArgs");
+const defaultOptions = require("./defaultOptions");
```
gigabo: Ooh, that's nice!  Didn't know about [`babel-plugin-add-module-exports`](https://www.npmjs.com/package/babel-plugin-add-module-exports).

packages/react-server-cli/src/parseCliArgs.js
```diff
@@ -162,18 +183,3 @@ const removeUndefinedValues = (input) => {
...
-	for (let key in Object.keys(inputCopy)) {
-		if (key.indexOf("-") !== -1) {
-			const newKey = key.replace(/-(.)/g, replaceFn);
-			inputCopy[newKey] = inputCopy[key];
-			delete inputCopy[key];
-		}
-	}
-
-	return inputCopy;
-}
```
gigabo: Wait, so how does `--js-url` get turned into `options.jsUrl` now?

packages/react-server-cli/src/run.js
```diff
@@ -33,7 +33,7 @@ export default function run(options = {}) {
 	options.outputUrl = jsUrl || `${httpsOptions ? "https" : "http"}://${host}:${jsPort}/`;
 
 	try {
-		return require("./" + path.join("commands", options.command)).default(options);
+		return require("./" + path.join("commands", options.command))(options);
```
gigabo: :metal:

packages/react-server-cli/src/parseCliArgs.js
```diff
@@ -124,27 +127,45 @@ export default (args = process.argv) => {
...
+		if ((httpsKey && httpsCert) || httpsPfx) {
+			argv.httpsOptions = {
+				key: httpsKey ? fs.readFileSync(httpsKey) : undefined,
+				cert: httpsCert ? fs.readFileSync(httpsCert) : undefined,
+				ca: httpsCa ? fs.readFileSync(httpsCa) : undefined,
+				pfx: httpsPfx ? fs.readFileSync(httpsPfx) : undefined,
+				passphrase: httpsPassphrase,
+			};
+		} else {
+			argv.httpsOptions = await new Promise((resolve, reject) => {
```
gigabo: > `... await new Promise(...)`

Mind. Blown.

packages/react-server-cli/src/cli.js
```diff
@@ -1,5 +1,4 @@
 require("babel-core/register");
 
 const cli = require(".");
-
-cli.run(cli.parseCliArgs());
+cli.parseCliArgs().then(cli.run);
```
gigabo: Okay, so let me make sure I understand this: Because `sslize` is an _async_ function now it implicitly returns a promise, which we chain off of here?

packages/react-server-cli/test/parseCliArgs.js
```diff
@@ -0,0 +1,326 @@
+import test from 'ava';
```
gigabo: Oh, man this is so great!  Thanks @jbenesch for working this out and locking it down with these tests!

packages/react-server-cli/src/parseCliArgs.js
gigabo: Oh, just read your summary. 😅 

> the yargs package already does that for us.

Nice.

jbenesch:
```
if (/-/.test(key) && !(flags.aliases[key] && flags.aliases[key].length) && configuration['camel-case-expansion']) {
      var c = camelCase(key)
      flags.aliases[key] = [c]
      newAliases[c] = true
    }
```
https://github.com/yargs/yargs-parser/blob/master/index.js#L336-L340

🐫  🎸  :metal:

packages/react-server-cli/src/cli.js
jbenesch: 👍 

```an Async Function always returns a Promise.``` ([ref](https://ponyfoo.com/articles/understanding-javascript-async-await))

PR #645, opened 2016-08-30
Updated docs. Removed repeating paragraph.

hzhu:
Updated docs. Removed a repeating paragraph that can be found earlier from line 94.


gigabo:
Oops, nice catch @hzhu!

PR #661, opened 2016-09-16
lerna => asini

gigabo:
Using asini to manage the repo now.


roblg:
Seems reasonable to me. I was able to pull the branch, run `nuke`, `bootstrap` and start `react-server-test-pages` with npm3, so it can't be _that_ broken, amirite?

PR #676, opened 2016-09-27
Clean corvair comments

sresant:
These comments were originally intended for internal usage, removing now that we have gone open source.


gigabo:
Thanks @sresant.

PR #677, opened 2016-09-28
Prevent "when" childProps from clobbering "listen" chlidProps in RootElement

TonyHYK:


packages/react-server/core/components/RootElement.js
```diff
@@ -175,7 +175,9 @@ RootElement.scheduleRender = function(element) {
 		.then(el => el && (rendered = el))
 		.then(() => when)
 		.then(childProps => childProps
-			?React.cloneElement(rendered, {childProps})
+			// merge "when" childProps and "listen" childProps preemptively
+			// to prevent cloneElement shallow merge from clobbering "listen" childProps
+			?React.cloneElement(rendered, {childProps: _.assign({}, rendered.props.childProps || {}, childProps)})
```
TonyHYK: it's a bit awkward that I have to "reach in" to grab the existing childProps

gigabo: Yeah... no real way around it, though.

gigabo:
Nice fix @TonyHYK.

PR #691, opened 2016-10-03
Fix inconsistent usage of routes.js vs. routes.json

sresant:
This pr addresses 2 issues
1. The init command was failing due to missing superagent dep
2. start command was not running correctly because the default is routes.json not routes.js


doug-wade:
I don't think either of these changes are good.  We removed superagent as a peer dependency and replaced it with a regular dependency in [0.4.7](https://github.com/redfin/react-server/releases/tag/v0.4.7), so client code shouldn't declare a dependency on superagent, lest there are duplicate superagents.  Also, [all of our examples](https://github.com/redfin/react-server/tree/master/packages/react-server-examples) use routes.js instead of routes.json, so we should use routes.js rather than routes.json.

sresant:
@doug-wade - yeah i undid the superagent changes. However the routes change will be needed but couldn't decide on whether js should be the default or json. The current problem is that if you create a new react-server project with our current HEAD start does not work because we set the default to routes.js but init generates routes.json. The other solution would be to change to js ofc.

gigabo:
If it's just a matter of internal consistency I'd prefer json over js, since that's easy to manipulate with the CLI (`react-server add-page ...`).  A js routes file will still _work_, but I'd rather not make it the recommended first choice.  Generally I'd say json is better until/unless you _need_ js for some reason.

gigabo:
@sresant What was the resolution on the superagent errors you were seeing?

sresant:
@gigabo yea not seeing that superagent error anymore 😕 so i just took those changes out.

gigabo:
> all of our examples use routes.js instead of routes.json

This is a good point.  Regardless of which we default to it would be good to add some documentation explaining that they're both supported and the pros/cons of each.

Added https://github.com/redfin/react-server/issues/693 to track that.

gigabo:
@doug-wade Does @sresant's explanation clear things up for you here?

This fixes a pretty bad bug with the CLI, so I'd like to get it into the next release if possible.

doug-wade:
:+1:

PR #696, opened 2016-10-06
Adding Asini Slackin service to the example website

davidalber:
- Doing a bunch of renaming of the form "slackin" --> "react-server-slackin" to make way for another Slackin service.
- Adding a second Slackin service for the Asini Slack team.


gigabo:
Thanks for dealing with this @davidalber.

Issue #170, opened 2016-05-18
Don't log client performance times without `window.performance`


gigabo:
Born from https://github.com/redfin/react-server/pull/168#discussion_r63759965.

gigabo:
@sresant Did #699 fix this?

sresant:
@gigabo Yes it should have, closing.

PR #699, opened 2016-10-07
Improve logging inside of ClientController

sresant:
Relates to #170


gigabo:
Nice, thanks @sresant!

Issue #694, opened 2016-10-05
Provide a hook to insert custom express middleware


gigabo:
The server set up by the CLI currently doesn't provide any way to insert custom _express_ middleware.

For example:

``` javascript
app.use(session(options))
```

mlolson:
Question I was pondering: Would it be feasible to re-factor react server so that it could be used as middleware in an express application? Like:

    ReactServer = require('react-server');   
    //pagesDir tells react server where to find the routes file and pages...
    app.use(ReactServer({pagesDir: 'pages'});

In my experience it's easier to adopt something when you can use it within an existing framework like express, rather than having to buy the whole enchilada.

Does that make any sense? :)

stanonyime:
My thoughts exactly

gigabo:
Hi @mlolson that's actually possible already!  The CLI wraps that up for you, but at its core all it's doing is creating an express instance and [inserting middleware](https://github.com/redfin/react-server/blob/master/packages/react-server-cli/src/commands/start.js#L76) for you.  It looks a little funny, because it's inverted so the `app.use` happens inside of a method call, but this method call can go wherever you'd ordinarily want to put `app.use(ReactServer({routesFile: './routes.js'});`.

```javascript
app.use(ReactServer({routesFile: './routes.js'});
```

mlolson:
Interesting! PS Is the README up to date? I tried to use the generator to get started but it seemed to fail.

sresant:
@mlolson Hey! Sorry to hear you are having issues can you outline what the error is and the version of npm you are using? I recently filed a bug related to npm 6.2.2 #698.

mlolson:
@sresant After I follow the instructions in the README, when I run `npm run start` it produces this error:

    Error: Cannot find module 'express'....

My directory doesn't look right. The app that is installed does not look like a react server app at all... Very odd

doug-wade:
@mlolson almost certainly you've installed generator-react-server v1.0.0, which is an unrelated project.  We got the name from the very gracious @StevenIseki, but there are still some versions of of his previous project with that name on npm.  Make sure you have v0.4.13 installed before you scaffold out a new project.

mlolson:
Lol! Thanks

drewpc:
This has already been merged.  Closing.

PR #708, opened 2016-10-11
Hook for custom middleware

sresant:
Relates to #694.

### DIFF
--- a/docs/guides/react-server-cli.md
+++ b/docs/guides/react-server-cli.md
@@ -115,7 +115,7 @@ export default (webpackConfig) => {
 In the `.reactserverrc` file add an option for `webpack-config` that points to that function file and when React Server is setting up Webpack it will call your function with the result of the built in Webpack options, allowing you to make any modifications needed.
 
 ### Use Custom Express Middleware
-React Server currently the default Express Middlewares used are compression, body-parser, cookie-parser and of course react-server. If you need to setup custom express middleware you can do it with a setup function.
+Currently the default Express Middlewares used are compression, body-parser, cookie-parser. If you need to setup custom express middleware you can do it with a setup function.
 
 ```javascript
 export default (server, reactServerMiddleware) => {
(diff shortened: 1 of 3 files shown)


docs/guides/react-server-cli.md
```diff
@@ -114,9 +114,22 @@ export default (webpackConfig) => {
 
 In the `.reactserverrc` file add an option for `webpack-config` that points to that function file and when React Server is setting up Webpack it will call your function with the result of the built in Webpack options, allowing you to make any modifications needed.
 
+### Use Custom Express Middleware
```
sresant: @gigabo i am the worst at documentation so let me know what you think

gigabo:
Hi @doug-wade!  It looks like @sresant has actually accounted for that.  The _default_ behavior remains to apply the various other middlewares (compression, etc) in a certain order and then apply the React Server middleware.  But if the new _hook_ is used the caller becomes responsible for applying _all_ middleware, _including_ compression, etc _and_ React Server's, in whatever order is desired.

docs/guides/react-server-cli.md
```diff
@@ -114,9 +114,22 @@ export default (webpackConfig) => {
...
+	server.use(bodyParser.urlencoded({ extended: false }));
+	server.use(bodyParser.json());
+	server.use(session(options));
+	rsMiddleware(); // Must be called once or server will not start
+}
+```
 
 
+In the `.reactserverrc` file add an option for `custom-middleware-path` that points to that function file and when React Server is setting up the server it will call your function for setup rather then the default middlewares mentioned above. On the command line the option is `--custom-middleware-path`.
+
```
gigabo: I think we've generally used the camelCase variants in the config file?  Like `customMiddlewarePath`?

Not sure if either will work, or even if I'm remembering right. ;)

sresant: You are correct, i was just tired when writing up the documentation good catch!

docs/guides/react-server-cli.md
```diff
@@ -114,9 +114,22 @@ export default (webpackConfig) => {
...
+### Use Custom Express Middleware
+Currently the default Express Middlewares used are compression, body-parser, cookie-parser. If you need to setup custom express middleware you can do it with a setup function.
 
+```javascript
+export default (server, reactServerMiddleware) => {
+	server.use(compression());
+	server.use(bodyParser.urlencoded({ extended: false }));
+	server.use(bodyParser.json());
+	server.use(session(options));
+	rsMiddleware(); // Must be called once or server will not start
```
sresant: Good catch!

PR #709, opened 2016-10-11
Add cli option  for webpack-config

sresant:
Currently do not have a command line option to pass a path to our webpack-config. This creates that option.

### DIFF
--- a/docs/guides/react-server-cli.md
+++ b/docs/guides/react-server-cli.md
@@ -140,7 +140,7 @@ export default (webpackConfig) => {
 }
 ```
 
-In the `.reactserverrc` file add an option for `webpackConfig` that points to that function file and when React Server is setting up Webpack it will call your function with the result of the built in Webpack options, allowing you to make any modifications needed. On the command use the option `--webpack-config`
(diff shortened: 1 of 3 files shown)


docs/guides/react-server-cli.md
```diff
@@ -112,7 +112,7 @@ export default (webpackConfig) => {
 }
 ```
 
-In the `.reactserverrc` file add an option for `webpack-config` that points to that function file and when React Server is setting up Webpack it will call your function with the result of the built in Webpack options, allowing you to make any modifications needed.
+In the `.reactserverrc` file add an option for `webpackConfig` that points to that function file and when React Server is setting up Webpack it will call your function with the result of the built in Webpack options, allowing you to make any modifications needed. On the command line use the option `--webpack-config`
```
gigabo: I might use wording that makes it clear that this is an _alternative_ to the option in the config file.  As it stands it's possible to misinterpret this to mean that even with the option in the config file you must also pass a CLI option to trigger its use.

Maybe something like:

> This may also be specified on the command line with the `--webpack-config=<FILE>` option.

doug-wade:
@sresant I didn't merge this immediately in case @gigabo wanted to take a look; seems like he doesn't.  Rebase this onto master and lets get it merged.

gigabo:
@sresant Please rebase this when you get a chance.

Issue #711, opened 2016-10-11
ClientController has weird behavior with slow Endpoints


sresant:
We were seeing an issue with a page that had a slow endpoint. Specifically when the endpoint timed out the server failed to write everything to the document and would continue to try after the stream had been closed. Additionally when ClientController woke up it blew away the document and the elements on the documents. Getting a repro in test-pages would be a good first step.

PR #716, opened 2016-10-14
Fix timeout error handling

sresant:
Relates to issue #711 

The main issues addressed here are:
1. When renderMiddleware timesout, it will send down what ever it has left but it fails to close out the body and html.
2. In ClientController there is a race condition between nodeArrival and failArrival. nodeArrival will receive a bunch of nodes since before failArrival is sent the nodes we have left are sent down. If failArrival runs an resolves `retVal` before the nodes can actually render the `_previouslyRender` flag will be set. Once that is set and the elements try to render, because `_previouslyRender` is set at the point `_cleanup` will be called which will wipe away everything on the page.

### DIFF
--- a/packages/react-server/core/renderMiddleware.js
+++ b/packages/react-server/core/renderMiddleware.js
@@ -736,6 +736,8 @@ function writeBody(req, res, context, start, page) {
 
 		// Let the client know it's not getting any more data.
 		renderScriptsAsync([{ text: `__reactServerClientController.failArrival()` }], res)
+		// Close off the body since the page life cycle will not complete
+		res.write("</div></body></html>");
 	});
 
 	Q.all(dfds.map(dfd => dfd.promise)).then(retval.resolve);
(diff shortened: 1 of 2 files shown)


packages/react-server/core/renderMiddleware.js
```diff
@@ -736,6 +736,8 @@ function writeBody(req, res, context, start, page) {
 
 		// Let the client know it's not getting any more data.
 		renderScriptsAsync([{ text: `__reactServerClientController.failArrival()` }], res)
+		// Close off the body since the page life cycle will not complete
+		res.write("</div></body></html>");
```
gigabo: I don't think this is enough.  It's addressing the _symptom_ (we're not sending the close tags) rather than the _cause_ (we're short-circuiting out of the normal control flow, which includes sending the close tags).

Can we handle individual element render failures without rejecting the `writeBody` return promise?

gigabo:
This looks good to me now.  Nice work @sresant!

Issue #600, opened 2016-08-22
How to implement Redux?


incubus8:
Hi guys, any idea how to implement Redux? where can I find example Redux + React Server?

hzhu:
Here's a basic example, a reimplementation of the Redux counter app in react-server:
https://github.com/hzhu/react-server-redux-basic

Now how you structure a react-server, Redux application is up to you.

gigabo:
Whoa! @hzhu that's awesome!

We should link to that example from our docs.  Leaving this open as a reminder.

doug-wade:
@hzhu @gigabo what about pulling it into `react-server-examples`?

hzhu:
Yeah I was planning on that! How about I submit a PR for review in a day or two?

gigabo:
Even better!  Thanks @hzhu!

incubus8:
@hzhu thanks a lot!!

hzhu:
@incubus8 I realized that the Counter example isn't too useful once one introduces data fetching & user interaction. Here's a better use case with react-server & redux. It's still a work in progress, but seems to work. https://github.com/hzhu/react-server-redux

It's a bit messy, but once I learn how to cleanly setup a redux project with react-server, I'll make the PR to react-server-examples

woomike:
@hzhu Any chance you've had any luck using the RootElement's 'where' with Redux?  I'd like to render a component only when an async data call has completed, but 'where' requires a Promise, and my dispatch will no longer work if I return a promise instead of a function.

https://react-server.io/docs/guides/understanding-rendering

hzhu:
@woomike, did you mean RootElement's `when` attribute?

You can try representing your async data fetching as a promise, and put that, as the value to `when`.
For example..

```
const myPromise = new Promise((resolve, reject) => {
  // data fetch
  // async stuff here
})

<RootElement when={myPromise} />
```

Or more likely you might want to use `ReactServerAgent.get()` to fetch data from an api.
ReactServerAgent is a promise that represents the eventual result of the fetch.

I have a [working demo](https://github.com/hzhu/react-server-redux) of rendering async data using redux and react-server. The specific line you might want to check out is https://github.com/hzhu/react-server-redux/blob/master/src/pages/explore/index.js#L35

Note: I'm finding it very difficult to structure a redux app properly in react-server, i.e. how to manage store/stores & reducers. Let me know if you have some tips :)

woomike:
[...]

PR #722, opened 2016-10-20
Improve Redux-Basic example

mocheng:
Related to #600 , Original Redux-Basic example doesn't have state initialization.

This change demonstrates ways to initialize Redux state with promises.

1) `CounterAPI` responds with initial count value;
2) In `CounterPage`, `CounterAPI` response is dispatched to Redux store and triggers `Counter` rendering through `when`.

### DIFF
--- a/packages/react-server-examples/redux-basic/components/Counter.js
+++ b/packages/react-server-examples/redux-basic/components/Counter.js
@@ -1,11 +1,20 @@
 import React, { Component, PropTypes } from 'react'
 import { connect } from 'react-redux'
+import { ReactServerAgent } from 'react-server'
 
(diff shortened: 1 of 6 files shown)


gigabo:
Hi @mocheng looks like travis is failing during npm install here.  I think that's due to an unrelated issue that has since been resolved in master.  If you get a chance please rebase so we can get a clean test run. Thanks!

drewpc:
@mocheng Thanks for the contribution!  This PR has been sitting out there for a while and still needs to be fixed so that the tests passed.  I'm closing it until you can fix the test issues.  Feel free to reopen after that!

PR #732, opened 2016-10-26
Use Asini's new autmatic hoisting of external dependencies

gigabo:
- Speeds up bootstrap: ~2:30 vs ~4:20 on my laptop.
- Puts dependencies back where they're used.
- Cleans up the top-level package.json

Hoisting was introduce in [`asini@1.1.0`](https://github.com/asini/asini/releases/tag/v1.1.0).


gigabo:
Oh, sweet.  The travis build was about twice as fast as recent master builds.

image

gigabo:
@doug-wade This will render GreenKeeper useless, unfortunately.  The only dependency remaining at the root is Asini, and that can't be upgraded without re-running `asini init`.

... maybe Asini should only require re-running `asini init` on breaking changes. :thinking:

PR #739, opened 2016-11-02
Get test cases synchronously for gulp-module-tagger

gigabo:
Travis is failing on master with:

```
Errored while running npm script 'test' in 'react-server-gulp-module-tagger'
Error: Command failed: npm run test
  ✔ module can be required (147ms)
  Unhandled Rejection: test/test.js
    Error: All tests and hooks must be declared synchronously in your test
	file, and cannot be nested within other tests or hooks.
```

This is an attempt to avoid that by synchronously declaring tests.

Not able to repro locally.


gigabo:
That did the trick. 👍

PR #740, opened 2016-11-03
Add superagent to example package.json and print errors in cli.js

### LINKED ISSUE #698: react-server start fails on node 6.2.2
Create a new dir and follow the steps [here](https://react-server.io/docs/guides/react-server-cli) using react-server-cli. Start does not start and has no output. When I use node 4.3.1 this is not the case.

emecell:
Fixes #698

### DIFF
--- a/packages/react-server-cli/src/commands/init.js
+++ b/packages/react-server-cli/src/commands/init.js
@@ -12,6 +12,7 @@ const DEPENDENCIES = [
 	// TODO: Modernize our peer deps and remove versions here.
 	"react@~0.14.2",
 	"react-dom@~0.14.2",
+	"superagent@1.8.4",
 ]
 
 const CONFIG = {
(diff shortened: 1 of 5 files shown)


packages/react-server-cli/src/commands/init.js
```diff
@@ -12,6 +12,7 @@ const DEPENDENCIES = [
 	// TODO: Modernize our peer deps and remove versions here.
 	"react@~0.14.2",
 	"react-dom@~0.14.2",
+	"superagent@~1.8.4",
```
sresant: @gigabo Opinion on the tilda here? or do we want to be strictly on 1.8.4?

gigabo: Yeah, probably safer to match up exactly with `react-server`.

gigabo:
Thanks @emecell!

PR #748, opened 2016-11-11
Add user timings for WebPageTest

gigabo:
Use the browser's `performance.mark` API.

Add two timings:

- `react-server.wake`
    - When the ClientController is instantiated.
- `react-server.renderAboveTheFold`
    - When the content above the fold has been re-rendered (visible page is interactive)
- `react-server.renderComplete`
    - When the whole page has been re-rendered (fully interactive)


gigabo:
Updated to eliminate the change in above-the-fold timing logic.  I added a `renderComplete` performance mark, so the `renderAboveTheFold` mark isn't useful unless `<TheFold/>` is actually present.

PR #750, opened 2016-11-12
Add new example, meteor-site, to react-server examples

emecell:
This is from the talk given at Redfin on 11/9.


.eslintrc
```diff
@@ -43,7 +43,8 @@
     "handle-callback-err": 0,
     "indent": [
       2,
-      "tab"
+      "tab",
+      { SwitchCase: 1 }
```
emecell: I could drop this but otherwise it requires switch statements to be formatted like:
```
switch(a) {
case "blah":
default:
    return true;
}
```

packages/react-server-examples/meteor-site/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "react": "0.14.8",
+    "react-dom": "0.14.8",
+    "react-redux": "4.4.5",
+    "react-server": "^0.4.8",
+    "redux": "3.6.0",
+    "redux-thunk": "2.1.0",
+    "superagent": "1.8.4"
+  },
+  "devDependencies": {
+    "babel-preset-react-server": "0.4.4"
```
doug-wade: You're missing `react-server-cli` as a dev dependency, so `npm start` horks if you don't have react-server-cli installed globally (which is nice if you have multiple react-server projects using different versions of the cli).

emecell:
@doug-wade I think you might have an older version of Chrome: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Browser_compatibility

I just switched it to use lodash's find instead though.

I updated the dependencies to include the `react-server-cli`.

doug-wade:
Got it up and running locally -- it's pretty slick.  Thanks @emecell

PR #753, opened 2016-11-14
Removing unused regular deps: mkdirp, bundle-loader

roblg:
`bundle-loader` and `mkdirp` were never actual runtime dependencies of `react-server`. It looks like us including `mkdirp` as a runtime dependency probably allowed `react-server-integration-tests` to work though. :grin:


gigabo:
Ooh, nice find!  Progress toward #277!

PR #755, opened 2016-11-15
Update generator dependencies

doug-wade:
Get generator-react-server up on the most recent versions of dependencies

### DIFF
--- a/packages/generator-react-server/generators/app/templates/package.json
+++ b/packages/generator-react-server/generators/app/templates/package.json
@@ -10,7 +10,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "babel-preset-react-server": "^0.3.3",
+    "babel-preset-react-server": "^0.4.10",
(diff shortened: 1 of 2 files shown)


packages/generator-react-server/generators/app/templates/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "babel-preset-react-server": "^0.3.3",
+    "babel-preset-react-server": "^0.4.10",
```
gigabo: Oh, yikes.  Totally forgot about this thing... I wonder if we can use Asini's configurable package locations to get this updating automatically. :thinking:

gigabo: Captured: https://github.com/redfin/react-server/issues/758

gigabo:
Thanks @doug-wade!

PR #756, opened 2016-11-15
Fix #693

### LINKED ISSUE #693: Documentation on routes.js vs routes.json
It's not obvious why you'd want to use one vs the other (or even that they're both supported).

doug-wade:
#693 Document routes.js vs routes.json

### DIFF
--- a/docs/guides/react-server-cli.md
+++ b/docs/guides/react-server-cli.md
@@ -57,6 +57,34 @@ By default this is created at the site's root directory as `routes.json`.
 }
 ```
 
+You can also provide routes as a `routes.js` file.  The following routes.js
+file loads all middleware and pages from a `pages` and `middleware` folder
+and registers them.
+
+```javascript
+const fs = require('fs');
+const path = require('path');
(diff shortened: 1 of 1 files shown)


docs/guides/react-server-cli.md
```diff
@@ -57,6 +57,35 @@ By default this is created at the site's root directory as `routes.json`.
 }
 ```
 
+You can also provide routes as a `routes.js` file.  The following routes.js
+file is equivalent to the json file above, but will work on Windows, in addition
+to OS X and Linux.
```
doug-wade: Iirc, its not that `routes.json` doesn't work on Windows, its that windows uses / where everything else uses \ to delimit file paths fail, so while using a routes.json you have to provide Windows-compliant file paths or POSIX-compliant file paths, and that path.join smooths out those differences. It has been 3 or 4 months since we were dealing with the Windows failures, though, so it may have been something else that was horked -- I can pull the line if we wanna roll the dice.

Issue #761, opened 2016-11-16
Cannot bootstrap project because of v8-debug failure


drewpc:
Running `asini bootstrap` fails because of the `v8-debug` package fails to compile using node v7.1.0.  This is caused by the fact that `v8-debug` hasn't kept pace with the latest Google V8 debug API and node.js continues importing recent releases of Google V8.

More detail on the cause can be found here: https://github.com/node-inspector/v8-debug/issues/7#issuecomment-261074685

A PR is coming to remove the dependency causing this issue.

PR #762, opened 2016-11-16
Patch fix bootstrap using modern nodejs

drewpc:
Fixes bootstrap process identified in #761.  Also fixes a minor issue when running tests in a folder structure with invalid shell characters in the folder names (spaces, single quotes, etc).

### DIFF
--- a/packages/generator-react-server/package.json
+++ b/packages/generator-react-server/package.json
@@ -37,7 +37,8 @@
     "yeoman-test": "^1.0.0",
     "babel-eslint": "^7.0.0",
     "eslint": "^3.8.1",
-    "eslint-plugin-react": "^6.4.1"
+    "eslint-plugin-react": "^6.4.1",
+	"shell-escape": "^0.2.0"
   },
   "repository": "redfin/react-server",
   "scripts": {
--- a/packages/generator-react-server/package.json
+++ b/packages/generator-react-server/package.json
@@ -38,7 +38,7 @@
     "babel-eslint": "^7.0.0",
     "eslint": "^3.8.1",
     "eslint-plugin-react": "^6.4.1",
-	"shell-escape": "^0.2.0"
+    "shell-escape": "^0.2.0"
   },
   "repository": "redfin/react-server",
   "scripts": {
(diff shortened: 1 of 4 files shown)


packages/react-server-test-pages/package.json
```diff
@@ -8,7 +8,6 @@
     "prepublish": "gulp build",
     "start": "react-server start",
     "test": "gulp test",
-    "debug": "node-debug --debug-brk=0 -p 9000 react-server start",
```
doug-wade: I'm pretty sure we want to keep the debug target for test pages -- it's purpose is a place to repro and debug issues.

packages/generator-react-server/package.json
```diff
@@ -37,7 +37,8 @@
     "yeoman-test": "^1.0.0",
     "babel-eslint": "^7.0.0",
     "eslint": "^3.8.1",
-    "eslint-plugin-react": "^6.4.1"
+    "eslint-plugin-react": "^6.4.1",
+	"shell-escape": "^0.2.0"
```
doug-wade: Looks like the indentation is off here

packages/generator-react-server/test/app.js
```diff
@@ -129,7 +130,7 @@ function installDeps() {
 			} else {
 				const localDeps = ['react-server-cli', 'react-server']
 					.map(dep => path.resolve(path.join(__dirname, '../..', dep)));
-				cp.exec('npm install ' + localDeps.join(' '), (error) => {
+				cp.exec(shellescape(['npm', 'install'].concat(localDeps)), (error) => {
```
doug-wade: Out of curiosity, is there a deps that needs escaping, or is this defensive in case we add one? I think the change is good either way

drewpc: when the current path has spaces in it, it screws up the `cp.exec()` call.  Example:

```cp.exec('npm install /my path with spaces/react-server-cli')```

What should happen instead is something like this:

```cp.exec('npm install /my\ path\ with\ spaces/react-server-cli')```

Same thing would happen if there were a `'` in the path...it would mess up the JavaScript string.

gigabo:
Looks like the indentation is cleaned up (thanks for catching that @doug-wade).

I've created https://github.com/redfin/react-server/issues/765 to capture the need to re-introduce a "debug" script to test pages.

Gonna get this merged to unblock folks on node 7.  Thanks @drewpc!

PR #764, opened 2016-11-17
Add additional fields to the Request objects

drewpc:
Request objects available through `this.getRequest()` in middleware and elsewhere in the codebase provide access to the path and query portions of a URI (`scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]`).  This PR incorporates changes to the client and server side request wrappers to return scheme, host, and "secure" (matches the ExpressJS request object) values.  This does not address port, user, password, or fragment at this time.  The client implementation uses `window.location` if available and falls back to returning null.  The server implementation uses the ExpressJS request object.

### DIFF
--- a/packages/react-server/core/ClientRequest.js
+++ b/packages/react-server/core/ClientRequest.js
@@ -63,6 +63,28 @@ class ClientRequest {
 		return decode(this._url.substring(indexOfQuestion + 1));
 	}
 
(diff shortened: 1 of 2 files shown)


gigabo:
Nice!  This is a really useful improvement.  Thanks @drewpc!

Issue #264, opened 2016-06-03
Clean up TODOs


doug-wade:
We have [14 outstanding TODOs](https://github.com/redfin/react-server/search?p=1&q=TODO&utf8=%E2%9C%93).  We should audit and remove them -- filing them as issues if they haven't been fixed yet.

gigabo:
Ooh, good one.

doug-wade:
I borrowed the idea from [nodejs](https://github.com/nodejs/node/issues/4641) 😀

GeoDoo:
Hi, how are you? I am new to open source contribution and I would like to help out here!

Is this ok with you? I suppose I should get started here, right?

https://github.com/redfin/react-server/blob/master/CONTRIBUTING.md

joncancode:
@doug-wade i'd love to help out on here if any outstanding tasks are still around.

andela-iamao:
@doug-wade If there are still outstanding TODOs I would love to contribute

jmsjtu:
@doug-wade I'd also love to help out on anything outstanding and would love to contribute!

VolantTyler:
Are there TODOs still pending here? I would like to help (first-time contributor)

dhuang612:
I'd like to help out with any remaining to dos

EllieLockhart:
Are there any remaining TODOs? If so I'd like to help out.

PR #767, opened 2016-11-22
Removed react server agent util

rodrijuarez:
I’ve removed this util because it only had one function that could be replaced by the merge function from lodash. This is related to the issue #264. 
Something to have in mind is that previously we were replacing the properties from the destination even if one of the key properties from the destination was undefined, now merge from lodash will ignore the properties from the source object that are undefined ([link to documentation](https://lodash.com/docs/#merge))


doug-wade:
Awesome! Thanks @rodrijuarez !

Issue #763, opened 2016-11-16
line 156 in RootElement.js "var subscribe" function def not hoisted


jjunkin:
In file react-server/packages/react-server/core/components/RootElement.js

on line 156: var subscribe = callback => {updater = callback};

"var subscribe" is declared and the variable is hoisted. Yet as far as I can tell the function definition is not hoisted. Therefore, any usage above the definition the variable "subscribe" will be undefined, even though it is declared.

The only place I see "subscribe" is used is line 151 which is above this line, therefore I believe the value it passes will always be "undefined".

It appears that "this.props.subscribe" is evaluated on line 14 and if it has a value it is used. As far as I can tell this is the same value and it is always undefined. Therefore, the "this.props.subscribe " on line 15 is never executed.

If I am wrong about this please let me know. It seems like a fairly critical issue if I am correct.

A possible solution would be to move the function expression definition of 
156: var subscribe = callback => {updater = callback};
above the definition of 
137: var unsubscribe = listen(childProps => {

aickin:
I think that you're right that this isn't ideal, but practically it shouldn't cause a bug because the `subscribe` reference on line 151 is in a callback, so it happens after `subscribe` has been assigned. But you're right, the code would be cleaner with the definition above, and it'd be a great first PR to help the project. Would you be willing to submit it?

jjunkin:
When I stepped through the code with the chrome debugger, subscribe was always undefined - including within the callback function. I realize this could be different on various browsers (using chrome on osx).

From my understanding the "listen={emitter}" functionality (documented in Understanding Rendering) depends on this subscribe function being defined, therefore does not work as documented. Using the "listen=" prop I am able to get data to load in the initial render but I am unable to trigger a data reload after that.

Do you have any examples of the "listen" prop working  after the initial data load? I was unable to find any in the codebase. I noticed the test PropsFromEmitterRootElement.js but this only tests the initial data load using a single setTimeout. Examples of this functionality seem critical for the project and would be helpful.

I would be happy to try and submit a change if I get the time next week. Thanks for the response!

gigabo:
Oh, it actually looks like #769 addresses this. 👀

aickin:
@jjunkin: Thanks for the info and for tracing through the code! I stand corrected.

[...]

PR #769, opened 2016-11-23
Code legibility improvement

rodrijuarez:
Moved the declaration from the variable subscribe above the declaration
of the var unsubscribe, this is related to the issue #763

PR #770, opened 2016-11-23
Implement the no-use-before-define ESLint rule

aickin:
Turned on the `no-use-before-define` rule, in reference to bug #763, and changed the files that didn't pass the new listing rule.

I first turned the rule on completely, but `renderMiddleware` got really hairy, so I went back and used the `{functions: false}` parameter to the rule. This means that basic functions *don't* need to be defined in order, which is OK from a functionality perspective because functions are always hoisted. `renderMiddleware` is a bit of function soup, so setting this parameter let me sidestep it without compromising on correctness. If someone else wants to tackle implementing the rule with `{functions:true}`, that's cool, but it's also probably fine not to do so, as well.

I just moved functions or variable definitions around in the files without changing their contents, except for two cases, both of which involved circular references, and which probably require closer review:

1. `ReactServerAgent`: this one had a function that referred to the `API` object, and the methods of the API object call the function. I fixed it by inlining and deleting the function.
2. `PageUtil`: the exported object, `PageUtil`, had a method named `makeArray`, which was referenced as `PageUtil.makeArray` in a bunch of functions in the file. Those functions, in turn, are members of `PageUtil`, causing a circular reference. To break that, I made `makeArray` a top level function; it doesn't use `this` at all and is a simple no side effects helper, then I changed all the calls to `PageUtil.makeArray` to `makeArray`. I left `makeArray` as a member of PageUtil, because I have no idea if clients of this file use it or not. Also of note but unrelated: I had to move around the variable declarations inside of the `PageConfig` definition.

Let me know what y'all think!


aickin:
FYI, because I didn't use comma-chained declaration in `RootElement.js`, this will have merge conflicts with PR #769. If you choose to merge #769 first, I'll happily do the merge; it should not be difficult at all.

packages/react-server-cli/src/compileClient.js
```diff
@@ -13,93 +13,6 @@ import callerDependency from "./callerDependency"
...
 // const logger = logging.getLogger(__LOGGER__);
 
-// compiles the routes file for browser clients using webpack.
-// returns a tuple of { compiler, serverRoutes }. compiler is a webpack compiler
-// that is ready to have run called, and serverRoutes is a promise that resolve to
-// a path to the transpiled server routes file path. The promise only resolves
-// once the compiler has been run. The file path returned from the promise
-// can be required and passed in to reactServer.middleware().
-// TODO: add options for sourcemaps.
-export default (opts = {}) => {
```
gigabo: Ugh.  This is exactly what you said would happen, but it's just as painful as expected.

It's nice to have default exports at the top of modules!

I guess this is the price we pay for the safety this rule provides. 😖

gigabo:
> this will have merge conflicts with PR #769... I'll happily do the merge

Oh, thanks for pointing that out.  Yeah, let's merge that first and have you handle the conflict resolution, since the author over there is a pretty new contributor.

packages/react-server-cli/src/compileClient.js
aickin: I hear you; I think I was the original author of a few of those files in that format.

Another option is to convert functions in those files from `const foo = (bar) => { /* body */}` to `function foo(bar) {/* body */}` and rely on function hoisting, which I *think* would allow you to put the exports at the top in most cases (but maybe not all? haven't thought it through). Less ES6-y and relies on some hoisting magic, but maybe more readable? (Maybe?)

packages/react-server-cli/src/parseCliArgs.js
```diff
@@ -2,6 +2,59 @@ import yargs from "yargs"
 import fs from "fs"
 import pem from "pem"
 
+const sslize = async argv => {
```
gigabo: Oh, heh, you had the [same thought](https://github.com/redfin/react-server/pull/770#discussion_r89396334). 😁

packages/react-server-cli/src/compileClient.js
gigabo: Yeah, IMO the little bit of hoisting magic there is worth it for the benefit of having the first thing you see when you open a module be the part you're looking for.

aickin: OK, gimme a sec. I'll try it out.

aickin:
OK, I updated the PR with a version that just converts `var foo = function() {}` to `function foo() {}` and depends on hoisting. It's much less disruptive (~300 lines changed vs. ~630 in the previous version), and it allows you to keep default exports at the top of the file if that's what you're into.

There are a couple files (`logging/server.js` and `logging/stats.js`) where I converted some functions that didn't strictly need to be converted in order to make the linting rule pass. I thought it was better to do that than to have functions defined a bunch of different ways in the same file.

gigabo:
Oh, man, that's so much better! 😌

gigabo:
> I converted some functions that didn't strictly need to be converted in order to make the linting rule pass. I thought it was better to do that than to have functions defined a bunch of different ways in the same file.

Yeah, much nicer that way.  Thanks for thinking of that.

packages/react-server/core/util/PageUtil.js
```diff
@@ -17,6 +17,85 @@ var {isTheFold, markTheFold} = require('../components/TheFold');
 // These three data structure define the page interface.
 
 
+var PageConfig = (function(){
```
aickin: OK. I was unclear if those comments were for clients of this file (in which case they make more sense to me at the top of the file) or folks who are going to muck with the impl of this file (in which I case they definitely belong below). I'll move them!

aickin: And, I should say, I skimmed the comments at best ;)

gigabo: > I was unclear if those comments were for clients of this file... or folks who are going to muck with the impl of this file

Both. 😹

gigabo:
Okay @aickin I've just merged #769.  Please rebase this when you get a chance.

aickin:
Great! It's rebased, and I'll keep an eye on the CI build; let me know if you need anything more.

gigabo:
Thanks @aickin!

PR #771, opened 2016-11-30
Fix bug with displayAboveTheFold logging

bharath-rengarajan:
displayAboveTheFold metric is used to track when the browser displays all the contents above `TheFold`. We were using the wrong start time while logging this stat.


gigabo:
Yikes, nice catch @bharath-rengarajan!

PR #778, opened 2016-12-02
Add support for logger labels

### LINKED ISSUE #738: TypeError __LOGGER__({ label: "foo" }) is not a function
I tried to use label for a logger like follow 

`var logger = logging.getLogger(__LOGGER__({ label: "foo" }));`

I received that error about `__LOGGER__` : 

`TypeError: (intermediate value)(intermediate value) is not a function`

I'm using node v6.2.2 and the following modules
```
└── react@0.14.8
└── react-server@0.4.8 
└─┬ react-server-cli@0.4.9
  └── webpack@1.13.2 
```

doug-wade:
Fixes #738 by adding support for labels


packages/babel-plugin-react-server/test/fixtures/configurable-token/expected.js
```diff
@@ -1 +1,7 @@
-var logger = require('react-server').logging.getLogger({"name":"babel-plugin-react-server.test.fixtures.configurable-token.actual","color":{"server":133,"client":"rgb(127,42,127)"}});
+var logger = require('react-server').logging.getLogger({
+  name: 'babel-plugin-react-server.test.fixtures.configurable-token.actual',
+  color: {
+    server: 133,
+    client: 'rgb(127,42,127)'
+  }
+});
```
gigabo: Ooh, this is nicer output even without a label!

packages/babel-plugin-react-server/src/index.js
```diff
@@ -1,35 +1,82 @@
...
+					// We're assuming that the options string will only ever contain a
+					// single option, `label`, and that users won't call a tag as a
+					// function without providing a label.  It's the only key we support
+					// (at least, its the only one documented in the gulp module tagger),
+					// but if we ever support more, we'll need to pass the arguments as a
+					// string to loggerSpec (probably by using babel-generator to generate
+					// the code again from the ast) or change the module tagger to take an
+					// options object, rather than a string.
+					const arg = node.arguments[0];
+					options.label = arg.properties[0].value.value;
```
gigabo: `loggerSpec` actually expects the label to be passed in, and it incorporates it into the `name`.

So I think something like this would work:

```javascript
const label = arg.properties[0].value.value;
const options = JSON.parse(loggerSpec({ filePath, trim, prefix, label }));
```

Though we might be able to more confidently pull out the value associated with a `label` key (rather than the value associated with the _first_ key) with something like:

```javascript
const {label} = arg.properties.reduce((o,p) => (o[p.key.value] = p.value.value, o), {});
```

Not familiar with babel plugins, so don't know if that's the right structure for the entries in `arg.properties`.

packages/babel-plugin-react-server/src/index.js
```diff
@@ -1,35 +1,82 @@
...
+			case 'object':
+				literal = convertObjectToObjectExpression(obj[key]);
+				break;
+			default:
+				throw new Error(`got unexpected objecy property type of ${typeof obj[key]}`);
+		}
+		properties.push(t.objectProperty(t.identifier(key), literal));
+	});
+	return t.objectExpression(properties);
+}
```
gigabo: This is awesome. :metal: 

Thanks for tackling this @doug-wade!

packages/babel-plugin-react-server/src/index.js
doug-wade: I wish it was that nice; `optString` also requires the parens from the function invocation; e.g. 

```
const label = arg.properties[0].value.value;
const options = JSON.parse(loggerSpec({ filePath, trim, prefix, optString: `{( label: ${label })` }));
```

I deferred implementation, though, since I'm unclear what the better change is: should I change the `loggerSpec` api to take an options object, or should I use babel-generator to turn the ObjectExpression I have in hand from the AST into a string, put some parens around it and pass that to `loggerSpec`.  Both have downsides: the first introduces a risk of regression on all the other module taggers, and introduces a breaking change; but the second introduces a dependency I shouldn't really need (and using a code generator in a code generator feels... dicey).

packages/babel-plugin-react-server/test/fixtures/configurable-token/expected.js
doug-wade: yeah that hack was pretty hacky...

packages/babel-plugin-react-server/src/index.js
gigabo: > the first introduces a risk of regression on all the other module taggers

We control them all, though, right?  So it's just a matter of updating our modules to conform to a new interface?  Or am I misunderstanding this option?

doug-wade: hypothetically, someone could be using [react-server-module-tagger](https://www.npmjs.com/package/react-server-module-tagger) directly

gigabo: notsureiftrolling

PR #779, opened 2016-12-03
Support labels harder

### LINKED ISSUE #738: TypeError __LOGGER__({ label: "foo" }) is not a function

doug-wade:
Improve module tagger interface per @gigabo's suggestion [in 778](https://github.com/redfin/react-server/pull/778#discussion_r90706006).  Resolves #738


packages/react-server-gulp-module-tagger/index.js
```diff
@@ -22,12 +22,13 @@ module.exports = function(config) {
 	config.basePath = module.filename.replace(THIS_MODULE,'');
 	return forEach(function(stream, file){
 		return stream.pipe(replace(REPLACE_TOKEN, (match, optString) => {
+			const opts = optString ? {label: optString.replace(/.+label:\s?"(\w+)".+/, "$1")} : undefined;
```
roblg: Couple questions about this regex:
- it looks like it assumes the label in opts will use a double quote, which is true in literally every case I could find, except for one: `PageUtil.js` :)
```
40365: 		var logger = __webpack_require__(303).getLogger({ "label": "{label: 'PageConfig'}", "name": "react-server.core.util.PageUtil.{label: 'PageConfig'}", "color": { "server": 217, "client": "rgb(212,127,127)" } });
```
Seems easy enough to support either quoting style though.

- I'm not sure if we care all that much, since I couldn't find any of these usages, but it looks like with the previous impl, we could write:
```
logger.getLogger(__LOGGER__({
   label: "foo"
}));
```
and it would work correctly. This regex causes invalid syntax to be output in this case, which causes the gulp build to barf. (I tested by adding a multiline options object to PageUtil.js and running `npm run prepublish`)

gigabo: Yeah, the eval hack was pure evil, but it worked well. 👹

doug-wade: I added a test case for single quotes.  I'd be okay with postponing multiline options until we have more than one key supported by logger options (which might make it unreasonably difficult to maintain a reasonable maximum line width) or there's a concrete need for it -- what do you think?

packages/babel-plugin-react-server/src/index.js
```diff
@@ -31,18 +31,7 @@ module.exports = function() {
...
-					// single option, `label`, and that users won't call a tag as a
-					// function without providing a label.  It's the only key we support
-					// (at least, its the only one documented in the gulp module tagger),
-					// but if we ever support more, we'll need to pass the arguments as a
-					// string to loggerSpec (probably by using babel-generator to generate
-					// the code again from the ast) or change the module tagger to take an
-					// options object, rather than a string.
-					const arg = node.arguments[0];
-					options.label = arg.properties[0].value.value;
+					const options = JSON.parse(loggerSpec({ filePath, trim, prefix, opts: {label: node.arguments[0].properties[0].value.value} }));
```
doug-wade: I've added a test case

packages/react-server-gulp-module-tagger/index.js
roblg: Sounds good to me. Let's file the multi-line options issue? It's probably on the challenging side, but I think building on top of this PR would make it super narrowly-scoped, so there might be someone out there who wants to tackle it.

PR #780, opened 2016-12-03
Upgrade examples and templates to 0.5

doug-wade:
Upgrade the examples and templates to use react-server 0.5 and react 15


gigabo:
Hmm... we should actually be able to have Asini keep React Server deps up-to-date for us now that it supports configurable package locations. :thinking:

gigabo:
Thanks @doug-wade!

PR #781, opened 2016-12-03
Add styled components example

doug-wade:
Demonstrate how to use [styled-components](https://github.com/styled-components/styled-components) with react-server


egid:
Like I mentioned last night, everything looks good aside from swapping themes on the fly, which updates the props but doesn't apply them to the child components. I don't know enough about styled components to provide any meaningful help with that, though.

doug-wade:
@egid I don't know what's up -- its working for me

egid:
Good call, that fixed it. Working as expected!

PR #782, opened 2016-12-03
Support legacy webpack config

doug-wade:
Prevent @drewpc's mystery for other users: https://github.com/redfin/react-server/commit/a23199e5506794feeefca16228001645b19b1a27


doug-wade:
This is a one-off implementation of https://github.com/redfin/react-server/issues/783 for only this config key

PR #784, opened 2016-12-07
Run bootstrap as a postinstall script

doug-wade:
Per https://react-server.slack.com/archives/discussion/p1481135585000084


doug-wade:
:rocket:

Issue #312, opened 2016-06-28
Requests other than "get" is not working


mattiasewers:
When setting `method:  'post` and trying to make post request to a the route I get `Cannot POST /foo`. Looked in the docs and didn't find if any config is needed for this.

gigabo:
Oh, yeah... we actually have an _old_ ticket open for that internally from before we even open-sourced `react-server`...

As a workaround you can actually just leave it `method: "get"` and it will accept `POST` requests. 😬 

Thanks for reporting @mattiasewers.  Good to have that tracked here.

mattiasewers:
@gigabo ahh I see, cool. Another thing, how do I get the payload sent with the request? `this.getRequest().getBody()` returns `undefined`, although I get the data in the config object return from the request....weird

gigabo:
Ah, yeah.  We use the [`body-parser`](https://github.com/expressjs/body-parser) express middleware internally.  #249 is open to add that to the server set up by `react-server-cli`.  Should be a straightforward patch.

mattiasewers:
@gigabo ok I got it working, I could make a PR?

gigabo:
Nice! Please do!

drewpc:
Solving this issue and addressing the [breaking changes](https://github.com/yahoo/routr/blob/master/CHANGELOG.md#breaking-changes-1) when upgrading `routr` from 0.1.3 to >1.0 need to be addressed prior to the next release.

drewpc:
Fixed!

PR #787, opened 2016-12-08
Update dependencies with david

doug-wade:
Since we merged https://github.com/redfin/react-server/pull/732, greenkeeper hasn't been keeping our deps up-to-date, and it never updated our packages dependencies.  Use [david](https://david-dm.org/) to update our dependencies instead, and upgrade most of them (`chunk-manifest-webpack-plugin` is for webpack 2, for instance) to the most recent available version


package.json
```diff
@@ -8,11 +8,13 @@
     "nuke": "asini clean && rm -r node_modules",
     "changelog": "asini-changelog",
     "debug": "cat asini-debug.log && for d in packages/*/npm-debug.log*; do echo $d; cat $d; done",
-    "postinstall": "npm run bootstrap-no-i"
+    "postinstall": "npm run bootstrap-no-i",
+    "update-dependencies": "asini exec david u"
```
gigabo: Should this also run david on the repo root?  Then we could disable the GreenKeeper integration and just manage everything with david?

gigabo:
Wow, thanks for dealing with this @doug-wade!

This seems like a generally pretty high-risk change... :thinking: 

Any additional testing we should do for this?

gigabo:
(Edited the title to reflect that this PR actually updates deps)

doug-wade:
@gigabo I've done all the testing I think I can -- it's passing ci, I linked up all the examples and ran them, and I ran through the test pages.  I agree there are risks, but I think we should try to keep the ecosystem up-to-date as much as reasonable.

gigabo:
@doug-wade Thanks for manually regressing.  Bombs away! 💣

packages/react-server/package.json
```diff
@@ -24,50 +24,50 @@
...
     "lodash": "^4.16.4",
-    "mobile-detect": "1.3.0",
+    "mobile-detect": "^1.3.5",
     "q": "1.4.1",
     "querystring": "0.2.0",
-    "request-local-storage": "1.1.0",
-    "routr": "0.1.3",
-    "winston": "2.2.0"
+    "request-local-storage": "^1.1.1",
+    "routr": "^2.1.0",
```
drewpc: @gigabo The upgrade of `routr` from version 0.1.3 to >1.0 introduced some [breaking changes](https://github.com/yahoo/routr/blob/master/CHANGELOG.md#breaking-changes-1) that impact us.  These changes and the bug in #312 are related, so I'll track it there.

doug-wade: how do those changes affect us?  and why didn't the automated tests catch the problems?

drewpc: See #837 and #312.  The change had a few impacts:

1. They no longer do anything with the `navigate` option in https://github.com/redfin/react-server/blob/master/packages/react-server/core/context/Navigator.js#L53.  I have NO idea with the `type` variable is really referring to in our code...it seems super weird to me.  The Routr project said in their [commit message](https://github.com/yahoo/routr/pull/29): "If needed, the navigation logic should happen at a higher level or remove that special case."
2. According to #312, we have never been passing the HTTP method information to Routr anyway, so Routr never checked the method properties.
3. We are improperly encoding the page attribute `method` in the `routes_(client|server).js` file  because our documentation says it can be an array or string (Routr now supports an array), but we're encoding it as a string only.

drewpc: The tests didn't catch this because all of our tests only send GET requests, no POST/HEAD/etc.  I added some unit tests, but we should probably add full integration tests for at least POST requests.

drewpc: Let me rephrase this.  The breaking change in Routr *may or may not* impact us.  It all has to do with what we use/expect regarding the `type` property on [this line](https://github.com/redfin/react-server/blob/master/packages/react-server/core/context/Navigator.js#L53).  Our tests might not catch this for two reasons: we don't have test coverage of this case or the change in Routr actually doesn't impact us.

Secondly, the bug identified in #312 means that we're not even passing the HTTP method to Routr for it to check.  The fix for their API change and #312 is to pass the HTTP method to Routr.  *That* change is a breaking change for *our* project because now people will have to allow POST requests in their `routesFile` or else Routr won't select them.

Issue #773, opened 2016-11-30
Add option to pre-babelify for production


gigabo:
Currently the CLI always runs through babel register to run transforms on the fly.  Need an option to save the transformed es5 output and run straight from that in production for faster startup.

gigabo:
I guess another option here would be to run webpack bundles on the server.  I remember thinking at one time that there were some benefits to that, but I can't recall what they were at the moment. 😅

doug-wade:
@gigabo the reason that I remember off the top of my head is making the server and the client build easier to reason about.  Currently, you have to ship everything twice, once as a `require` hook server-side, and once as a webpack hook client-side.  If we moved to webpack-only, it would likely be easier to reason about and explain (and would give users the ability to modify how modules are loaded server-side, which they can't do currently, by modifying their webpack config).

aickin:
@gigabo are you talking about the use of `babel-register` to transform files as part of `require` on the server side? If so, I'm pretty sure that it [keeps a cache in between runs](https://babeljs.io/docs/usage/babel-register/#environment-variables) automagically. 

Now, maybe it's not using the cache for some reason, but if it is, I'd guess that pre-transforming isn't going to help very much.

drewpc:
Maybe this is a bit off topic, but I'm for anything that makes the client and server sides mirror each other.  If Webpack is used for the client side, it should be for the server side as well.  It seems easier to understand/develop a truly isomorphic app if the client and server are as close to the same as possible.

gigabo:
@aickin Oh, interesting.  Maybe we could set `process.env.BABEL_CACHE_PATH` to the build directory and ship that (along with the webpack manifest.json) for fast startup in production?

@drewpc Not off topic at all!  That's a good point about keeping the client and server as close to each other as possible.  I'm not so familiar with webpack on the server.  Is that a common practice?

roblg:
I think next.js is using webpack on the server as well. One benefit is that then you can just delete the single built asset from the `require.cache`, and you get server-side "hot reload". (Or if you have multiple bundles, you have to delete more than one, but still)

gigabo:
> server-side "hot reload"

![ib](https://cloud.githubusercontent.com/assets/115908/20806827/45442de8-b7b1-11e6-9ee8-1d0f0a0c1b46.gif)

drewpc:
[...]

Issue #759, opened 2016-11-15
Unexpected token in required images


svassr:
Having images integrated in some component like follow, I'm frequently receiving errors like the next following

```
<img src={require('../images/logo.png')} >
```

```
2016-11-15T21:22:54.636Z - error: [react-server.core.renderMiddleware] Error with element Unknown's lifecycle methods message=/Users/me/Projects/myProject/dev/images/logo.png: Unexpected character '�' (1:0), stack=SyntaxError: /Users/me/Projects/myProject/dev/images/logo.png: Unexpected character '�' (1:0)
> 1 | �PNG
    | ^
  2 | 
  3 | 
  4 | IHDR�8�h	pHYs

                    ��
    at Parser.pp.raise (/Users/me/Projects/myProject/dev/node_modules/babylon/lib/parser/location.js:22:13)
    at Parser.getTokenFromCode (/Users/me/Projects/myProject/dev/node_modules/babylon/lib/tokenizer/index.js:525:10)
    at Parser.readToken (/Users/me/Projects/myProject/dev/node_modules/babylon/lib/tokenizer/index.js:154:19)
    at Parser.<anonymous> (/Users/me/Projects/myProject/dev/node_modules/babylon/lib/plugins/jsx/index.js:49:20)
    at Parser.readToken (/Users/me/Projects/myProject/dev/node_modules/babylon/lib/plugins/flow.js:179:22)
    at Parser.nextToken (/Users/me/Projects/myProject/dev/node_modules/babylon/lib/tokenizer/index.js:144:19)
    at Parser.parse (/Users/me/Projects/myProject/dev/node_modules/babylon/lib/parser/index.js:95:10)
    at parse (/Users/me/Projects/myProject/dev/node_modules/babylon/lib/index.js:45:47)
    at File.parse (/Users/me/Projects/myProject/dev/node_modules/babel-core/lib/transformation/file/index.js:532:15)
    at File.parseCode (/Users/me/Projects/myProject/dev/node_modules/babel-core/lib/transformation/file/index.js:617:20)
2016-11-15T21:22:54.636Z - fine: [react-server.core.renderMiddleware] renderElement.individual.Unknown ms=200
```

Looks like babel try to validate the png file content or something. Could it be a preloader? I already checked webpack module loaders and doesn't seem something is wrong there.

I'm using node v6.2.2 and the following modules
```
└── react@0.14.8
└── react-server@0.4.8 
└─┬ react-server-cli@0.4.9
[...]

Issue #649, opened 2016-08-30
Problem with static asset server & secure connections


hzhu:
There are problems with the [static asset server and secure connections](https://react-server.slack.com/archives/discussion/p1472252040000870) recently -- the wrong URL is being passed to the loading middleware.

Deployed react-server [app](https://react-server-redux.herokuapp.com/) is logging an error related to static assets.
<img width="1090" alt="screen shot 2016-08-30 at 12 59 48 pm" src="https://cloud.githubusercontent.com/assets/1811365/18104821/b7f1b1b0-6eb1-11e6-86f3-b21de4047469.png">

kennethormandy:
I’m having this same problem.

The only way I can think of to get around it is to publish the static assets elsewhere and then pass in a `--js-url`, but even that would all need to get deployed on the server because the hashes change vs. compiling locally.

It looks like [this is the line](https://github.com/redfin/react-server/blob/5e3714e119a3b7994adfe441856e43b014e0eb4e/packages/react-server-cli/src/compileClient.js#L58) where the `outputUrl` is used and I think it’s being set incorrectly [here](https://github.com/redfin/react-server/blob/5e3714e119a3b7994adfe441856e43b014e0eb4e/packages/react-server-cli/src/run.js#L33).

I think the protocol is based on the `https` config option instead of whether the request coming into the server is using https, which probably only makes sense in development mode. Maybe a quick fix would be to just use `//` instead of the protocol? I know that’s not the fastest option, but it would be a start.

drewpc:
The next release of `react-server` will include a fix to handle the issue described here.  After it is released, please try it again and let us know if there are any further issues.

doug-wade:
@drewpc this was fixed in #836?

drewpc:
Yes, because it uses a protocol relative request now.

Issue #622, opened 2016-08-25
How to add loader in Bike Share example?


incubus8:
Hi guys & redfin,

Any idea, How to add loader for main content in bike share example https://github.com/redfin/react-server/tree/master/packages/react-server-examples/bike-share?

Also it seems that the footer section will not be rendered before the content is rendered / timeout. Or on the other hand, the footer render is being blocked by the main content to be rendered. How can we make it async render like redux async and universal rendering?

gigabo:
Hi @incubus8 - On the _server_ the footer will render right away, but the resulting HTML string obviously needs to be held until the preceding content is sent.

In the _browser_ we render elements in order to avoid having content appear and then get pushed down the page.  Rendering later elements as they become ready could work well in a context where they won't get moved around later, but that's not the common case.  Might work as an option?

doug-wade:
@incubus8 Currently, React Server uses webpack only for the client, and not for the server.  We've had some discussion recently about moving to webpacking the server too, but for now, we use `require.extensions` to [override how files are loaded server side](https://github.com/redfin/react-server/blob/master/packages/react-server-cli/src/index.js#L5).

As to your question about the footer being rendered before the main content, this is intentional.  One of the things we've focused on for React Server is that the page loads smoothly and consistently, and that it avoids the "popping" effect that you get when you allow elements lower on the page to render before elements higher on the page.  We believe that this leads to better customer experience, since it avoids behaviors like buttons moving out from under your cursor or thumb when you go to click a button on a page -- you know that once it is rendered, it shouldn't "jump" down the page.

incubus8:
@doug-wade  @gigabo Sorry late reply, I think let's us firstly forget about customer experience. Since it is just for demo purposes. I know well that it will have "popping" effect / the content pushed down the page.

Apart from that, I also think it is important to have async browser rendering, not just holding data in server and pushed it sequentially after the waited request timeout/error. I have found some other implementations, it may differ from what ReactServer does. Here are the resource examples I found:
1. http://www.juhonkoti.net/bigpipe/example.php
2. http://chrischandler.name/

Now, notice that above resources are displaying the loader or giving user information to wait for the data being streamed from server. What do you think?

mocheng:
[...]

PR #791, opened 2016-12-20
Server Side HMR/Webpack & Other Improvements

drewpc:
This PR covers a number of things that are outlined below.  I'm pretty much at a stopping point where I need some assistance from others.  The unchecked items are things that aren't working/implemented and I would love some assistance.  Feedback on everything would be great.

- [x] Combined the JS and HTML servers into one.  Now, there is a single ExpressJS server created and `webpack-dev-middleware` is added to it only if the server is started with `hot: true`
- [x] Removed entries throughout the application for `jsPort` since it is no longer required.
- [x] The default jsUrl is now `/` instead of `http[s]?://{host}:{jsPort}`.  This makes loading the assets cleaner because the URL is relative instead of absolute.
- [x] Created two Webpack compiler paths: client and server bundles.  Both outputs live in __clientTemp to prevent breaking existing applications.  This covers issues: #773, #759, #649, #622, and #595 
- [x] Enabled hot module reload of server-side assets after they are compiled by Webpack.  Discussed in issue #773.
- [x] Added an example showing some of these features.
- [x] The server watches for changes in `.reactserverrc`, `routes.js[on]`, and any custom webpack configs.  If any of those files change, the server restarts itself.  Other files that change just cause a recompile by Webpack.
- [x] Added a `compileOnStartup` option to `.reactserverrc` that allows you to disable compilation when starting the server.  Discussed in #773.
- [x] Some of the modules installed in `node_modules` are being bundled into the server side Webpack config instead of left out.  I need help tracking down where/why this is happening because I have the node modules all listed as externals in the Webpack config. UPDATE (12/26/2016): fixed 
- [x] Some of the tests are blowing up and I don't understand how they are working.  Can someone help me look into this?  UPDATE (12/29/2016): All tests are fixed.
- [ ] Server-side chunking isn't working properly.  For some reason, I can't get any of the existing/previous manifest stuff to work properly and I need to discuss how all of this is supposed to work with someone.  @jhnns - can you help with this?  UPDATE (12/26/2016): spent ~6 hours working on this and it seems that chunked code being bundled for Node with a libraryTarget of 'commonjs2' sets module.exports but it sets it to an array like this: [2] (the chunk ID).  It doesn't actually export any of the module code, therefore the bundle can't be imported.  If I limit the number of chunks to 1 and disable a common chunk, modules.exports is set to the webpackBootstrap function itself.


packages/react-server-cli/src/buildWebpackCompilers.js
```diff
@@ -324,37 +204,21 @@ module.exports = {
...
-	var outputFile = outputDir + "/entry.js";
+function writeClientBootstrapFile(outputFile, opts) {
 	fs.writeFileSync(outputFile, `
 		if (typeof window !== "undefined") {
 			window.__setReactServerBase = function(path) {
-				// according to http://webpack.github.io/docs/configuration.html#output-publicpath
-				// we should never set __webpack_public_path__ when hot module replacement is on.
-				if (!module.hot) {
-					__webpack_public_path__ = path;
-				}
```
drewpc: Yes, they don't.  Here are the references:

http://code.fitness/post/2016/02/webpack-public-path-and-hot-reload.html#overriding-the-public-path:12018ae465828e3168201053550649e1

https://github.com/webpack/webpack/issues/1650

packages/react-server/core/renderMiddleware.js
```diff
@@ -43,18 +42,8 @@ var ELEMENT_ALREADY_WRITTEN = -2;
...
-	expressState.extend(server);
-
-	// parse cookies into req.cookies property
-	server.use(cookieParser());
-
-	// sets the namespace that data will be exposed into client-side
-	// TODO: express-state doesn't do much for us until we're using a templating library
-	server.set('state namespace', '__reactServerState');
-
-	server.use((req, res, next) => { RequestLocalStorage.startRequest(() => {
```
drewpc: I purposely didn't touch that...although it was tempting 😄

roblg: `:facepalm:`. I see what happened. github didn't render the `start.js` diff at page load (I had to click on it), so it looked like these had been removed rather than just moved. I get it now. :)

jhnns:
@drewpc Thank you for your suggestions. I reviewed your PR and tried to understand as much as possible about what's going on in react-server under the hood. I haven't a full understanding yet, but I've already built some universal javascript apps and know the pitfalls. Let me share my experience and my current knowledge with you all (@roblg @doug-wade).

First of all, I think it is very important to also bundle your server-code with webpack as it streamlines your environment. Code re-use is easier when the client and server environment is as similar as possible. That's why this change feels like a "must have" for me. It also fixes some common problems with the current setup as already mentioned in https://github.com/redfin/react-server/issues/773#issuecomment-268033323 

Unfortunately, building your project with webpack two times can lead to an unsatisfying developer experience. In a recent project of mine, the initial build time in development was around 50s. The project is medium-sized, so I'd expect even longer build times in bigger projects.

The main reason for this was that the build was done sequentially, very similar to your current implementation (first client, then server). You could try to start both builds in parallel, but this leads to another problem: Many webpack loaders, like the babel-loader and the postcss-loader, are hogging the CPU. Since these loaders perform AST transformations entirely in node, node's single-threaded architecture becomes a real problem. With this architecture, you can't benefit from multi-core processors. In order to illustrate the problem, I've recorded a flame graph in this medium sized project I've already mentioned:

flame graph

You can clearly see the client and the server compilation – and that the CPU is always busy :)

We've spotted this issue already in the webpack community and are currently reworking the loader API so that loader execution can be pushed to different processes in the future. But this will still take some time.

---

I've refactored the build architecture in this project and came to the following solution.

In development, you have:

dev architecture

- the main development process (MDP) was invoked by the developer (like `npm start`) and is in charge of orchestrating the following processes
- the webpack-dev-server (WDS) process which watches and compiles the client code in memory
- a webpack watcher process (WWP) which watches and compiles the server code to an output folder (like `dist`)
- the application server (AS) (usually just your express server) which picks up the server code from `dist`

All 3 child processes are forked by the MDP and report their status to the MDP via node's built-in IPC channel.

The AS is the only process which is killed and restarted as soon as the WWP reported a successful build.

The WDS proxies all requests that do not match a static resource generated by webpack to the AS (usually your HTML pages). One gotcha is, that you have to defer all incoming requests as long as the AS is restarting – otherwise the developer will get ugly 404s because the proxy is broken during that period.

In production, this setup is much simpler of course. You just need to invoke webpack with two different configs and output the result into `dist` and `public`. Then you start the AS which picks up the server built from `dist` and serves the static files from `public`.

With this architecture and by using [happypack](https://github.com/amireh/happypack) for parallel loader execution, I was able to reduce the initial build time from 50s down to 9s.

---

Since I will be using react-server in production, I would be happy to contribute to this. If you're interested in this solution, I can also give more explanations and create a PR for it. I would use @drewpc 's PR as a starting point.

drewpc:
@jhnns I'd love to discuss this further on Slack if you have signed up.  A couple of points that I'd like to respond to:

1. There are two reasons I chained the two compile processes together (listed below).  I do not want them chained together...parallel is much better
    a.  `webpack-dev-middleware` only supports one compiler output
    b.  As of right now, the server compilation requires the chunk manifest from the client compilation.  I want to do some more testing to see if I can rip that part out and not make them dependent on each other.  That's part of why I asked for some assistance in that area.
2.  I don't want to use `webpack-dev-server` because we don't need to have applications listening on different ports (it creates complexity where it doesn't need to be).  We can use `webpack-dev-middleware` and accomplish the same goals--combining the WDS and AS components from your diagram.  That also allows us to use relative URLs in various places, which further reduces complexity.
3.  I added the WWP component from your diagram already, but I was only watching for changes to specific `react-server` files such as `.reactserverrc` and `routes.js[on]`.  If I remove the chain between client and server, I can still use the `webpack-dev-middelware` compile event notifications to trigger a compilation of the server code, therefore not needing a WWP component that is different than what I already have now.
4.  Can we achieve the same concept that you're recommending by adding `webpack-dev-middleware` as a forked process to ExpressJS?  That way it's still "middleware" but in a separate process?  The server-side compilation can definitely be forked.

jhnns:
> webpack-dev-middleware only supports one compiler output

That's ok since the webpack-dev-middleware should only be concerned with bundling and serving the client code

> As of right now, the server compilation requires the chunk manifest from the client compilation

Yes, I've already seen this in universal javascript setups. I don't think that this is a regular dependency because the contents are unknown at build time. It's a *dynamic dependency* which can not be handled by webpack. I think it's better to `fs.readFileSync` the JSON on application startup.

> I don't want to use webpack-dev-server because we don't need to have applications listening on different ports

I do understand this. It feels a little more complex but I think it's necessary. You need a server that accepts all requests while the application server is restarting. There will always be the case where HMR is not possible (for instance when the module can not be replaced without side-effects). In this case, the webpack client will automatically refresh the page. If the server is currently restarting, the developer would be left with a broken 404 page.

> I can still use the webpack-dev-middelware compile event notifications to trigger a compilation of the server code

I'm not sure if I understood this correctly, but you need to keep in mind that it is possible that client and server do not share the same set of files. It's also more efficient to keep the webpack compilation on hold (with all the dependencies in memory) than to "just" trigger a new compilation.

I'll sign up to Slack during the next days so that we can discuss this further :) although I think it's good that it's documented here.

packages/react-server/core/renderMiddleware.js
```diff
@@ -494,8 +483,18 @@ function renderScriptsAsync(scripts, res) {
...
+			// execution.  Thus, we need to force LABjs to not use local XHRs in order for all tests to work properly.
+			// This used to work prior to combining the HTML and JS servers into a single server because the JS files
+			// were considered part of a different domain when running tests, so it never used XHR preloading, it used
+			// "cache preloading".
+			// https://github.com/getify/LABjs/blob/master/LAB.src.js#L152-L175
+			res.write("$LAB.setGlobalDefaults({UseLocalXHR:false,AlwaysPreserveOrder:true});");
+		} else {
+			// We always want scripts to be executed in order.
+			res.write("$LAB.setGlobalDefaults({AlwaysPreserveOrder:true});");
+		}
```
roblg: Good find!

roblg:
@drewpc sorry for the delay getting to this. I got sick right before Christmas and wasn't in any shape to think :(

This looks really good. I'm taking it for a spin right now, and I might have more feedback, but overall I'm really excited.

re: @jhnns 's suggested architecture -- that architecture looks _eerily_ similar to what Redfin is using internally right now (our internal build predates the `react-server-cli` package), and I'm pretty confident that getting Redfin onto `react-server-cli` would require a similar architecture. I also hadn't seen `happypack` before -- I'm going to try that out internally and see if it can't solve some other problems we've been having :grin:

I'm curious to hear @gigabo's thoughts as well, but I think I'm OK merging this first (when it's ready), and then @jhnns (or whoever) can submit a patch for the proxy/separate process architecture at some point in the future. The main downside I see to that is that there's a little bit of back-and-forth code shuffling around webpack-dev-middleware vs. webpack-dev-server, and we need to solve the chunk manifest thing, but the upside is better support for hot reload in the meantime.

Thoughts?

packages/react-server/core/renderMiddleware.js
roblg: Not a problem!

I added the `breaking change` label to this so that we remember to update our code internally when we merge it.

jhnns:
> that architecture looks eerily similar to what Redfin is using internally right now

That's funny. I'm not a spy I promise 😁. But I think it confirms that this solution has some real benefits. It's probably possible to omit the webpack-dev-server, but for me, this is a clean separation of concerns. And with the webpack-dev-server you will be using a module that is maintained by the webpack team and widely used in the webpack community (no offense to the author(s) of webpack-dev-middleware, it's definitely a great addition to the eco system 👍).

> but I think I'm OK merging this first (when it's ready), and then @jhnns (or whoever) can submit a patch for the proxy/separate process architecture at some point in the future

Before merging it, you should run this PR against a sufficient big code base to check the build time. If my assumption is correct, it would double the build time which can be pretty annoying for your existing users.

Again: No offense against @drewpc 's awesome work. For me, it's a great blueprint! It's just that I went this way already and I know the downsides of it.

I'm able to dedicate some time for my proposal starting with Jan, 2th. I just need to know if you're still interested and what my branch should be based upon (this PR or master).

drewpc:
@jhnns I'm all for your suggestions, thanks for helping out!  Here's my defense of `webpack-dev-middleware` over `webpack-dev-server`:

- The middleware component is written by @webpack as well.
- The dev server component uses the middleware component for all of the actual compilation (see https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L53)
- I assume  `react-server` is using `webpack-dev-(server|middleware)` for is the file watching + compilation, not necessarily to have an additional TCP port to listen on.
- I TOTALLY agree that having this in a separate process would be better for CPU utilization.  Can we do this without listening on a  separate TCP port?

I'll look at detaching the server side compilation so that it runs in parallel instead of in sequence today.

drewpc:
[trivago/parallel-webpack](https://github.com/trivago/parallel-webpack) has some nice features on creating multiple Webpack variant configurations that could be useful.  It doesn't hook into `webpack-dev-(server|middleware)`, so you don't get some of the in-memory and direct HTTP serving stuff, but it could be a good fit in other areas.  I'll look into the `worker-farm` stuff too.  That may be the best way to fork at least some of the compilation stuff.

jhnns:
@drewpc

> The middleware component is written by @webpack as well.

I'm sorry, you're right. Must have mixed something up here 😁 

> I assume react-server is using webpack-dev-(server|middleware) for the file watching + compilation, not necessarily to have an additional TCP port to listen on.

The most important benefit of the webpack-dev-server is that you have always a server that accepts incoming requests, even when the application server is currently restarting. We *could* implement HMR for the server, but in my experience there are always parts of the application that cannot be hot replaced and thus need to be restarted.

Imagine, there is a module that cannot be hot replaced. webpack will bail out from hot replacement and reload the website. If the server is currently also restarting, the developer will get a broken page. It *could be* possible that HMR will always be possible in react-server's case (I don't know), but only as long as the user sticks to the react-server architecture with its pages and components. 

---

@roblg

> I'm not sure it'll be 2x slower, because with @drewpc's setup here, there's no webpack-ing of node_modules for the server.

That optimization might make sense in many use cases, but in my particular use case this would be a problem. We have separate application modules located in node_modules that need to be webpacked. We cannot webpack them before publishing them because that would make some optimizations impossible. So, hopefully I'm able to override the webpack config in this regard.

> My suggestion that we merge this first was under the assumption that there aren't currently very large apps using react-server-cli for their build. (Redfin is using react-server, and our code base is pretty large IMO, but we have a custom build for our app that predates react-server-cli.) I could totally be wrong about this though, and I would actually be pleased to find out I'm wrong

The client I'm working for is currently going to build a large scale app using react-server-cli :grin:. That's why I would like to help to improve the build process.

> The proxy ("main development process") that dispatches incoming requests to the right server, (based on URL prefix, I presume)

The "main development process" does not listen on a port. It just receives messages from and sends messages to its child processes. The webpack-dev-server allows to configure a proxy that proxies all urls that are not provided by the webpack bundle. So, instead of showing a 404, it just proxies the request.

> With this setup, the additional port occupied by the webpack-dev-server would be an implementation detail, as opposed to something that has to be handled and set by the developer

The additional port is definitely an implementation detail the developer does not need to know. Especially since there is no additional port in production. That's why I don't see that the additional port is any problem that makes things more complex. It's just the result of separated processes.

---

That being said, it is possible to build this without the webpack-dev-server. But then I don't know how you could safely restart the server without risking to break the page in the browser. How is this currently done?

Personally I don't think that the multi-process architecture makes things more complicated. It just separates things more cleanly and gives us the ability to restart parts of the application without affecting other parts.

packages/react-server-cli/package.json
```diff
@@ -44,8 +46,9 @@
     "sass-loader": "^4.0.0",
     "style-loader": "~0.12.3",
     "webpack": "^1.13.1",
-    "webpack-dev-server": "~1.13.0",
-    "webpack-stats-plugin": "^0.1.3",
+    "webpack-dev-middleware": "^1.9.0",
+    "webpack-hot-middleware": "^2.13.2",
```
jhnns: Ahh... that's why I mixed this up. I meant that `webpack-hot-middleware` is not maintained by the webpack team.

packages/react-server-cli/package.json
```diff
@@ -44,8 +46,9 @@
...
+    "webpack-stats-plugin": "^0.1.4",
```
jhnns: It looks like this module is not actually required. The generated JSON file is never read. Or am I wrong?

packages/react-server-cli/src/buildWebpackCompilers.js
```diff
@@ -0,0 +1,208 @@
...
+	} = webpackInfo.paths;
+
+	mkdirp.sync(workingDirAbsolute);
+	mkdirp.sync(outputDirAbsolute);
+	mkdirp.sync(serverOutputDirAbsolute);
+
+	writeClientBootstrapFile(clientBootstrapFile, opts);
+
+	// now rewrite the routes file out in a webpack-compatible way.
+	writeWebpackCompatibleRoutesFile(routes, routesDir, workingDirAbsolute, null, true);
```
jhnns: I know that this function was already called at this location before the PR, but why is it necessary?

packages/react-server-cli/src/buildWebpackCompilers.js
```diff
@@ -0,0 +1,208 @@
...
+		webpackInfo.client.compiler.plugin("done", (stats) => {
+			const manifest = statsToManifest(stats);
+
+			// TODO: need to figure out why these two manifest files aren't working or used.
+			//fs.writeFileSync(path.join(outputDirAbsolute, "manifest.json"), JSON.stringify(manifest));
+
+			// this file is generated by the build in non-hot mode, but we don't need
+			// it (we got the data ourselves from stats in statsToManifest()).
+			if (!hot && longTermCaching) {
+				//fs.unlinkSync(path.join(outputDirAbsolute, "chunk-manifest.json"));
```
jhnns: If the file is not needed, why not remove the chunk-manifest-webpack-plugin?

packages/react-server-cli/src/buildWebpackConfigs.js
```diff
@@ -0,0 +1,355 @@
...
+			path: outputDir,
+			filename: '[name].bundle.js',
+			chunkFilename: '[id].bundle.js',
+			libraryTarget: 'commonjs2',
+			pathinfo: true,
+		},
+		devtool: 'source-map',
+		plugins: [
+			...commonWebpackConfig.plugins,
+			new webpack.SourceMapDevToolPlugin(),
```
jhnns: This plugin and `devtool: "source-map"` are mutual exclusive. In fact, the `devtool` option is a shortcut for the plugin. You should only use one of them. I would stick to the `devtool` option.

packages/react-server-cli/src/buildWebpackConfigs.js
```diff
@@ -0,0 +1,355 @@
...
+			new webpack.NormalModuleReplacementPlugin(/\.(css|less|sass|scss)$/, 'node-noop'),
```
jhnns: This is a well intended performance improvement, but it breaks CSS modules. In order to determine CSS class names, we must compile all CSS files for the server code as well. We just need to omit the style-loader because it depends on the DOM.

packages/react-server-cli/src/buildWebpackCompilers.js
```diff
@@ -0,0 +1,208 @@
...
+	const routesFilePath = `${workingDirAbsolute}/routes_${isClient ? "client" : "server"}.js`;
+	fs.writeFileSync(routesFilePath, routesContent);
+
+	return routesFilePath;
+}
+
+// writes out a bootstrap file for the client which in turn includes the client
+// routes file. note that outputDir must be the same directory as the client routes
+// file, which must be named "routes_client".
+function writeClientBootstrapFile(outputFile, opts) {
```
jhnns: This file doesn't need to be actually written to disk. There are several possibilities to achieve the same effect. You could use the DefinePlugin to define the log levels. You could also use the imports-loader to inject build time variables.

packages/react-server-cli/src/buildWebpackConfigs.js
```diff
@@ -0,0 +1,355 @@
...
+			alias: {
+				// These need to be singletons.
+				"react"        : callerDependency("react"),
+				"react-dom"    : callerDependency("react-dom"),
+				"react-server" : callerDependency("react-server"),
+			},
+		},
+		resolveLoader: {
+			root: [
+				path.resolve(path.join(path.dirname(require.resolve("webpack")), "../..")),
```
jhnns: What was the original motivation to change the resolving behavior?

packages/react-server-cli/src/buildWebpackConfigs.js
```diff
@@ -0,0 +1,355 @@
...
+			chunkFilename: '[id].bundle.js',
+			libraryTarget: 'commonjs2',
+			pathinfo: true,
+		},
+		devtool: 'source-map',
+		plugins: [
+			...commonWebpackConfig.plugins,
+			new webpack.SourceMapDevToolPlugin(),
+			new webpack.NormalModuleReplacementPlugin(/\.(css|less|sass|scss)$/, 'node-noop'),
+			new webpack.BannerPlugin('require("source-map-support").install();', {
```
jhnns: It would be sufficient to put this in our `routes_server.js`

packages/react-server/core/renderMiddleware.js
jhnns: I don't know enough about react-server yet, but why is LAB necessary at all? Shouldn't all scripts be bundled with webpack?

roblg: LABjs allows us to load the webpack scripts themselves at page load as async as possible in all browsers to prevent blocking paint, etc.

jhnns: Why is it no option to include a `<script>` tag with a [`defer`](http://calendar.perfplanet.com/2016/prefer-defer-over-async/) attribute?

drewpc: 1. This is a single loader that loads external script files, internal script files, and inline `<script>` tags for consistency.
2. It notifies the caller when scripts have finished loading
3. It guarantees order of execution and can `wait()` for required scripts to load before execution.

I'd recommend moving this type of discussion to a different area to discuss architecture decisions.

karlhorky:
Looking forward to this change! What is remaining to get this merged?

drewpc:
@karlhorky There's some discussion about how to move forward with some of the ideas that I've proposed in this PR.  That and I kinda went "big" here, so it's harder to validate that it won't completely break everything.  Have you tried beta testing this patch in your project(s) to see if it works well for you?  I know I'd appreciate that!

karlhorky:
Ok, let me know if I can help in any other way to resolve the problems.

As for beta testing, we're going to build something off of it next week. I will post feedback here if I have any!

drewpc:
Lots of great discussion on here.  In an effort to shrink the complexity of all this, I will close this PR and break it up into three separate PRs: replace `webpack-dev-server` with `webpack-dev-middleware`, server-side HMR, and then multiple webpack configs.

jhnns:
> replace webpack-dev-server with webpack-dev-middleware

Is everyone convinced in doing that? As I already said, I would not replace the webpack-dev-server. It already takes care of serving the client bundle **and** HMR out-of-the-box. And by using the proxy feature, the react-server user does not need to know that there is actually a webpack-dev-server in place. It's an implementation detail.

---

As I already mentioned in the chat: Your PR is solving some issues and should be merged/published if it's not degrading the DX with long build times too much.

drewpc:
@jhnns Okay--I give in.  I'll refactor that PR so that it uses `webpack-dev-server` and still uses a single web server.

jhnns:
@drewpc I just noticed that the `react-server-cli` is still calling `babel-core/register` inside `cli.js`. Is this still necessary? The babel-loader has already compiled everything.

The only problem I see is when calling a custom webpack config, it won't be compiled with babel.

If the cli and the application server were two different processes, you could use `babel-core/register` just inside the cli without affecting the (already compiled) application server.

karlhorky:
@drewpc we've started building a large-scale distributed codebase on the changes in this pull request. It's working so far, and I'll post any further updates once we have any. We will open a few pull requests to `react-server` to fix some nits that we've run into.

I noticed that this pull request is being split up to ease reviewing and discussion. Are all the changes from this PR contained in #807, #808 and #809?

drewpc:
@karlhorky Awesome!  Let me know if there are any issues and I can help work through them.  I closed this PR because there were some changes that I suggested which may not be the right direction for where we want to take `react-server`.  The end result of server-side Webpack _is_ the right direction, but this may require some refactoring.  #809 has been integrated into the master branch as of yesterday.  #808 is still being discussed.  The rest of the changes that are not included in #808 and #809 are still in this branch.  Specifically, the rely on the changes in #808, so it would be very hard to completely separate the two.

My recommendation would be to continue using this branch until we decide on a new architecture approach (should be a few days/weeks).  Then I'll refactor this branch to use the new architecture.  Does that help?

karlhorky:
Great, that sounds good. I'll continue working with this branch and subscribe to #808 for any followup PRs for the server-side webpack. Thanks!

PR #803, opened 2017-01-04
Adding bottleneck vanilla test sites

krisxw:
Adding some vanilla test sites in the react-server-test-pages package. These are very straightforward test sites, where the elements site loads 99,999 random elements from a set of predetermined elements, and the data requests site makes 999,999 data requests. Finally, the middleware site has a chain of 100 (identical) middleware that performs a bunch of no-ops on each. That site is currently blank, but there is work being done when it is accessed.

Some performance metrics are produced by default if one just looks in the browser console. Ideally we will integrate these with automated testing suites in the near future. I'm planning on sending out another PR for some other simple test sites in the near future, but wanted to get the ball rolling on these changes sooner rather than later.

### DIFF
--- a/packages/react-server-test-pages/pages/bottleneck/dataRequests.js
+++ b/packages/react-server-test-pages/pages/bottleneck/dataRequests.js
@@ -1,0 +1,35 @@
+import {ReactServerAgent, RootContainer, RootElement} from "react-server";
+import "./colors/red.scss";
+import "./colors/green.scss";
+
+const elements = [];
+/**
+* This page is a smoke test to determine whether or not the number of data requests in
+* a page is a performance bottleneck for react-server. It performs a thousand data
+* requests before returning a simple message. Metrics are created in the browser's console
+* related to performance metrics (see react-server.core.ClientController).
+*/
+export default class DataRequestsPage {
+
+	handleRoute() {
+		//Reset elements, then perform one thousand local data requests before returning
+		elements.length = 0;
+		for (var i = 1; i <= 1000; i++) {
+			let current = i;
+			let promise = ReactServerAgent.get('/data/delay')
+				.then(response => response.body);
+			elements.push(<RootElement when={promise}><div>Data request {current} complete.</div></RootElement>);
+		}
+		return { code: 200 };
+	}
+
+	getElements() {
+		return [
+			<div className="red-thing">Data requests starting...</div>,
+			<RootContainer>
+				{elements}
+			</RootContainer>,
+			<div className="green-thing">Content should be above me.</div>,
+		];
+	}
+}
--- a/packages/react-server-test-pages/pages/bottleneck/elements.js
+++ b/packages/react-server-test-pages/pages/bottleneck/elements.js
@@ -1,0 +1,39 @@
+import "./colors/red.scss";
+import "./colors/blue.scss";
+import "./colors/purple.scss";
+import "./colors/green.scss";
+import "./colors/orange.scss";
+import "./colors/indigo.scss";
+import "./colors/yellow.scss";
+
+const RedThing = () => <div className="red-thing">This should be red</div>;
+const BlueThing = () => <div className="blue-thing">This should be blue</div>;
+const PurpleThing = () => <div className="purple-thing">This should be purple</div>;
+const GreenThing = () => <div className="green-thing">This should be green</div>;
+const OrangeThing = () => <div className="orange-thing">This should be orange</div>;
+const IndigoThing = () => <div className="indigo-thing">This should be indigo</div>;
+const YellowThing = () => <div className="yellow-thing">This should be yellow</div>;
+const ColorWheel = [RedThing(), BlueThing(), PurpleThing(), GreenThing(),
+	OrangeThing(), IndigoThing(), YellowThing()];
+const ColorSize = ColorWheel.length;
+
+/**
+* This page is a smoke test to determine whether or not the number of elements in
+* a page is a performance bottleneck for react-server. It returns a huge assortment of
+* randomly generated color-coded elements. Metrics are created in the browser's console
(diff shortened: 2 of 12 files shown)


packages/react-server-test-pages/pages/bottleneck/dataRequests.js
```diff
@@ -0,0 +1,12 @@
...
+export default class DataRequestsPage {
+
+	getElements() {
+		//Perform 1 million data requests before returning
+		for (var i = 0; i < 999999; i++) {
+			ReactServerAgent.get('/data/delay');
+		}
+		return (<div>Data requests complete!</div>);
+	}
+}
```
roblg: Technically I think we'd want the data requests to be kicked off in `handleRoute`. Practically, I'm not sure if that will make a difference.

packages/react-server-test-pages/pages/bottleneck/elements.js
```diff
@@ -0,0 +1,33 @@
...
+const ColorWheel = [RedThing(), BlueThing(), PurpleThing(), GreenThing(),
+	OrangeThing(), IndigoThing(), YellowThing()];
+
+export default class ElementsPage {
+
+	getElements() {
+
+		const colorThings = [];
+		for (var i = 0; i < 99999; i++) {
+			let selection = Math.floor(Math.random() * 7);
```
roblg:
```javascript
let selection = Math.floor(Math.random() * ColorWheel.length)
```
?

packages/react-server-test-pages/pages/bottleneck/elements.js
```diff
@@ -0,0 +1,33 @@
...
+			colorThings.push(ColorWheel[selection]);
+		};
+
+		return (<div>
+			{colorThings}
+			</div>);
+	}
```
roblg: Do we want to return a single element here with 100k children? Or 100k elements?

packages/react-server-test-pages/pages/bottleneck/elements.js
krisxw: I can comment this if it's confusing to read. Basically, I wanted a site that served something a bit more interesting than a bunch of monochrome elements, so it serves a random assortment of various elements from the colors of the rainbow. A bit unnecessary maybe, but it was very little effort.

packages/react-server-test-pages/pages/bottleneck/elements.js
krisxw: I didn't appreciate the difference. I'm wanting 100k elements, so I'll adjust this accordingly.

packages/react-server-test-pages/pages/bottleneck/middleware.js
```diff
@@ -0,0 +1,32 @@
...
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware,
+			JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware, JsonResponseMiddleware];
```
gigabo:
```javascript
_.range(80).map(() => JsonResponseMiddleware);
```

;)

packages/react-server-test-pages/pages/bottleneck/elements.js
gigabo: I think @roblg might be suggesting a replacement for the hard-coded 7 here.

I'll add that this can be a `const`. 👹

packages/react-server-test-pages/pages/bottleneck/dataRequests.js
gigabo: Yeah, these should get kicked off in `handleRoute`, and then we should make root elements wait for them.

Ideally we could style the root elements so they all fit above the fold and affect the WPT speed index.

packages/react-server-test-pages/pages/bottleneck/elements.js
```diff
@@ -15,19 +15,25 @@ const IndigoThing = () => <div className="indigo-thing">This should be indigo</d
...
+* related to performance metrics (see react-server.core.ClientController).
+*/
 export default class ElementsPage {
 
 	getElements() {
 
 		const colorThings = [];
-		for (var i = 0; i < 99999; i++) {
-			let selection = Math.floor(Math.random() * 7);
+		for (var i = 0; i < 10000; i++) {
```
krisxw: After changing this behavior such that I was correctly returning a large number of elements (not one element with a large number of children), I found myself running out of heap space in javascript, so this number came down intentionally.

packages/react-server-test-pages/pages/bottleneck/dataRequests.js
krisxw: I did not incorporate this feedback into my most recent commit. I'll address this in the next commit.

drewpc:
Looks good, thanks!

PR #804, opened 2017-01-06
Add CLI argument for configuration path

karlhorky:
**Disclaimer** This pull request was closed because of:
- [alternative ways to include configuration files](https://github.com/redfin/react-server/pull/804#discussion_r98417797)
- there was no concrete need for this at the time of writing
- less configuration is better

Hi, firstly I would like to thank the contributors for all their hard work on this important project! It pushes the state of the art for universal rendering to new limits and introduces ideas such as streaming that hopefully make their way back into React itself at some point.

During integration of `react-server` in a project, we ran into an issue with the configuration files. Certain projects such as those set up with `create-react-app` (and ours) decide to keep their configuration files such as `.babelrc` and `.eslintrc` in a non-project-root location. We came up with the idea of configuring the path to these files via a command line argument, which is what resulted in this proof of concept pull request.

Since then our requirements have changed and we will keep the files in the root after all. Because of this and because I wasn't sure what response this was going to get, I didn't take this further than a first proof of concept. Feel free to reject this, since the best use case that I can present is that it improves configurability for projects structured this way.

However, if this configuration option is of interest to you, I will add the following to this pull request:

- [x] Documentation
- [x] Tests
- [x] #828 merged

Thanks!

### DIFF
--- a/packages/react-server-cli/src/cli.js
+++ b/packages/react-server-cli/src/cli.js
@@ -3,7 +3,10 @@ const path = require("path");
 const cli = require(".");
 
 cli.parseCliArgs().then(args => {
-	const config = args.configPath ? JSON.parse(fs.readFileSync(path.join(args.configPath, ".babelrc"))) : {};
+	const babelrcPath = path.join(args.configPath, ".babelrc");
+	const config = args.configPath && fs.existsSync(babelrcPath) ?
+		JSON.parse(fs.readFileSync(babelrcPath)) :
+		{};
 	require("babel-core/register")(config);
 	cli.run(args);
 }).catch(console.error);
--- a/packages/react-server-cli/src/cli.js
+++ b/packages/react-server-cli/src/cli.js
@@ -3,7 +3,7 @@ const path = require("path");
 const cli = require(".");
 
 cli.parseCliArgs().then(args => {
-  const config = args.configPath ? JSON.parse(fs.readFileSync(path.join(args.configPath, ".babelrc"))) : {};
-  require("babel-core/register")(config);
-  cli.run(args);
+	const config = args.configPath ? JSON.parse(fs.readFileSync(path.join(args.configPath, ".babelrc"))) : {};
+	require("babel-core/register")(config);
+	cli.run(args);
 }).catch(console.error);
--- a/packages/react-server-cli/src/cli.js
+++ b/packages/react-server-cli/src/cli.js
@@ -3,8 +3,9 @@ const path = require("path");
 const cli = require(".");
 
 cli.parseCliArgs().then(args => {
-	const babelrcPath = path.join(args.configPath, ".babelrc");
-	const config = args.configPath && fs.existsSync(babelrcPath) ?
+	const {configPath = ''} = args;
+	const babelrcPath = path.join(configPath, ".babelrc");
+	const config = configPath && fs.existsSync(babelrcPath) ?
 		JSON.parse(fs.readFileSync(babelrcPath)) :
 		{};
 	require("babel-core/register")(config);
--- a/packages/react-server-cli/src/cli.js
+++ b/packages/react-server-cli/src/cli.js
@@ -1,4 +1,9 @@
-require("babel-core/register");
-
+const fs = require("fs");
+const path = require("path");
 const cli = require(".");
-cli.parseCliArgs().then(cli.run).catch(console.error);
+
+cli.parseCliArgs().then(args => {
+  const config = args.configPath ? JSON.parse(fs.readFileSync(path.join(args.configPath, ".babelrc"))) : {};
+  require("babel-core/register")(config);
+  cli.run(args);
+}).catch(console.error);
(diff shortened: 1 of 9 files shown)


gigabo:
Yikes, config files outside the package root sounds like a recipe for pain.

This option seems worthwhile, though, if it supports a workflow that's awkward without it.  Would be nice to have a note about it in [the CLI docs](https://github.com/redfin/react-server/blob/master/docs/guides/react-server-cli.md) with some explanation about why you might want to do this to yourself. :feelsgood: 

Thanks for contributing @karlhorky!

markthethomas:
FWIW, this was something I ran into when initially bringing RS into my project. Had a primary express app server that sent client code down and wanted this to replace the additional server I had set up for the public-facing server (it basically served user sites) 👍

karlhorky:
> Would be nice to have a note about it in the CLI docs with some explanation about why you might want to do this to yourself.

I've added an option to the CLI docs.

---

I also wanted to add a test for this, but maybe it would be good to get some guidance on where to put this. I took a look at [`packages/react-server-cli/test`](https://github.com/redfin/react-server/tree/4b0e43dffb6286a2d3f6b16e3832bdbbbd9e4202/packages/react-server-cli/test) but this folder appears to contain only tests for the parsing of the CLI arguments. Another location that I checked was [`packages/react-server-integration-tests/src/__tests__`](https://github.com/redfin/react-server/tree/4b0e43dffb6286a2d3f6b16e3832bdbbbd9e4202/packages/react-server-integration-tests/src/__tests__) - maybe some tests for this CLI argument would fit here?

gigabo:
> I've added an option to the CLI docs.

Thanks! 🍻 

> I also wanted to add a test for this

Hmm... the CLI tests are pretty thin (as you pointed out), but that _seems_ like the right place for this.

doug-wade:
I think the tests for this would belong in `packages/react-server-cli/test`.  There isn't a lot of precedent; I can lend a hand if you need some help setting up more testing.

karlhorky:
Just getting my head into the current techniques for writing tests for CLI tools (looking at `babel`, `next.js`, `jest`) and then I'll come up with a strategy that can work for more than just one test (since #828 also needs this).

karlhorky:
Based on the test setup added in #828, I've added a test for the `configPath` argument. Because of this, this PR is blocked on #828.

karlhorky:
Rebased on `master` and added a test. Ready for review.

cc @gigabo @doug-wade

packages/react-server-cli/src/cli.js
```diff
@@ -1,4 +1,13 @@
-require("babel-core/register");
-
+const fs = require("fs");
+const path = require("path");
 const cli = require(".");
-cli.parseCliArgs().then(cli.run).catch(console.error);
+
+cli.parseCliArgs().then(args => {
+	const {configPath = ''} = args;
```
doug-wade: Destructuring with defaults doesn't seem to be supported in node 4.0.0 (at least in my little experiment in the repl just now), so I don't think you'll be able to use it before you require babel.

packages/react-server-cli/src/compileClient.js
```diff
@@ -159,9 +160,13 @@ function statsToManifest(stats) {
 	};
 }
 
-function packageCodeForBrowser(entrypoints, outputDir, outputUrl, hot, minify, longTermCaching, stats) {
+function packageCodeForBrowser(entrypoints, outputDir, outputUrl, hot, minify, longTermCaching, stats, configPath) {
 	const NonCachingExtractTextLoader = path.join(__dirname, "./NonCachingExtractTextLoader");
 	const extractTextLoader = require.resolve(NonCachingExtractTextLoader) + "?{remove:true}!css-loader";
+	const babelrcPath = path.join(configPath || '', ".babelrc");
+	const babelConfig = configPath && fs.existsSync(babelrcPath) ?
```
doug-wade: hmm... babel users can also put their [configuration in package.json](https://babeljs.io/docs/usage/babelrc/#use-via-package-json).  I _think_ we currently support that; does this change break that?

packages/react-server-cli/src/findOptionsInFiles.js
```diff
@@ -13,12 +13,15 @@ const PACKAGE_JSON = "package.json";
 // the package.json. if it finds neither, it goes up a directory and looks again.
 // it returns the contents of the first config file found; it never merges multiple
 // configurations.
-export default (dir = process.cwd()) => {
+export default (args = process.argv) => {
+	let dir = process.cwd();
```
doug-wade: Why are you removing the ability to override dir?

packages/react-server-cli/src/parseCliArgs.js
```diff
@@ -107,6 +107,11 @@ export default (args = process.argv) => {
 			default: false,
 			type: "boolean",
 		})
+		.option("config-path", {
+			describe: "Path to configuration files such as .reactserverrc, .eslintrc and .babelrc.",
```
doug-wade: "Path to a directory that contains configuration files..." (right?)

packages/react-server-cli/src/findOptionsInFiles.js
karlhorky: I removed it because it wasn't used in any of the function calls. I can add it back though!

packages/react-server-cli/src/compileClient.js
karlhorky: Hm, allowing Babel, ESLint, etc config in `package.json` is a really important feature here, and feels like it is getting into deep complexity.

I think for now I'll make a note in the description above and close this pull request, because:
- less configuration is better
- there's not a concrete use case for this right now
- this may create weird edge case bugs

karlhorky:
Closing as a result of reasons mentioned in https://github.com/redfin/react-server/pull/804#discussion_r98417797

PR #806, opened 2017-01-06
Test page for page forwarding

lidawang:
Specifically, to make sure that data can be passed down from the forwarding page to the forwarded-to page via the data cache, and that we can do code splitting between the forwarded-to pages. (Requirements for a particular use case I'm anticipating.)

### DIFF
--- a/packages/react-server-test-pages/pages/navigation/forward.js
+++ b/packages/react-server-test-pages/pages/navigation/forward.js
@@ -17,7 +17,7 @@ export default class ForwardPage {
 			//then depending on said data, forward to one of two pages, and pass along the data we pre-fetched
 			if (res.body % 2 === 0) {
 				if (typeof window !== 'undefined') { //would be nice if this is `process.env.isServer`
-					require.ensure(["./forwardEven"], () => {
+					return require.ensure(["./forwardEven"], () => {
 						return {
 							page: require("./forwardEven").default,
 						};
@@ -29,7 +29,7 @@ export default class ForwardPage {
 				}
 			} else {
 				if (typeof window !== 'undefined') { //would be nice if this is `process.env.isServer`
-					require.ensure(["./forwardOdd"], () => {
+					return require.ensure(["./forwardOdd"], () => {
 						return {
 							page: require("./forwardOdd").default,
 						};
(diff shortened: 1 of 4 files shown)


packages/react-server-test-pages/pages/navigation/forward.js
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		//fetch some data
+		this.data = ReactServerAgent.get('/data/delay?ms=1000&val='+params);
+
+		return this.data.then((res) => {
+			//then depending on said data, forward to one of two pages, and pass along the data we pre-fetched
+			if (res.body % 2 === 0) {
+				if (typeof window !== 'undefined') { //would be nice if this is `process.env.isServer`
+					require.ensure(["./forwardEven"], () => {
```
doug-wade: I can't quite tell because babelification messes with line numbers, but it _looks_ like the problem is that you're not returning the `require.ensure` result.

lidawang: Oh oops - thanks for the catch!

roblg:
LGTM

packages/react-server-test-pages/pages/navigation/forward.js
```diff
@@ -0,0 +1,50 @@
...
+		} else {
+			params = 0;
+		}
+
+		//fetch some data
+		this.data = ReactServerAgent.get('/data/delay?ms=1000&val='+params);
+
+		return this.data.then((res) => {
+			//then depending on said data, forward to one of two pages, and pass along the data we pre-fetched
+			if (res.body % 2 === 0) {
```
drewpc: Can you refactor this section a bit to limit code reuse?  Something like this should work:

```
const pageName = (res.body % 2 === 0) ? "./forwardEven" : "./forwardOdd";
if (typeof window !== 'undefined') { //would be nice if this is `process.env.isServer`
	return require.ensure([pageName], () => {
		return {
			page: require(pageName).default,
		};
	});
} else {
	return {
		page: require(pageName).default,
	};
}
```

packages/react-server-test-pages/pages/navigation/forward.js
```diff
@@ -0,0 +1,50 @@
...
+				if (typeof window !== 'undefined') { //would be nice if this is `process.env.isServer`
```
drewpc: If you change this comment to be a `// TODO:`, we can track it better.  Eventually when we get server side Webpack up and running, it'll include a fix for this.

lidawang:
Woops I went and neglected this PR, but looks like @drewpc got a handle on things, thank you very much for fixing + merging!

Issue #807, opened 2017-01-07
Consolidate HTML + Javascript Servers


drewpc:
This issue serves to simplify the setup of `react-server` and bring all HTTP(S) requests under the same domain.  Currently, if a person configures `react-server` with the following settings:

```
host: 0.0.0.0,
port: 3000,
jsPort: 3001,
hot: true,
```

and opens a browser window pointing to `http://www.myhost.com:3000`, all requests for Webpack files will go to http://0.0.0.0:3001.  Those requests are considered part of a different domain according to the [CORS same-origin policy](https://en.wikipedia.org/wiki/Same-origin_policy#Origin_determination_rules).

Additionally, the `jsPort` requirement is another port that needs to be configured when developing.  This isn't usually a problem when developing on the same machine, but under a Vagrant environment, it can be problematic and unexpected.

Consolidating the HTML + JS servers into a single HTTP/HTTPS server will fix all of these issues and simplify the overall project.  Lastly, it is a stepping stone to further enhancements such as server-side HMR + Webpack configs.

aickin:
> Those requests are considered part of a different domain according to the CORS same-origin policy.

For what it's worth (which, frankly, is maybe not much?), I considered this to be an advantage, because I've often run into problems when my dev experience serves static files and dynamic HTML from the same server. In every production setup I've used, static files are placed on a CDN, so they are subject to CORS restrictions. But when you serve static and dynamic from the same server, the dev experience isn't subject to CORS, and you end up with confusing CORS bugs in prod that are hard to reproduce in the (non-CORS) dev setup. My solution has generally been to simulate a CDN in dev so that both environments are subject to CORS. Thoughts?

drewpc:
@aickin Fair point.  I imagine that this was an unintended side effect instead of an architecture decision, don't you?  I agree that simulating CORS issues prior to moving to production is important.  Do you think that many people realize that `myhost.com:3000` and `myhost.com:3001` are considered different origins?  I didn't until I looked it up to determine what was causing certain issues.  I'd say its not obvious and therefore wouldn't be expected when someone sets up a simple HMR environment.  Just my $0.02.

aickin:
> I imagine that this was an unintended side effect instead of an architecture decision, don't you?

I mean, my memory is notoriously faulty, but I'm pretty sure I set up the CLI with separate JS and HTML servers in order to make sure to simulate CORS issues in dev. Whether that's a good idea or not remains an open question. ;) 

[...]

PR #808, opened 2017-01-07
Consolidate HTML + JS Servers

drewpc:
This PR addresses #807 with the *least* controversial set of changes.  See PR #791 for more discussion on this.  Essentially, this PR does the following:

- Consolidates all Express servers into a single server (removing the requirement for `jsPort`)
- If `hot === true`, `webpack-dev-server` is used as to create the HTTP/HTTPS server and Express application.  `react-server` piggy backs off of the Express application by injecting itself into the middleware chain after `webpack-dev-middleware` is injected.  No proxying is required.
- If `hot === false`, `react-server` creates the HTTP/HTTPS server and Express application on its own and manually creates and runs a Webpack compiler for the client side code.

In my opinion, this is not the cleanest path for `react-server`, but it is the *least* controversial set of changes to the overall infrastructure while still reducing complexity at runtime.

Another method of solving this problem is to stop using [`webpack-dev-server`](https://github.com/webpack/webpack-dev-server) and attach [`webpack-dev-middleware`](https://github.com/webpack/webpack-dev-middleware) to the Express application that `react-server` creates.  This will more easily allow people to use alternatives to Express in the future and keep `react-server` operating the same in hot/not-hot modes.  It does, however, move the client-side HMR mechanism away from something that the Webpack people maintain and over to [`webpack-hot-middleware`](https://github.com/glenjamin/webpack-hot-middleware).  This project supports multiple compilers (which will be useful later), Webpack 2, and seems to be well maintained.

Checkout [this patch](https://github.com/redfin/react-server/compare/master...SoprisApps:patch-use-wepack-dev-middleware?diff=unified&name=patch-use-wepack-dev-middleware) to see what this code would look like.

As always, feedback is appreciated.


packages/react-server-cli/src/commands/start.js
```diff
@@ -13,239 +13,172 @@ import logProductionWarnings from "../logProductionWarnings";
...
+
+		expressServer = express();
+		expressServer.use('/', compression(), express.static(`__clientTemp/build`, {
+			maxage: longTermCaching ? '365d' : '0s',
+		}));
+		httpServer = httpsOptions ? https.createServer(httpsOptions, expressServer) : http.createServer(expressServer);
+	}
+
+	let middlewareSetup = (server, rsMiddleware) => {
+		server.use(compression());
```
jhnns: I don't know if that's a problem but it looks like the `compression()` middleware is used two times in non-hot mode.

drewpc: thanks!  I fixed that.

drewpc:
Closing this PR, pending further architecture discussion.

PR #809, opened 2017-01-07
Server side hot reload

drewpc:
This patch enables server-side hot module reload on the existing master branch.  Note: this does *not* include any Webpack changes nor does it include any changes to ExpressJS unless required for HMR to work.  It simply reloads the existing changes when Webpack successfully recompiles the client code.  This is a refactor of the code from #791 to only include server side HMR changes.  Discussed in #773.

UPDATE: this branch was rebased and changes from #808 were removed.  It is now a standalone patch.


gigabo:
This is a cool solution.  Thanks for working it out @drewpc!

LGTM 👍

packages/react-server-cli/src/commands/start.js
```diff
@@ -80,11 +82,23 @@ const startHtmlServer = (serverRoutes, port, bindIp, httpsOptions, customMiddlew
...
+					server.use(cookieParser());
+
+					// sets the namespace that data will be exposed into client-side
+					// TODO: express-state doesn't do much for us until we're using a templating library
+					server.set('state namespace', '__reactServerState');
+
+					server.use((req, res, next) => {
+						reactServer.middleware(req, res, next, require(serverRoutes));
+					});
+				};
```
drewpc: Fixed.  I don't think that eslint is working properly.  `asini run lint` blows up and there's no script in the root level package.json to run lint.  I'll look into it.

gigabo: 🙀

drewpc: (at least in my setup it's blowing up)

gigabo: Hmmm... we just run `asini run test` in CI.  So each package is responsible for making sure that linting is part of its test script.  It _looks_ like the CLI's package.json should be running tests as part of the `gulp test` phase of its test script.

Wonder where that's breaking... 🤔

drewpc: This is what I get when I run `asini run lint` in my setup:

```
asini run lint

> react-server-module-tagger@0.4.10 lint /Users/drewpc/Documents/Juno Tech/git-repos/react-server/packages/react-server-module-tagger
> eslint *.js


NpmUtilities.execInDir        ("run lint", [], "./packages/react-server-module-tagger")

> react-server-module-tagger@0.4.10 lint /Users/drewpc/Documents/Juno Tech/git-repos/react-server/packages/react-server-module-tagger
> eslint *.js


NpmUtilities.runScriptInDir   ("lint", [], "./packages/react-server-module-tagger")

> react-server-module-tagger@0.4.10 lint /Users/drewpc/Documents/Juno Tech/git-repos/react-server/packages/react-server-module-tagger
> eslint *.js


Errored while running npm script 'lint' in 'react-server-module-tagger'
Error: Command failed: npm run lint 
sh: eslint: command not found
```

drewpc: When I fix that issue (and the resulting things eslint complains about), I get this:

```
asini run lint
Successfully ran npm script 'lint' in packages:
- generator-react-server
- react-server-module-tagger
```

drewpc: so...`lint` doesn't exist anywhere else.

gigabo: Oh, nice catch on the missing dep.  Yeah, some modules (like `react-server-cli`) use the eslint gulp plugin instead of the executable.  It's all kind of a weird mishmash right now. 😖

drewpc: I'll clean it up and submit a PR.  I already have the previous issue fixed.

drewpc: This is the offender: https://github.com/redfin/react-server/blame/master/packages/react-server-cli/gulpfile.babel.js#L7  Need to use globs...

drewpc: PR #818

drewpc:
@jhnns I did some basic testing on memory leaks with regard to deleting require.cache entries (modifying [this code](https://www.alexkras.com/simple-guide-to-finding-a-javascript-memory-leak-in-node-js/))  and came to the following conclusion: my server side patch HMR solution doesn't do anything good or bad to resolve those memory leaks.  In other words, if memory leaks exist in the code being required `delete require.cache[<module>]` doesn't "fix" that.  Your solution of restarting a separate process would fix the symptom, but not the problem: memory leaks still exist.

The topic of memory leaks would be a great documentation item: patterns to avoid, ways to detect memory leaks, etc.  The page cycle of `react-server` would be the most vulnerable, especially since it is called for each HTTP request.  I'll have to look into this, but I see it as a different problem than server side HMR.

roblg:
Added `breaking-change` to this, due the moving of the express middleware and express-state setup out of `renderMiddleware`. (Which is a change that has been overdue. Thanks!)

jhnns:
> my server side patch HMR solution doesn't do anything good or bad to resolve those memory leaks. In other words, if memory leaks exist in the code being required delete require.cache[<module>] doesn't "fix" that.

Of course, you're right. These memory leaks exists in the code because you don't write tear-down code if you don't need it. But usually you don't just delete `require.cache`, so it's not an actual memory leak in production because the module is not replaced in production. It just becomes a (potential) memory leak when the cache is deleted. 

But besides the memory leak I rather worry about the unknown application state after modules have been swapped out without actually tearing down everything. Your application may behave strangely in development after a while (because of stale connections, etc.). I'd rather prefer to restart the server cleanly than having to deal with an unknown application state. That's why [create-react-app is also using HMR just for CSS](https://github.com/facebookincubator/create-react-app/blob/15feb02e9cae7f737f44bd6a3ddf2103961fa362/packages/react-scripts/config/webpack.config.dev.js#L219), because CSS can be replaced without introducing side-effects.

I've explained some details of HMR in [this SO answer](http://stackoverflow.com/questions/37016683/what-aspect-of-hot-module-replacement-is-this-article-for/37022884#37022884).

PR #810, opened 2017-01-08
Fix import inconsistencies in page-writing guide

markthethomas:
Came across this when I was reading through the docs for React Server and integrating it into my project. Copied over the component just to get a sense of it in my text editor and the linter immediately caught `import x from ('y');`. Guessing this was just leftovers from converting things to import syntax, or at least that's what's happened to me before :) For the sake of consistency I changed the rest of the exports to be in line, not sure if that's not preferred for some reason I can't think of. Absolutely love the lib, phenomenal work everyone! 👍


roblg:
Thanks @markthethomas!

PR #812, opened 2017-01-09
Helmet as default express middleware in react-server-cli

feychou:
Addresses these website issues:

https://github.com/redfin/react-server/issues/640
https://github.com/redfin/react-server/issues/641
https://github.com/redfin/react-server/issues/639

It relies on [helmet's default configuration](https://www.npmjs.com/package/helmet#how-it-works); its overall purpose is to make react-server apps more secure by default.

PR #817, opened 2017-01-09
Documentation update: bullet list for express middlewares

feychou:
And links to npm packages as suggested [here](https://github.com/redfin/react-server/pull/812#discussion_r95263075)


doug-wade:
@feychou you might wanna consider joining our slack channel so we can chat about changes and plans and stuff: https://slack.react-server.io/ (I only suggest it cuz I just went to slack you about why you closed this and then you weren't there :stuck_out_tongue:)

gigabo:
Yes, I was just doing the same thing. 😁 

This is awesome!

feychou:
@doug-wade @gigabo Sure! The reason why I closed this is that I noticed it listed old commits in my fork and I did not understand why since they have been merged, but now I suspect it is just because I merged the upstream master after all.

gigabo:
Thanks @feychou!

PR #818, opened 2017-01-10
Make ESLint Work

drewpc:
ESLint wasn't properly listing all files in `react-server-cli` because if a missing glob pattern.  Additional cleanup was to add a `lint` run script to each package.json file and fix a dependency issue with `react-server-module-tagger`


packages/react-server-cli/src/commands/start.js
```diff
@@ -13,53 +13,44 @@ import logProductionWarnings from "../logProductionWarnings";
...
-		jsUrl,
-		httpsOptions,
-		longTermCaching,
-		customMiddlewarePath,
-	} = options;
 
-	const {serverRoutes, compiler} = compileClient(options);
+// returns a method that can be used to stop the server. the returned method
+// returns a promise to indicate when the server is actually stopped.
+const serverToStopPromise = (server) => {
```
gigabo: Ugh, thanks for going through all this!

gigabo:
Thanks for tracking down the problem in `react-server-cli` and getting this all cleaned up @drewpc!

drewpc:
No problem!  It's all happy now:

```
$ asini run lint
Successfully ran npm script 'lint' in packages:
- babel-plugin-react-server
- generator-react-server
- react-server-cli
- react-server-core-middleware
- react-server-data-bundle-cache
- react-server-examples
- react-server-gulp-module-tagger
- react-server-module-tagger
- react-server-test-pages
- react-server-website
- react-server
```

roblg:
🚀!

Does this depend on #809 to be merged first? If not, let's merge!

drewpc:
@roblg Nope, no dependencies at all.  The only reason that `start.js` was modified is because ESLint complained about functions being used before they were defined in that file.

Issue #680, opened 2016-10-02
Resolve nsp exemption


doug-wade:
We've had to add an exemption to [nsp 118](https://nodesecurity.io/advisories/118).  We could upgrade to [gulp 4](https://github.com/gulpjs/gulp/milestone/1) now, or wait until the vulnerable version is flushed out of the dependency closure in gulp 3.  Either way, we will need to resolve the exemption.

```
┌───────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│               │ Regular Expression Denial of Service                                                                                                                                                                        │
├───────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Name          │ minimatch                                                                                                                                                                                                   │
├───────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Installed     │ 2.0.10                                                                                                                                                                                                      │
├───────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Vulnerable    │ <=3.0.1                                                                                                                                                                                                     │
├───────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Patched       │ >=3.0.2                                                                                                                                                                                                     │
├───────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Path          │ react-server-test-pages@0.4.7 > gulp@3.9.1 > vinyl-fs@0.3.14 > glob-stream@3.1.18 > minimatch@2.0.10                                                                                                        │
├───────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ More Info     │ https://nodesecurity.io/advisories/118                                                                                                                                                                      │
└───────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
[...]

PR #819, opened 2017-01-11
Remove outdated nsprcs

doug-wade:
The fix made its way through the npm ecosystem


doug-wade:
Uh... I did a bad job titling #680 -- I named it after how I thought we'd fix the issue, rather than the actual issue.  This closes #680, which I've updated to better describe the issue.

PR #825, opened 2017-01-11
Use express custom middleware API instead of a page with content-type JSON in bike-share example

jhnns:
The current bike-share example uses a page to implement the JSON api. I don't know if that's really the recommended way but this will expose the API implementation to the client because it will be included into the client bundle.

To me, the current example is a potential security leak because I assume that most developers are not aware that this code will be shipped to the client. For instance, imagine you import a DB password or key inside the API because you need to read something from the DB, this would be included into the client bundle.

The refactored example uses a custom express middleware to implement the actual API. This way we can be sure that this code will not be shipped to the client.


drewpc:
I think this should be a different example rather than a modification of this example.  This example is great because it specifically highlights how to create an isomorphic call to an API.

doug-wade:
First off, thank you for this contribution!  I :heart: :blue_heart: :green_heart: having a custom express middleware example; its something that's been on my wishlist for a long time, and I very much appreciate you taking the time to write htis @jhnns.  

I'm not sure what the build failure was, so I kicked the build in case its transient.

As to @drewpc's comment about making this a separate example, I agree that we should keep the isomorphic api call as part of the bike share example, but I don't know that it has to be a separate example.    I think we've sort of let our vision drift on `react-server-examples`; some of them demonstrate a single idea succinctly (`redux-basic`, for instance) and some of them are more "this is what a realistic app might look like" (`bike-share` and `meteor-app`).  I don't think we should try to tighten our vision in this pr in particular, but I think we might want to revisit how the examples work and whether we want to better curate the examples.  

For this pr, my thought is that if you would like to include this custom express middleware in the bike share example specifically, change the isomorphic api call to call the express api instead of the remote api directly, instead of replacing the isomorphic api call with the express api.  If you would like, you could also spin up a separate example only for adding custom express middleware, but that's above and beyond, and I don't think we should require that before merging.

jhnns:
> change the isomorphic api call to call the express api instead of the remote api directly, instead of replacing the isomorphic api call with the express api

Which isomorphic api call do you mean? `api/network.js` is never executed on the client.

jhnns:
@doug-wade @drewpc I'm willing to improve this, but I don't know in what direction.

In particular, I don't understand what call should be changed:

> For this pr, my thought is that if you would like to include this custom express middleware in the bike share example specifically, change the isomorphic api call to call the express api instead of the remote api directly, instead of replacing the isomorphic api call with the express api

But maybe we need to separate two things here:

1. It would be useful to have an express middleware example
2. I want to change the current bikeshare example because it imposes a possible security problem

If you agree with me on 2., we could merge this PR and we would also have an express middleware example. If you do not agree with me, I can create a separate PR with just the express middleware. But in that case I would like to understand why you want to keep the current example.

jhnns:
One source of confusion to me is the isormorphic api call @drewpc has mentioned. Could you show me which call you mean?

doug-wade:
hey @jhnns reading back over what I wrote earlier, I'm not sure whether it actually makes any sense.  Ima page @gigabo and @roblg in case they have a moment to look over this in the meantime, but I'm going to need to tinker a little bit with the existing example to make sure I understand what its doing.

drewpc:
@jhnns my understanding was that the calls to `/api/network` were made through the PageUtil and thus were isomorphic.  After looking into it further, it seems that's not the case, so my initial comment about that was incorrect.

karlhorky:
Ah cool, so does that mean that we can move forward with this PR as it is? Or should it be changed still?

packages/react-server-examples/bike-share/api/index.js
```diff
@@ -0,0 +1,13 @@
+import { getNetworks, getNetwork } from "./networks";
+
+function setJsonContentType(req, res, next) {
+	res.type('json');
+	next();
+}
+
+export default function registerCustomMiddlewares(server, registerRenderMiddleware) {
+	server.use('/api', setJsonContentType);
```
roblg: Was going to suggest putting a link here to the section on using custom express middleware, but then realized that you can't link to the headers. Filed #862.

jhnns:
Yay! 😁

gigabo:
Thanks @jhnns!

PR #490, opened 2016-08-02
Cli commands

gigabo:
Now that the CLI can be [installed globally](https://github.com/redfin/react-server/pull/484) it can be used to bootstrap a React Server site.

This is totally new functionality that's only tangentially related to the previous "start server" behavior.  So... I've added "commands".

```
$ react-server
Usage: react-server <command> [options]

Commands:
  start     Start the server
  compile   Compile static assets
  init      Initialize a React Server site
  add-page  Add a page to an existing site

Options:
  --routes-file        The routes file to load. Default is 'routes.js'.
  -p, --port           Port to start listening for react-server. Default is 3000.
...
```

The new quick start in the CLI doc looks like:

```bash
$ npm install -g react-server-cli
$ react-server init
$ react-server add-page '/' Homepage
$ react-server start
```

This is a pretty big change to the interface of `react-server-cli`.

@aickin - How does this look to you?

PR #828, opened 2017-01-11
Improve error messaging when requiring routes, add CLI tests

karlhorky:
If there is an error in your `routes.js` file (for instance, referencing an undeclared variable), `react-server-cli` will fail with a confusing error message:

```sh
TypeError: Cannot read property 'routes' of undefined
    at exports.default (~/projects/rs/node_modules/react-server-cli/target/buildWebpackConfigs.js:112:77)
    at buildWebpack (~/projects/rs/node_modules/react-server-cli/target/commands/start.js:302:54)
    at start (~/projects/rs/node_modules/react-server-cli/target/commands/start.js:109:20)
    at run (~/projects/rs/node_modules/react-server-cli/target/run.js:61:74)
    at process._tickCallback (internal/process/next_tick.js:103:7)
    at Module.runMain (module.js:609:11)
    at run (bootstrap_node.js:420:7)
    at startup (bootstrap_node.js:139:9)
    at bootstrap_node.js:535:3
```

I saw that @gigabo added this `try` / `catch` in redfin/react-server#490 so that the errors can be handled in the commands but I cannot see how this case is handled.

- [x] Tests


doug-wade:
Hi @karlhorky @kaytcat!  Thanks for pitching in!

I definitely agree that's a really unhelpful error, and that we need to improve the user experience.  Unfortunately, since our test coverage for the cli is pretty low, I can't quite be certain whether this change is good.  In particular, we added that when we made `react-server-cli` globally installable, and I suspect that this change will introduce a regression for when you're using `react-server-cli` _outside_ of an existing React Server project.  What happens when you're bootstrapping a new project with `react-server init`?  Won't the `require` throw an equally-confusing error?

I think we'll need to keep the try-catch, and inspect the error.  I think something like this should work

```
try {
  options.routes = require(options.routesPath);
} catch (e) {
  if (isParseError(e)) {
    throw e;
  }
  if (isNonDefaultRoutes(options.routes) && isFileNotFoundError(e)) {
    throw new Error("You've specified a custom routes file that does not exist.  Please add it");
  }
  // Pass. Commands need to check for routes themselves.
}
```

Tests would also be wonderful.

karlhorky:
Ah right, good catch. Looking through the commands, they all need the `options.routes` except for `init`. I've wrapped the `require` in a conditional to test for that.

I will take a look at how other popular command line libraries like `babel` test their CLI tools and do something similar in [`packages/react-server-cli/test`](https://github.com/redfin/react-server/tree/4b0e43dffb6286a2d3f6b16e3832bdbbbd9e4202/packages/react-server-cli/test).

karlhorky:
As for tests, I will take a look at how other popular command line libraries like `babel` test their CLI tools and do something similar in [`packages/react-server-cli/test`](https://github.com/redfin/react-server/tree/4b0e43dffb6286a2d3f6b16e3832bdbbbd9e4202/packages/react-server-cli/test).

doug-wade:
> Error: Cannot find module '~/projects/rs/packages/react-server-cli/src/routes.json'

That seems reasonable to me

karlhorky:
> why "fail hard" instead of output an appropriate, specific error and then end the process

Ah maybe that's just my error in bad wording, sorry. I do mean that what you're describing, I think. I'll change the title of the PR.

karlhorky:
Tests are coming along! It took a bit of work but I have AVA now running `react-server start` and checking some output.

```shell
➜  react-server-cli git:(add-cli-tests) ✗ npm run ava

> react-server-cli@0.5.1 ava ~/projects/rs/packages/react-server-cli
> ava

TAP version 13
2017-01-16T16:10:54.061Z - warning: [react-server-cli.src.logProductionWarnings] PRODUCTION WARNING: the following current settings are discouraged in production environments. (If you are developing, carry on!):
2017-01-16T16:10:54.064Z - warning: [react-server-cli.src.logProductionWarnings] -- Hot reload is enabled. To disable, set hot to false (--hot=false at the command-line) or set NODE_ENV=production.
2017-01-16T16:10:54.065Z - warning: [react-server-cli.src.logProductionWarnings] -- Minification is disabled. To enable, set minify to true (--minify at the command-line) or set NODE_ENV=production.
2017-01-16T16:10:54.066Z - warning: [react-server-cli.src.logProductionWarnings] -- Long-term caching is disabled. To enable, set longTermCaching to true (--long-term-caching at the command-line) or set NODE_ENV=production to turn on.
2017-01-16T16:10:54.066Z - info: [react-server-cli.src.logProductionWarnings] NODE_ENV is set to undefined
2017-01-16T16:10:54.172Z - notice: [react-server-cli.src.commands.start] Starting servers...
2017-01-16T16:10:54.172Z - info: [react-server-cli.src.commands.start] Starting hot reload JavaScript server...
2017-01-16T16:10:54.285Z - info: [react-server-cli.src.commands.start] Starting HTML server...
2017-01-16T16:10:54.400Z - info: [react-server-cli.src.commands.start] Started hot reload JavaScript server over HTTP on 0.0.0.0:3001
2017-01-16T16:10:54.401Z - info: [react-server-cli.src.commands.start] Started HTML server over HTTP on 0.0.0.0:3000
2017-01-16T16:10:54.401Z - notice: [react-server-cli.src.commands.start] Ready for requests on 0.0.0.0:3000.
# test
ok 1 - test

1..1
```

Going to add a few test cases including a `routes.js` file with an error in it.

drewpc:
@karlhorky Thanks for continuing to work on this.  Often the smallest changes take the longest time to write tests!  I wasn't so much as concerned with the wording as the question of: why output an error + stack trace instead of something like "The routesFile specified doesn't exist.  Please create a proper routesFile and re-run this command again." followed by `process.exit(1)`

karlhorky:
> why output an error + stack trace instead of something like "The routesFile specified doesn't exist. ... "

Good point, we may want to improve the error message if the error I posted above (https://github.com/redfin/react-server/pull/828#issuecomment-272211072) isn't a good enough Developer Experience. I'm good with changing it if that's what we want to do!

karlhorky:
> why output an error + stack trace instead of something like "The routesFile specified doesn't exist. ... "

Good point, in the case that it doesn't find the file we may want to improve the error message if the error I posted above (https://github.com/redfin/react-server/pull/828#issuecomment-272211072) isn't a good enough Developer Experience. I'm good with changing it if that's what we want to do!

In the case that there is really an error in `routes.js`, because it could be anything we probably shouldn't try to be smart about that one - just let it crash with the real error message.

drewpc:
@karlhorky Agreed.  So maybe a happy medium?  Something like this:

```
try {
    if (options.command !== 'init') {
        options.routes = require(options.routesPath);
    }
} catch (e) {
    console.error(chalk.red(`The routes file '${options.routesPath}' failed to load properly`));
    throw e;
}
```

karlhorky:
Ok, tests are done, I'll now improve the messaging guided by @drewpc 's suggestion and then we can get this one in!

karlhorky:
Ok, tests are done (inspired by Babel's testing system from babel/babel#1978), I'll now improve the messaging guided by @drewpc 's suggestion and then we can get this one in!

karlhorky:
I've improved the error message. Summoning @drewpc and @doug-wade for review!

gigabo:
Nice improvement @karlhorky.  Love the tests! 🤘

packages/react-server-cli/package.json
```diff
@@ -56,18 +57,28 @@
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-preset-react-server": "^0.4.10",
     "eslint-plugin-react": "^6.4.1",
+    "fs-readdir-recursive": "^1.0.0",
     "gulp": "^3.9.1",
     "gulp-babel": "^6.1.2",
     "gulp-eslint": "^3.0.1",
     "nsp": "^2.6.2",
+    "output-file-sync": "^1.1.2",
+    "react": "^15.4.2",
```
drewpc: Change to match the rest of the project: "react": "~0.14.2 || ^15.1.0"

packages/react-server-cli/package.json
```diff
@@ -56,18 +57,28 @@
...
+    "react-dom": "^15.4.2",
```
drewpc: Change to match the rest of the project: "react-dom": "~0.14.2 || ^15.1.0"

gigabo: Nice catch @drewpc.  Wonder how we could automate this check... 🤔

doug-wade:
This is awesome and I love it.  Thanks for sticking with it @karlhorky

packages/react-server-cli/package.json
karlhorky: Would peer dependencies help?

karlhorky:
@drewpc Fixed the versions, please re-review

Issue #830, opened 2017-01-14
Add an "isBrowser" variable


drewpc:
This way applications can test whether or not code is being executed client side or server side.

PR #831, opened 2017-01-14
Add isBrowser Variable

drewpc:
Solves #830.

### DIFF
--- a/docs/guides/understanding-rendering.md
+++ b/docs/guides/understanding-rendering.md
@@ -73,3 +73,18 @@ In the example above:
 - Elements are always rendered in-order.
 - The initial render is always the _same_ as the server render.
 - Elements may re-render if `listen` emitters fire after the initial render.
+
+## Render Context
+Even in isomorphic applications, it is sometimes important for a component to execute some action only on the server or
+in the browser.  There is a variable available at load time that time you can test to determine if the component is
+running in the browser or not:
+
+```javascript
+import { isBrowser } from "react-server";
+
+if (isBrowser) {
+    // do something when in the browser
+} else {
+    // do something different when on the server
+}
+```
--- a/packages/react-server-examples/bike-share/middleware/request-to-port.js
+++ b/packages/react-server-examples/bike-share/middleware/request-to-port.js
@@ -1,7 +1,7 @@
-import {ReactServerAgent} from 'react-server';
+import {ReactServerAgent, isBrowser} from 'react-server';
 export default class RequestToPortMiddleware {
 	handleRoute(next) {
-		if (typeof window === 'undefined') { //eslint-disable-line
+		if (!isBrowser) {
 			ReactServerAgent.plugRequest(req => {
 				req.urlPrefix('localhost:3000');
 			});
--- a/packages/react-server-integration-tests/src/__tests__/isBrowserSetting/IsBrowserSettingPage.js
+++ b/packages/react-server-integration-tests/src/__tests__/isBrowserSetting/IsBrowserSettingPage.js
@@ -1,10 +1,13 @@
+import React from 'react';
 import { isBrowser } from "react-server";
 
 export default class IsBrowserSettingPage {
 	getElements() {
-		return <div>
-			<span id="question">Is Browser:</span>
-			<span id="answer">{JSON.stringify(isBrowser)}</span>
-		</div>;
+		return (
+			<div>
+				<span id="question">Is Browser:</span>
+				<span id="answer">{JSON.stringify(isBrowser)}</span>
+			</div>
+		);
 	}
 }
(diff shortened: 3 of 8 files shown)


roblg:
Just so I'm clear on the intended users, this is targeted more towards clients of `react-server` than internal code, right? The one concern I have is that someone might try:

```
if (!isBrowser) {
    require("database-client");
}
```

which could blow up once webpack gets to it.

Internally, Redfin uses `process.env.IS_SERVER`, which is set to `true` server-side, and is replaced by webpacks' `DefinePlugin` client-side, resulting in a boolean literal, which allows webpack to ignore that code, and uglify to strip it out completely. I suppose it might be possible to do the same thing with a smart-ish webpack loader that understood `react-server` requires, and could identify the `isBrowser` variable?

drewpc:
@roblg Yes, definitely for clients of react-server...not internal code.  We can use `process.env.IS_SERVER`, no issues.  We discussed it on slack and my only recommendation for this way was to make it super obvious to new users how to use that.  We can put it in `process.env` and document that it's there. and why.

doug-wade:
@roblg note that accessing `process.env` while doing react server-side rendering is generally frowned upon for [performance reasons](https://github.com/facebook/react/issues/812), so its likely that the `isBrowser` flag will improve performance for clients, including Redfin.

roblg:
> We can use process.env.IS_SERVER, no issues... my only recommendation for this way was to make it super obvious to new users how to use that.

Agreed, that would be my primary concern as well, since it's not super discoverable as an env var

gigabo:
Yeah, this is cool.  I like that it's just exported from `react-server` and not some magic, but... we kind of _need_ the magic for conditional imports. 😬

roblg:
> accessing process.env while doing react server-side rendering is generally frowned upon for performance reasons,

@doug-wade yep. I was imagining that we'd replace it w/ webpack when we do that server-side. The key is that it's an easily identifiable global "constant" that doesn't require any context to replace.

drewpc:
Here's an interest tidbit regarding uglify removing dead code.  Copy/paste the following into the [Uglify Demo Page](http://lisperator.net/uglifyjs/#demo):

```
const a = false;
if (a == true) {
    console.log('this is dead code and should be removed');
} else {
    console.log('this is live code');
}
```

When you run the compressor, uglify will remove the dead code because `a` was declared as a `const`.  If you change it to `var` or `let`, uglify won't remove it.  Neat!  The problem is, we're babelifying the code down to ES2015 prior to running uglify, so we lose the `const` declarations....gah!

Here are the steps we take now:
1. In `react-server`, gulp performs the following task when building client and server libraries: `			.pipe(replace("SERVER_SIDE", serverSide ? "true" : "false"))`.  This replacement creates dead code, but it isn't removed by gulp, babel, or uglify at this stage.
2. Babel is run on the `react-server` code, effectively removing all of the ES6 features (like `const` variable declarations).
3.  The webpack config includes the uglify plugin, so client application code (including the `react-server` components) is uglified and dead code is removed only for client side.  Dead code remains in the server side code at this stage.

If `isBrowser` were declared as a `const` in the package, it would be transpiled to a `var` by babel and not removed by uglify.  If we could keep `isBrowser` as a `const` and let Webpack/Uglify handle transpiling and removing, then we'd be in business!

drewpc:
[babel's babili handles it properly](http://babeljs.io/repl/#?babili=true&evaluate=false&lineWrap=false&presets=react%2Cstage-2&code=var%20a%20%3D%20false%3B%0Aif%20(a%20%3D%3D%3D%20true)%20%7B%0A%20%20console.log('dead')%3B%0A%7D%20else%20%7B%0A%20%20console.log('live')%3B%0A%7D) using `var`, `const`, and `let`.

roblg:
> When you run the compressor, uglify will remove the dead code because a was declared as a const

Oh that's neat. Did not know uglify could do that.

It seems like we'd still have a problem though, since we're doing something like `const isBrowser = require("react-server").isBrowser`, and there's no way for a transform that's only looking at a single file at a time to know that the expression on the right is a constant (unless we replace it with a constant in a plugin). I also think we can't send `const` in the final webpack bundle, so we'd have to make sure whatever the final step is in the transform converts that `const` to `var`?

re: babili -- is babili itself actually doing the ES2015->ES5 transform there? I removed some presets to see if I could tell, and it _seems_ like it is. If so, that could potentially solve our problem. It looks like uglify doesn't do that, at least based on their demo page.

drewpc:
@roblg yeah...after some testing, it seems that you're right.  Importing the variable using `require()` means that the transform can't optimize it because it's looking at a single file.

Yes, Babili is doing the transform + minification.  See [their blog post](http://babeljs.io/blog/2016/08/30/babili) for more info.  Kinda neat!

drewpc:
I'm going to close this and put together an "architecture decision record" instead.  This should be implemented after server side compiling/Webpack is finished to meet all of the requirements.

Issue #534, opened 2016-08-12
Cannot start server on Windows


kimgronqvist:
When writing `routes_client.js` on Windows, you currently output require-statements with single backslashes - which doesn't work. 

A small change to `compileClient.js` (308) is needed. There's problaby a nicer solution, but this fixed it for me.

```js
const routesContent = routesOutput.join("").split('\\').join('\\\\');
```

doug-wade:
@kimgronqvist I would love to get react server up and building on windows, but I don't have a machine to debug on.  We hit some errors trying to [set up appveyor](https://github.com/redfin/react-server/issues/316) to prevent these kinds of compatibility issues from sneaking in.  Would you be willing to help us get the build passing on windows?

kimgronqvist:
Sure! On what command is the build failing? I've tried running `npm run bootstrap`, which worked nicely.

doug-wade:
Yay! 🎉 

I [opened up a pr](https://github.com/doug-wade/react-server/pull/14) to add appveyor to the build in my fork, since I'm not sure how to set it up so that it only builds for a single branch and not for a whole repo.  

You can see the full logs [here](https://ci.appveyor.com/project/doug-wade/react-server/build/1.0.2), or by following the link in the failing check, but currently its failing `npm test` with 

```
Error: Command failed: npm run test 
'mocha' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files (x86)\\nodejs\\node.exe" "C:\\Program Files (x86)\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "test"
```

I don't really understand why that would be the case, though, since its passing our ubuntu build just fine.  The [dev dependency we've declared](https://github.com/redfin/react-server/blob/master/packages/babel-plugin-react-server/package.json#L16) should be added by [npm to the path at runtime](https://docs.npmjs.com/misc/scripts#path).

doug-wade:
Also, the issue is https://github.com/redfin/react-server/issues/316

gigabo:
@doug-wade - Looks like there were some build failures higher up.

![image](https://cloud.githubusercontent.com/assets/115908/17637451/7524675a-6098-11e6-83b4-bba86d450977.png)


Maybe mocha (or one of its deps) just failed to build?

DraconPern:
[...]

PR #833, opened 2017-01-14
Fix file path issues

### LINKED ISSUE #832: File Path Issues in routes_client.js
When creating the client bundle with Webpack, the `__clientTemp/routes_client.js` file fails to handle Windows file paths such as `C:\my\path\to\module.js` and any path with a single apostrophe in them.  This prevents `react-server` from running on Windows and some Unix environments.

drewpc:
Fixes #832 and includes tests!  The fix was easy...the tests were hard to develop.

### DIFF
--- a/packages/react-server-cli/gulpfile.babel.js
+++ b/packages/react-server-cli/gulpfile.babel.js
@@ -1,7 +1,20 @@
 import eslint from "gulp-eslint";
 import gulp from "gulp";
 import babel from "gulp-babel";
+import jasmine from "gulp-jasmine";
 import logging from "react-server-gulp-module-tagger";
+
+function getSpecGlob (prefix) {
+	// add a wildcard onto the end if no file extension or wildcard
+	// currently present
+	let specGlob = "*[Ss]pec.js";
+	if (!specGlob.endsWith(".js") && !specGlob.endsWith("*")) {
+		specGlob += "*";
+	}
+
+	const specs = prefix + specGlob;
+	return specs;
+}
 
 gulp.task("default", () => {
 	return gulp.src("src/**/*.js")
@@ -24,7 +37,11 @@ gulp.task("eslint", [], () => {
 });
 
 // there are no tests for this project :(
-gulp.task("test", ["eslint"]);
+gulp.task("test", ["default", "eslint"], () => {
+	process.env.NODE_ENV = "__react-server-cli-unit-test__"; // eslint-disable-line no-process-env
+	return gulp.src(getSpecGlob("target/__tests__/**/"))
+		.pipe(jasmine({verbose:true, includeStackTrace: true}));
+});
 
 gulp.task("watch", () => {
 	gulp.watch("src/*.js", ['default']);
--- a/packages/react-server-cli/gulpfile.babel.js
+++ b/packages/react-server-cli/gulpfile.babel.js
@@ -37,7 +37,6 @@ gulp.task("eslint", [], () => {
 });
 
 gulp.task("test", ["default", "eslint"], () => {
-	process.env.NODE_ENV = "__react-server-cli-unit-test__"; // eslint-disable-line no-process-env
 	return gulp.src(getSpecGlob("target/__tests__/**/"))
 		.pipe(jasmine({}));
 });
--- a/packages/react-server-cli/gulpfile.babel.js
+++ b/packages/react-server-cli/gulpfile.babel.js
@@ -36,7 +36,6 @@ gulp.task("eslint", [], () => {
         .pipe(eslint.failAfterError());
 });
 
-// there are no tests for this project :(
 gulp.task("test", ["default", "eslint"], () => {
 	process.env.NODE_ENV = "__react-server-cli-unit-test__"; // eslint-disable-line no-process-env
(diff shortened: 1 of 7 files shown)


packages/react-server-cli/gulpfile.babel.js
```diff
@@ -1,8 +1,21 @@
...
 import gulp from "gulp";
 import babel from "gulp-babel";
+import jasmine from "gulp-jasmine";
 import logging from "react-server-gulp-module-tagger";
 
+function getSpecGlob (prefix) {
+	// add a wildcard onto the end if no file extension or wildcard
+	// currently present
+	let specGlob = "*[Ss]pec.js";
+	if (!specGlob.endsWith(".js") && !specGlob.endsWith("*")) {
```
drewpc: I just copied that from the other gulp files where jasmine was being used.

packages/react-server-cli/src/callerDependency.js
```diff
@@ -1,7 +1,17 @@
...
+import path from "path";
 
 export default function callerDependency(dep) {
 	// TODO: We should grab stuff based on what the routes file would get out
 	// of `require.resolve(dep)`.  Using `process.cwd()` instead for now.
-	return lookup("node_modules/" + dep, {cwd: process.cwd()});
+	let cwd = process.cwd(),
+		lookupResult;
+
+	if (process.env.NODE_ENV === '__react-server-cli-unit-test__') { // eslint-disable-line no-process-env
```
drewpc: I agree.  Got any ideas on how to bypass `callerDependency` when running unit tests inside the same directory?

drewpc:
This also fixes #534.

packages/react-server-cli/gulpfile.babel.js
```diff
@@ -24,7 +37,11 @@ gulp.task("eslint", [], () => {
 });
 
 // there are no tests for this project :(
```
roblg: I think this comment can go away now :)

packages/react-server-cli/src/callerDependency.js
roblg: I'm not sure if it's better in the overall scheme of things (mockery is black magic), but could we use `mockery` to intercept requires of `callerDependency` from the spec itself? At least that way the complexity is in the spec, rather than here.

packages/react-server-cli/gulpfile.babel.js
roblg: The other usage allows a CLI flag on the tests themselves to override `specs` (e.g. `gulp test --specs Foo`) and then appends a `*` to turn it into a glob prefix match if it isn't one already. I think here we could get rid of the `if`, unless we plan on supporting the CLI flag later.

drewpc:
@roblg I can try to refactor that using a mock.  That would be much cleaner, you're right.

drewpc:
Finished refactoring the PR to use a mock of `callerDependency` instead of mangling it like I did before.  Thanks for the ideas, guys!

packages/react-server-cli/src/__tests__/compileClient/compileClientSpec.js
```diff
@@ -0,0 +1,92 @@
...
+				const filePathRegexStrings = [
+					"var coreJsMiddleware = require(" + coreMiddlewareStringified + ").coreJsMiddleware;",
+					"var coreCssMiddleware = require(" + coreMiddlewareStringified + ").coreCssMiddleware;",
+					"require.ensure(" + filePathStringified + ", function() {",
+					"cb(unwrapEs6Module(require(" + filePathStringified + ")));",
+				];
+
+				const fileData = mockFs.toString();
+				filePathRegexStrings.map((regex) => {
+					expect(fileData).toMatch(regex.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&"));
```
gigabo: Wow, that's... quite a regex! 😳

packages/react-server-cli/src/compileClient.js
```diff
@@ -268,18 +268,18 @@ function packageCodeForBrowser(entrypoints, outputDir, outputUrl, hot, minify, l
 }
 
 // writes out a routes file that can be used at runtime.
-function writeWebpackCompatibleRoutesFile(routes, routesDir, workingDirAbsolute, staticUrl, isClient, manifest) {
+export function writeWebpackCompatibleRoutesFile(routes, routesDir, workingDirAbsolute, staticUrl, isClient, manifest) {
 	let routesOutput = [];
 
-	const coreMiddleware = require.resolve("react-server-core-middleware");
+	const coreMiddleware = JSON.stringify(require.resolve("react-server-core-middleware"));
```
gigabo: Oh, yeah, good call!

packages/react-server-cli/gulpfile.babel.js
gigabo: Oh, just realized this was left outstanding.

@drewpc @doug-wade Should we file a followup ticket for this?

gigabo: Talked with @drewpc.  Turns out this _did_ get resolved. 😌

PR #837, opened 2017-01-16
Address routr api changes & Handle POST

drewpc:
This PR makes the following changes:

- Handle the new Routr API according to their changelog (remove the `navigate` object and only pass the `method` option).
- Supports (and properly encodes) handling multiple HTTP methods other than just 'GET' (such as `method: ['get', 'post']`).  This is a breaking change because now the recommended solution in #312 of setting `method: "get"` for POST requests will no longer work.
- Changes the default method type from `method: "get"` to `method: ["get", "head"]` for feature parity between GET and HEAD requests.  Prior to this change, a HEAD request for a page that returns HTTP code 200 normally would return a 404 error.
- Adds unit tests.

### DIFF
--- a/docs/page-api.md
+++ b/docs/page-api.md
@@ -10,7 +10,7 @@ A binding of URL paths to a Page class that can render the results for those URL
 
 * The HTTP method(s) that are acceptable.
 
-* Default: “get”
+* Default: [“get”, "head"]
 
 `page: Page`
 * The page to render
--- a/docs/page-api.md
+++ b/docs/page-api.md
@@ -6,11 +6,13 @@ A binding of URL paths to a Page class that can render the results for those URL
 * A string or regex that will be tested against the path of the URL to
   determine if this route applies.
 
-`method, optional: String | [String]`
+`method, optional: undefined | String | [String]`
 
-* The HTTP method(s) that are acceptable.
+* The HTTP method(s) that are acceptable.  Use an array of strings to specify multiple methods.  Do not
+  set this value to `null` or `[]`.  To match all HTTP methods, either leave out the parameter altogether or explicitly
+  set the value to `undefined`.
 
-* Default: [“get”, "head"]
+* Default: undefined
 
 `page: Page`
 * The page to render
--- a/packages/react-server/core/__tests__/context/navigator/NavigatorRoutes.js
+++ b/packages/react-server/core/__tests__/context/navigator/NavigatorRoutes.js
@@ -10,19 +10,39 @@ const pageWrapper = {
 };
 
 const Routes = {
-	"BasicPage": {
-		"path": ["/basicPage"],
+	"GetPage": {
+		"path": ["/getPage"],
 		"page": pageWrapper,
+		"method": "get",
 	},
-	"BasicPageCaps": {
-		"path": ["/basicPageCaps"],
+	"HeadPage": {
+		"path": ["/headPage"],
 		"page": pageWrapper,
-		"method": "GET", // this should be all caps because the req.getMethod() will return 'get'.
+		"method": "head",
 	},
 	"PostPage": {
 		"path": ["/postPage"],
 		"page": pageWrapper,
 		"method": "post",
+	},
+	"PatchPage": {
+		"path": ["/patchPage"],
+		"page": pageWrapper,
+		"method": "patch",
+	},
+	"PutPage": {
+		"path": ["/putPage"],
+		"page": pageWrapper,
+		"method": "put",
+	},
+	"GetPageCaps": {
+		"path": ["/getPageCaps"],
+		"page": pageWrapper,
+		"method": "GET", // this should be all caps because the req.getMethod() will return 'get'.
+	},
+	"GetAndHeadPage": {
+		"path": ["/getAndHeadPage"],
+		"page": pageWrapper,
 	},
 	"GetAndPostPage": {
 		"path": ["/getAndPostPage"],
--- a/packages/react-server/core/__tests__/context/navigator/NavigatorRoutes.js
+++ b/packages/react-server/core/__tests__/context/navigator/NavigatorRoutes.js
@@ -40,15 +40,27 @@ const Routes = {
 		"page": pageWrapper,
 		"method": "GET", // this should be all caps because the req.getMethod() will return 'get'.
 	},
-	"GetAndHeadPage": {
-		"path": ["/getAndHeadPage"],
-		"page": pageWrapper,
-	},
 	"GetAndPostPage": {
 		"path": ["/getAndPostPage"],
 		"page": pageWrapper,
 		"method": ["get", "post"],
 	},
+	"AllMethodsPage": {
+		"path": ["/allMethodsPage"],
+		"page": pageWrapper,
+		"method": undefined,
+	},
+	"NullMethodsPage": {
+		"path": ["/nullMethodsPage"],
+		"page": pageWrapper,
+		"method": null,
+	},
+	"EmptyArrayMethodsPage": {
+		"path": ["/emptyArrayMethodsPage"],
+		"page": pageWrapper,
+		"method": [],
+	},
+
 };
 
 export default Routes;
--- a/packages/react-server/core/__tests__/context/navigator/NavigatorRoutes.js
+++ b/packages/react-server/core/__tests__/context/navigator/NavigatorRoutes.js
@@ -1,0 +1,34 @@
+import DumbPage from "./DumbPage";
+
+// This wrapper is taken from the compileClient function that writes the routes_(client|server).js file
+const pageWrapper = {
+	default: () => {
+		return {
+			done: (cb) => { cb(DumbPage); },
+		};
+	},
+};
+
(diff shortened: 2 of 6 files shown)


docs/page-api.md
```diff
@@ -10,7 +10,7 @@ A binding of URL paths to a Page class that can render the results for those URL
 
 * The HTTP method(s) that are acceptable.
 
-* Default: “get”
+* Default: [“get”, "head"]
```
doug-wade: but why?

packages/react-server/core/__tests__/context/navigator/NavigatorRoutes.js
```diff
@@ -0,0 +1,34 @@
...
+
+const Routes = {
+	"BasicPage": {
+		"path": ["/basicPage"],
+		"page": pageWrapper,
+	},
+	"BasicPageCaps": {
+		"path": ["/basicPageCaps"],
+		"page": pageWrapper,
+		"method": "GET", // this should be all caps because the req.getMethod() will return 'get'.
```
doug-wade: the fact that case sensitivity matters smells to me

packages/react-server/core/__tests__/context/navigator/NavigatorSpec.js
```diff
@@ -0,0 +1,140 @@
...
+
+
+describe("Navigator", () => {
+	let requestContext;
+	const options = {
+		routes: NavigatorRoutes,
+	};
+
+	beforeEach(() => {
+		//requestContext = new RequestContext(options);
```
doug-wade: don't leave in commented-out code without an explanation

packages/react-server/core/__tests__/context/navigator/NavigatorSpec.js
```diff
@@ -0,0 +1,140 @@
...
+		const expressRequest = new ExpressServerRequest(req);
+
+		requestContext.navigator.on('page', () => {
+			expect(expressRequest.getRouteName() === 'GetAndPostPage');
+			done();
+		});
+
+		requestContext.navigate(expressRequest, History.events.PAGELOAD);
+	});
+});
```
doug-wade: Can you add some failure cases here too?  E.g. doesn't respond to a POST request when configured with PUT only &c.

packages/react-server/core/__tests__/context/navigator/NavigatorRoutes.js
drewpc: This is testing that case sensitivity doesn't matter in the Routr package.  In versions <0.1.1, it did a case _sensitive_ match.  >=0.1.1, it added a `toLowerCase()` call first.  We want to make sure that doesn't regress.

docs/page-api.md
drewpc: Not strictly necessary and if people disagree, it can easily be changed back.  I like to have parity between GET/HEAD in most cases.

packages/react-server/core/context/Navigator.js
```diff
@@ -50,7 +50,7 @@ class Navigator extends EventEmitter {
 
 		this._haveInitialized = true;
 
-		var route = this.router.getRoute(request.getUrl(), {navigate: {path:request.getUrl(), type:type}});
+		var route = this.router.getRoute(request.getUrl(), {method: request.getMethod()});
```
roblg: I found myself wondering: how come we don't need `type` in here anymore? And then I found myself wondering: why did we need it in the first place?

The only thing I can think of offhand that might break here, is if `type` is being stored on the `route` object that's returned and we're reaching into it instead of using the `type` we're passing around everywhere.

I looked at [the commit](https://github.com/yahoo/routr/pull/29/files) in `routr`, and did some grepping, and I think we're probably in the clear. 👍

packages/react-server/core/__tests__/context/navigator/NavigatorSpec.js
```diff
@@ -0,0 +1,169 @@
+import Navigator from "../../../context/Navigator";
```
roblg: THANK YOU FOR THESE TESTS!

docs/page-api.md
roblg: Does react-server even understand `head`, or does it just return the same thing as with `get`? (You'd think I'd know the answer to this...)

roblg: Oh... is it common to have an API respond to, e.g. `POST` or `HEAD`? (i.e., should we be auto-adding `head`, to everything-ish?)

roblg:
To clarify, this is a breaking change due to the reasons outlined here: https://github.com/redfin/react-server/pull/787#discussion_r96300438

I actually wonder based on that discussion whether we should _downgrade_ `routr` in this PR (but include the new tests!), and open up another one to upgrade it? I'm torn because the earlier we change this, the better for clients of `react-server`, but...dang... this could be a nightmare to test for regressions, since `get`/`post` weren't being respected before anyway.

roblg:
> I actually wonder based on that discussion whether we should downgrade routr in this PR

(replying to myself...)

Or do we at least have a way of making it extremely obvious what one would do to verify their code doesn't break with this change in `react-server`? Do we get to publish notes about breaking changes? @gigabo

docs/page-api.md
drewpc: As it stands right now, `react-server` throws a 404 error for all HEAD requests because routes being passed to `routr` don't send the `method` attribute with it.  Therefore, `routr` defaults to 'GET' and checks the request.  `head !== get`, therefore there's no route found for HEAD/POST/PUT/PATCH/etc.

drewpc:
@roblg I think the latter: publish a HUGE notice that the bug identified in #312 was squashed and it everyone should check their routes to make sure they are explicitly specifying what the kinds of methods they expect.

packages/react-server/core/context/Navigator.js
drewpc: I'm confident in all of the changes here except this `type` thing.  I really don't understand it at all.

roblg: I don't understand why it was necessary either, since it didn't really seem to be used. It _might_ have been used with `fluxible`, which `routr` was originally built for. But we basically picked the History API and `routr` from that project (and then modified the history API, IIRC) and called it a day. So it's not particularly surprising that we weren't using the extras on `route`.

roblg:
> publish a HUGE notice that the bug identified in #312 was squashed

I dig it. 👍

drewpc:
What's the consensus on changing the default methods from `'get'` to `['get', 'head']`?  I'm happy to drop it if you all think it's weird.

doug-wade:
Embarrassing disclosure: I don't think I've ever made a `HEAD` request in my life.  Iiuc, a HEAD request is just a GET with the body pulled; if we can implement it such that if you write a reasonable GET method, we give you a reasonable HEAD method for free (by calculating the full GET and pulling the message-body), I'm down with making it a default, but otherwise, I wouldn't add it (unless of course everyone always write a HEAD method by default and I'm in the minority here).

roblg:
@drewpc I was just thinking about spring framework over in Java-land. I think that they default to _anything_ unless you specify something? 

> The above example does not specify GET vs. PUT, POST, and so forth, because @RequestMapping maps all HTTP methods by default. Use @RequestMapping(method=GET) or @GetMapping to narrow the mapping.

-- https://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html

That might actually be simpler to explain, and might be a little more intuitive? "it'll respond to anything by default; you can limit it if you want"

drewpc:
@doug-wade run the following two commands:

```
curl -I 'https://api.citybik.es/v2/networks/opole-bike'
curl 'https://api.citybik.es/v2/networks/opole-bike'
```

The first one performs a HEAD request.  The second one performs a GET request.  If you have a fronting server like nginx, it makes it even weirder because often the fronting server will transform the HEAD request into a GET request and then strip the body itself!

drewpc:
@doug-wade @roblg Looking at the latest version of `routr`, their matching algorithm is [here](https://github.com/yahoo/routr/blob/v2.1.0/lib/router.js#L49-L57).  If we remove all of the defaults, then it `routr` will match any method and it wouldn't be such a huge breaking change.  Thoughts?

roblg:
@drewpc I'm on board with matching everything by default. Curious if @gigabo has strong feelings either way as well.

drewpc:
The more I think about it, the more I like getting rid of the default.  That's definitely the best option and it makes it sooooo much smoother for our users.

drewpc:
After much discussion, I modified the PR/documentation to show that the default method is now `undefined`.  That tells `routr` to match any method.  This is still a breaking change for people who explicitly set `method: "get"`, but it won't be a breaking change for those that left method unset.

gigabo:
> This is still a breaking change for people who explicitly set method: "get"

Oh, yeah.  That'll hork Redfin.  Thanks for the heads up. 😬

drewpc:
@gigabo I updated my previous previous comment slightly for further clarification.  May still hork Redfin 😉

PR #838, opened 2017-01-16
Fix ReactServerAgent CORS preflight

### LINKED ISSUE #619: ReactServerAgent simple GET request forces CORS preflight
Make a simple GET request via ReactServerAgent and CORS will kick off a OPTIONS preflight.   Make the exact same request directly via SuperAgent and it will work fine.   Best I can tell is that ReactServerAgent is setting Content-Type on the GET request.   Perhaps here?
`
if (this._type && this._type !== 'form-data') {
        req.type(this._type);
}
`

This may be as simply as adding `&& this._method !== 'GET'` but it could be somewhere else as well.

drewpc:
Fixes #619.  ReactServerAgent shouldn't create a CORS preflight request when executing simple GET queries of an external API.  This patch fixes that problem as well as an unidentified problem where a HEAD request generate a CORS preflight request as well, because it was attempting to pass data to the request using SuperAgent's `.send()` method instead of `.query()`.

### DIFF
--- a/packages/react-server/core/ReactServerAgent/Request.js
+++ b/packages/react-server/core/ReactServerAgent/Request.js
@@ -15,7 +15,10 @@ function Request(method, urlPath, cache) {
 	this._cache = cache;
 
 	this._queryParams = [];
-	this._postParams = {};
+	// _postParams should initially be null instead of {} because we want to allow people to send POST requests with
+	// empty data sets (if they need to).  This way, we can differentiate between a default of null, an empty data set,
+	// and a populated data set.
+	this._postParams = null;
 	this._headers = {};
 	this._timeout = null;
 	this._type = "json"; // superagent's default, only for POST/PUT/PATCH methods
@@ -71,9 +74,14 @@ Request.prototype.query = function (queryParams) {
 
 Request.prototype.send = function (postParams) {
 	if (typeof postParams === 'undefined') {
-		return merge({}, this._postParams);
+		return merge({}, this._postParams || {});
 	}
-	merge(this._postParams, postParams);
+	if (postParams !== null) {
+		if (this._postParams === null) {
+			this._postParams = {};
+		}
+		merge(this._postParams, postParams);
+	}
 	return this;
 }
 
@@ -228,8 +236,10 @@ function buildSuperagentRequest() {
 		}
 	}
 
+	// query parameters exist if there are more than one set
+	// post parameters exist if the value is not null
 	const hasQueryParams = (this._queryParams.length > 0),
-		hasPostParams = (Object.keys(this._postParams).length > 0);
+		hasPostParams = (this._postParams !== null);
 
 	if (hasPostParams) {
 		// superagent has some weird, implicit file upload support
--- a/packages/react-server/core/ReactServerAgent/Request.js
+++ b/packages/react-server/core/ReactServerAgent/Request.js
@@ -18,7 +18,7 @@ function Request(method, urlPath, cache) {
 	this._postParams = {};
 	this._headers = {};
 	this._timeout = null;
-	this._type = "json"; // superagent's default
+	this._type = "json"; // superagent's default, only for POST/PUT/PATCH methods
 
 	// public field
 	this.aborted = undefined; //default to undefined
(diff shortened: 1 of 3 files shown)


packages/react-server/core/ReactServerAgent/Request.js
```diff
@@ -206,34 +206,48 @@ function buildSuperagentRequest() {
...
+				req.send(postParams);
 			}
-			postParams = formData;
-		} else {
-			throw new Error(`ReactServerAgent.type("form-data") not allowed server-side`);
-		}
-	}
 
-	req.send(postParams);
+			break;
```
doug-wade: I'm being a little stickler meeseeks

but what about the other http methods (CONNECT, OPTIONS, TRACE)?  I feel like we should at least have a default case to emit a warning or error or something.

drewpc: Fair point.  We don't have functions for those methods, so I didn't implement a check for those.  I can make this more generic.

roblg:
FYI, I'd like to take a closer look at this one. I'll try to get to it this morning.

packages/react-server/core/ReactServerAgent/Request.js
```diff
@@ -233,7 +228,35 @@ function buildSuperagentRequest() {
 		}
 	}
 
-	req.send(postParams);
+	let hasQueryParams = (this._queryParams.length > 0),
+		hasPostParams = (Object.keys(this._postParams).length > 0);
```
roblg: Nit: I think these can be `const`?

packages/react-server/core/ReactServerAgent/Request.js
```diff
@@ -233,7 +228,35 @@ function buildSuperagentRequest() {
...
+
+	if (hasPostParams) {
+		// superagent has some weird, implicit file upload support
+		// that only works if you don't set `type`.
+		if (this._type && this._type !== 'form-data') {
+			req.type(this._type);
+		}
+
+		req.send(postParams);
+	}
```
roblg: I think there might be an edge case here. It seems like one should be able to POST an empty JSON object to an endpoint, and I think we can't anymore. (Though to be fair, I think that may have been the default before, which probably also doesn't make sense?)

I suspect that there have always been lots of edge cases here that we just haven't run into yet. (I just realized that superagent accepts strings _and_ objects for `send`, and we only accept objects.

roblg:
Gah. I clicked 'approve' went I meant to just leave the review comments.

packages/react-server/core/ReactServerAgent/Request.js
drewpc: Correct...any thoughts on the right way to handle this?  Here's the basic situation: on GET/HEAD requests, if we set `type` and/or `send` to any value (even empty values), the GET/HEAD requests will send a CORS preflight.

drewpc: @roblg I handled the edge case were someone wants to send an empty JSON object now.  And there's a test for it!

roblg: Nice and nice!

packages/react-server/core/ReactServerAgent/Request.js
```diff
@@ -233,7 +236,37 @@ function buildSuperagentRequest() {
...
+			}
+			break;
+
+		case 'POST':
+		case 'PATCH':
+		case 'PUT':
+			if (hasQueryParams) {
+				logger.info(`Attempting a ${this._method} request with query data (using SuperAgent's .query() function). This might not be what you want.`);
+			}
+			break;
```
doug-wade: should we have a warning or error or something for the default case?

packages/react-server/core/ReactServerAgent/Request.js
```diff
@@ -71,9 +74,14 @@ Request.prototype.query = function (queryParams) {
 
 Request.prototype.send = function (postParams) {
 	if (typeof postParams === 'undefined') {
-		return merge({}, this._postParams);
+		return merge({}, this._postParams || {});
+	}
+	if (postParams !== null) {
```
doug-wade: I always get a little skittish when `undefined` and `null` are treated differently; is there some way we can consolidate them?

drewpc: I 100% agree.  However, it seems that the current implementation of all of these functions does something different if you call it like `.send()` (returns the currently set post parameters) vs `.send({})` (which saves the post parameters and returns `this`).  In my opinion, that is a bad design for all of those functions...trying to combine setters + getters.  That's outside the scope of this PR though.

packages/react-server/core/ReactServerAgent/Request.js
drewpc: I don't think so.  If someone is using a method other than GET, HEAD, POST, PUT, or PATCH and they're setting query/post parameters, then they probably know what they are doing and aren't worried about a CORS preflight request being emitted.

packages/react-server/core/ReactServerAgent/Request.js
drewpc: FWIW, looking at the corresponding SuperAgent functions, they _always_ return `this`.  They are not trying to combine a setter with a getter function.

roblg: FWIW, I have no recollection of why I thought it was a good idea to combine set/get. I think I was probably thinking of properties, or looking at something else. I doubt they need to be getters as well, but I haven't poked around to see what would need to be updated.

PR #844, opened 2017-01-20
Adding the source code to Chrome Timeline into React Server codebase

sbr1601:
https://chrome.google.com/webstore/detail/react-server-timeline/pockbdjjnklbgemhkhdlhbhlcfgohocc


gigabo:
Yeah!  This thing is awesome! 🤘

roblg:
This looks sweet!

Does it make sense to put this in `packages`? It doesn't _seem_ like it's being managed as an npm package, so it could probably be another top-level folder or something.

gigabo:
Only reason I could think of to have it in `packages/` is if we had tests for it (or if we wanted to lint it).

If it's in `packages/` it needs to be private and it needs a `test` script that exits zero.

doug-wade:
I vote to keep it as a package and add at least a linting target to the test script.  It might not _need_ to be an npm package, but it'd sure be easier to manage and test if we pretend like it is.

packages/react-server-chrome-extension-timeline/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "react-server-chrome-extension-timeline",
+  "version": "1.0.0",
+  "description": "This extension will display the timeline for react-server applications",
+  "main": "popup.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
drewpc: This is causing the Travis CI tests to fail.  Please add automated tests to this package so that we can keep it up to date!

drewpc:
@gigabo @doug-wade @roblg If this is something that a dev can install with `npm install react-server` and use with Chrome simply by opening up some magic plugin file from the `node_modules` folder, then I'm all for it.  If it's a plugin that needs to be installed through the Chrome App Store, then it doesn't make sense to me to include it in everyone's `node_modules` folder when they can't interact with it that way.

doug-wade:
> include it in everyone's node_modules folder when they can't interact with it that way

I don't follow.  Which node_modules folder is it going to end up in?  We build lots of projects that aren't distributed through npm in the monorepo...

drewpc:
@doug-wade I may not fully understand how the monorepo interacts with the rest of the world.  If it doesn't get included in a normal install, then my comment is irrelevant 😁

packages/react-server-chrome-extension-timeline/package.json
```diff
@@ -0,0 +1,15 @@
...
+  },
+  "keywords": [
+    "waterfall",
+    "extension"
+  ],
+  "author": "Bharath Rengarajan",
+  "license": "Apache-2.0"
```
doug-wade: Make sure to add `"private": true` so this doesn't get published to npm inadvertently.

doug-wade:
@drewpc the monorepo design doesn't affect how users install React Server, only how developers work on it.  Since none of the other modules declare a dependency on react-server-chrome-extension-timeline, it won't ever end up in their dependency closure.  Once react-server-chrome-extension-timeline is marked private, it won't even end up on npm, so it'd be impossible for a react server user to get it in their node_modules.

Issue #253, opened 2016-06-02
Download critical assets faster with <link rel="preload">


bartkusa:
(Issue derived from failed pull-request https://github.com/redfin/react-server/pull/60)

LABjs does not always download JS bundles eagerly. Nobody knows why.

We can work around this by using [<link rel="preload">](https://www.smashingmagazine.com/2016/02/preload-what-is-it-good-for/) (or "prefetch"?) to eagerly fetch resources. I ran some experiments that indicate it would speed up Time-to-Interactive between 0.5s and 1.0s for Redfin's mobile Chrome users, in semi-realistic conditions.

The trick is building a good API for marking assets (JS, CSS, images, maybe even XHRs) as critical/preloadable.

* [CanIUse: prefetch/preload](http://caniuse.com/#search=prefetch)

kcjonson:
Likewise, I think we may needs parts of this work for client side transitions.  I'm imagining that we use prefetch to grab the CSS bundles eagerly for some set of routes so that when a client transition occurs its a seamless transition.  The hard part is determining which ones should be prefetched...

gigabo:
Related: https://github.com/redfin/react-server/pull/850

CC @lidawang

doug-wade:
Check out how sweet [next.js prefetching](https://github.com/zeit/next.js#with-link-1)

```
import Link from 'next/prefetch'
// example header component
export default () => (
  <nav>
    <ul>
      <li><Link href='/'><a>Home</a></Link></li>
      <li><Link href='/about'><a>About</a></Link></li>
      <li><Link href='/contact'><a>Contact</a></Link></li>
    </ul>
  </nav>
)
```

dfabulich:
The discussion in this bug seems confused. `link rel=preload` and `rel=prefetch` have nothing to do with each other. https://css-tricks.com/prefetching-preloading-prebrowsing/

`preload` (previously known as `subresource`) is a hint to the browser's HTML parser to immediately begin downloading scripts that LAB will eventually request. http://caniuse.com/#feat=link-rel-preload `preload` is currently only supported in Google Chrome, but Apple has enabled support in the Safari Technical Preview, so it will presumably become available in iOS/Safari 11 later this year. `rel=preload` support would improve performance on Google Chrome today.

[...]

Issue #77, opened 2016-04-06
Preload JS and CSS assets also


gigabo:
https://github.com/redfin/react-server/pull/74 introduced data preload to improve client transition performance between pages on the same route, but fetching static assets will still drag down client transition performance across routes.

@aickin [proposed a solution](https://github.com/redfin/react-server/pull/74#issuecomment-206031278) that supports not only prefetch of webpack asset bundles, but also any independently included assets required by a page.

This would significantly improve first-time client transition performance across routes.

aickin:
[As @bartkusa noted](https://github.com/redfin/react-server/pull/74#issuecomment-206510291), there may still be a bug in Chrome that makes `rel=prefetch` links have higher priority than they should.

PR #846, opened 2017-01-21
Build LABjs string from source

gigabo:
I'm going to start making substantive changes to the LABjs code we put at the top of every page.  I want to have useful diffs and commit history for those changes, so as a baseline this PR replaces the pre-minified source string with the (mostly) vanilla [source file](https://github.com/getify/LABjs/blob/master/LAB.src.js).

A few benefits:

- Can opt-in to viewing debug output with `DEBUG_LAB=true react-server start`
- It's 210 bytes smaller (I'm going to make it smaller still in follow up PRs)
- It's human-readable (and hackable)

A few drawbacks:

- A dep on uglify
- A few hundred millisecond startup penalty

Could minify during build to avoid those, but we'd lose the debug-ability which is really nice.


gigabo:
I guess we could pre-build _two_ copies (normal and debug) and select among them with an option... 🤔 

That would actually let us turn on debugging at run-time which could be useful.

gigabo:
Heh. 👹 

image

doug-wade:
:thinking: iiuc, we're just checking in the source of lab.js to react server core, but is that really where it belongs?  Maybe it makes sense to make it a separate npm package?  Maybe it even makes sense to make it into an npm package that lives outside of the react server repo, maybe something like https://npmjs.com/package/labjs?  I'm not sure if the license would permit that, but it doesn't really seem like its something react server specific...

doug-wade:
Actually, looks like there's [plans to put it on npm](https://github.com/getify/LABjs/issues/113#issuecomment-236385245)

gigabo:
Oh, good thought.  I'm going to make some modifications for its use in React Server, so if anything it would probably have to be `react-server-labjs` or something like that.  But really, I'm not sure that it would be useful externally?  It's basically part of the React Server runtime in the browser.

doug-wade:
Makes sense.  I vote we pry it out into `react-server-labjs` (which also means you can postpone fixing the linting errors) and hope that we can take a dependency on the npm package when it comes out.

gigabo:
> hope that we can take a dependency on the npm package when it comes out.

We already can't.  We would need https://github.com/getify/LABjs/pull/81 to land for non-debug.

gigabo:
But anyway once _this_ (or whatever this turns into) lands I've got another PR coming on its heels that starts _really_ hacking it up. 😈

packages/react-server/core/util/LABString.js
```diff
@@ -1,11 +1,26 @@
-// Careful: We've got a local patch here!
-//
-// We've added support for a `crossOrigin` parameter.
-//
+const path = require("path");
+const {readFileSync} = require("fs");
+
+var lab = readFileSync(path.join(__dirname, "LAB.src.js"), "utf-8");
```
drewpc: Doesn't this mean that this module can't be used on the client side?

packages/react-server/core/util/LABString.js
```diff
@@ -1,11 +1,26 @@
...
+
+// Run with `DEBUG_LAB=true` to see its output in console.log.
+if (!process.env.DEBUG_LAB) { // eslint-disable-line
+
+	// This is exactly what @getify does to remove debug code.
+	// https://github.com/getify/LABjs/blob/master/BUILD.md
+	lab = lab.replace(/\/\*!START_DEBUG(?:.|[\n\r])*?END_DEBUG\*\//g, "");
+
+	// After the debug code is removed just uglify.
+	lab = require("uglify-js").minify(lab, {fromString: true}).code;
```
drewpc: We already run this whole package through Babel with Gulp, why not uglify it there and save the uglified version in `target/(client|server)/util/LABString.js`?  What am I missing?

packages/react-server/core/util/LABString.js
gigabo: Yep, this module is only used in the render middleware.  Its output is a string of JS that is put into an inline script tag.

packages/react-server/core/util/LABString.js
gigabo: > What am I missing?

The full extent of my laziness. 👹 

As I think about it more I like @doug-wade's idea of pulling this out into a separate package.  Going to manage it inside the React Server repo for now just for convenience.  We'll be able to pull it out later using [`lerna export`](https://github.com/lerna/lerna/pull/486) if we want to.

gigabo:
I've updated this PR to put LAB into a dedicated package and move the minification to build time. 🍻

gigabo:
Hard to believe [`flab`](https://www.npmjs.com/package/flab) wasn't taken. 😈

packages/flab/stringify.js
```diff
@@ -0,0 +1,9 @@
+console.log(
+	'module.exports = '+
+	require("fs").readFileSync("/dev/stdin", "utf-8")
```
roblg: I feel like I should know this, but... do we officially support windows? Seems like `/dev/stdin` wouldn't work there?

gigabo: Ugh... this is only at build time.  Do we need to support windows even for _that_? 😩

drewpc: Great catch, @roblg.  @gigabo use the `memory-stream` package for that!  Examples:

- mock socket: https://github.com/redfin/react-server/blob/master/packages/react-server/core/__tests__/reactMiddlewareSpec.js
- mock fs: https://github.com/redfin/react-server/blob/master/packages/react-server-cli/src/__tests__/compileClient/compileClientSpec.js

drewpc: Ignore the `mockery` stuff.  The `memory-stream` package allows you to pass a stream object to a library and then call `.toString()` on what was written to the stream.

gigabo: Oh cool thanks @drewpc.

I'm tempted to use [`get-stdin`](https://www.npmjs.com/package/get-stdin), which is a bit more targeted to what this little one-liner needs to do.

gigabo:
Okay, got rid of the use of `/dev/stdin`. 👍

drewpc:
While you're involved in this, it probably is a good idea to look into https://github.com/getify/LABjs/issues/113.  It might address #77 and #253 without much work from this end.

gigabo:
Oh, nice.  Thanks for the link.

packages/flab/README.md
```diff
@@ -0,0 +1,5 @@
+Future Loading and Blocking JS
+
+This is based on the awesome [LABjs](https://github.com/getify/LABjs) package.
+
+TODO: Flesh out this README. :grin:
```
doug-wade: 😆

packages/flab/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "flab",
```
doug-wade: 😆

packages/flab/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "keywords": [
+    "loading",
+    "blocking",
+    "js",
+    "asynchronous",
+    "loader"
+  ],
+  "author": "Bo Borgerson",
+  "license": "MIT",
```
doug-wade: Apache 2?

packages/flab/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "bugs": {
+    "url": "https://github.com/redfin/react-server/issues"
+  },
+  "homepage": "https://github.com/redfin/react-server#readme",
```
doug-wade: out-of-scope: should we update all of the modules to point to react-server.io as their homepage?

packages/flab/package.json
gigabo: LABjs is MIT.  Can MIT be converted to Apache 2?

packages/flab/package.json
gigabo: Oh, maybe... 🤔

gigabo:
Onward! 👉

PR #850, opened 2017-01-25
Add getCustomScripts as a lifecycle method for pages

lidawang:
Note: this is actually all @sbr1601's work, he just gave me the diff to make a separate PR from his other one.

We wanted to be able to add a custom script to our `<head/>` in order to preload some images on certain browsers (*cough* Safari *cough*). This has been tested with a test page - which is not part of this PR, b/c that was originally built in `react-server-test-pages`, but I'm wondering if that should go into `react-server-integration-tests` instead? For that matter, what's the difference between those two folders' intentions? (If it doesn't matter, I'll just push the test page I have right now into this PR.)

### DIFF
--- a/packages/react-server/core/renderMiddleware.js
+++ b/packages/react-server/core/renderMiddleware.js
@@ -311,6 +311,7 @@ function writeHeader(req, res, context, start, pageObject) {
 	// note: these responses can currently come back out-of-order, as many are returning
 	// promises. scripts and stylesheets are guaranteed
 	return Q.all([
+		renderCustomScripts(pageObject, res),
 		renderDebugComments(pageObject, res),
 		renderTitle(pageObject, res),
 		// PLAT-602: inline scripts come before stylesheets because
@@ -1048,6 +1049,24 @@ function getNonInternalConfigs() {
 		}
 	});
 	return nonInternal;
+}
+
+function renderCustomScripts(pageObject, res) {
+
+	// Want to gather these into one list of scripts, because we care if
+	// there are any non-JS scripts in the whole bunch.
+	var scripts = pageObject.getCustomScripts();
+
+	var thereIsAtLeastOneNonJSScript = scripts.filter(
+		script => script.type && script.type !== "text/javascript"
+	).length;
+
+	if (!thereIsAtLeastOneNonJSScript) {
+		renderScriptsSync(scripts, res);
+	}
+
+	// resolve immediately.
+	return Q("");
 }
 
 module.exports._testFunctions = {
--- a/packages/react-server-test-pages/pages/scripts/getCustomScripts.js
+++ b/packages/react-server-test-pages/pages/scripts/getCustomScripts.js
@@ -1,0 +1,16 @@
+export default class CustomScriptPage {
+	getElements() {
+		return [
+			<div>I'm just some test stuff</div>,
+		];
+	}
+
+	getCustomScripts(next) {
+		var inlineScript = {
+			text: "console.log('This is a custom script');",
+			type: "text/javascript",
+			strict: false,
+		}
+		return [inlineScript].concat(next());
+	}
+}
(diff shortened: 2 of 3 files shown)


gigabo:
> what's the difference between those two folders' intentions?

`react-server-integration-tests` is for automated tests.  No easy way to view the pages that are tested.

`react-server-test-pages` is a React Server site with pages that exercise interesting behavior in React Server.  You can view them and play with them.  It would be _awesome_ to have some [automated tests around these pages](https://github.com/redfin/react-server/issues/147).

I'd say if you have a page that works in `react-server-test-pages` that's a nice place to have it.  Wanna take on #147? 🙏 😁

packages/react-server/core/renderMiddleware.js
```diff
@@ -44,7 +44,7 @@ var ELEMENT_ALREADY_WRITTEN = -2;
  */
 module.exports = function(req, res, next, routes) {
 	RequestLocalStorage.startRequest(() => {
-		ACTIVE_REQUESTS++;
+		ACTIVE_REQUESTS;
```
gigabo: Huh?  What's this about?

packages/react-server/core/renderMiddleware.js
```diff
@@ -185,7 +185,7 @@ function renderPage(req, res, context, start, page) {
 
 	var routeName = context.navigator.getCurrentRoute().name;
 
-	logger.debug("Route Name: " + routeName);
+	logger.debug("Route Name: "  routeName);
```
gigabo: :squint:  Something fishy's going on here.  Did this PR just remove all of the `+` characters?

lidawang: Oops I might indeed have done that... gimme a sec

gigabo:
I feel like this feature needs to come with a big warning saying "DON'T USE THIS!!!".

Which makes me wonder if this is the right way to accomplish the image preloading.  Is there a more targeted API we could expose?

lidawang:
We could expose a `getPreloadedImages` method that just takes an array of image urls and creates a bunch of `Image`'s with them (which is what we would have done with our custom script). That would certainly be more targeted, though I wonder if there are other applications where we'd want some custom script executed in the `<head/>`, like custom tracking, or do some setup with a third party integration? Granted, if that request comes by, we could always open up this more flexible (albeit more vulnerable) API in the future.

gigabo:
> custom tracking, or do some setup with a third party integration?

Don't do that in the `<head/>`. 👿 

We control where JS goes so the page author can't accidentally block the critical rendering path.  I worry that `getCustomScripts` is a performance foot gun.  I like your `getPreloadedImages` idea.  It's safer, and actually more helpful for page authors who actually want to do this.

dfabulich:
Long term, the "right" way to handle this is with `link rel=preload` (not to be confused with `rel=prefetch`, `preconnect`, `prerender`) but Safari doesn't support `rel=preload` yet.

So I think it makes total sense to have a `getPreloadedImages` (`getPreloadedResources`?!) that renders `rel=preload` on browsers that support it (currently just Chrome, but it's spec'd/standardized) and renders optimized inline script on browsers that don't support it.

gigabo:
@dfabulich Wow, good point!

lidawang:
I am convinced! I'll get cracking on that change then 😊

gigabo:
[Support matrix](http://caniuse.com/#feat=link-rel-preload) for `rel=preload`:
image

doug-wade:
@lidawang can we close this pr, then?

dfabulich:
Yeah, just Chrome (and Opera, which is Chrome wearing a funny hat and mustache)

lidawang:
@doug-wade omg I have the power to close things? 😮 It shall be done!

gigabo:
BTW - We have https://github.com/redfin/react-server/issues/253 open for `rel=preload` of _JS bundles_.

Something to consider for your API.

Maybe not `getPreloadedImages` but a slightly more general `getAssetPreloads`?

PR #854, opened 2017-01-27
Fix Forward Test Page

drewpc:
Restore @lwang's original commit. I dorked this up in an attempt to optimize the code but didn't realize that Webpack can't statically analyze the require statements when they are in a variable.


gigabo:
Thanks @drewpc.

PR #856, opened 2017-01-28
Small edit to production guide expanding upon nginx gzip compression

krisxw:
Super small edit, but one that might save someone else a bit of time. Expanding a bit upon gzip compression as it's explained in the guide to get react-server production ready. I just hit some snags around gzip compression and spent a good thirty minutes just figuring this stuff out. The link and small explanation may save someone in the future that time..

### DIFF
--- a/docs/guides/production.md
+++ b/docs/guides/production.md
@@ -138,6 +138,7 @@ static assets from `nginx` for HTTP only.  Details are provided after the exampl
 server {
     listen 80;
     server_name www.myhostname.com;
+    gzip on;
 
     location / {
         proxy_pass http://localhost:3000;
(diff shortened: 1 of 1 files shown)


drewpc:
@krisxw Thanks!  I have the following in my `http { }` block.  Based on what you worked through, do you think this would help or be a good addition too?

```
http {
        #...other stuff....

        ##
        # Gzip Settings
        ##

        gzip on;
        gzip_disable "msie6";

        gzip_vary on;
        gzip_proxied any;
        gzip_comp_level 6;
        gzip_buffers 16 8k;
        gzip_http_version 1.1;
        gzip_types text/plain text/css application/json application/x-javascript text/xml application/xml application/xml+rss text/javascript;

        #...other stuff....
}
```

krisxw:
@drewpc overall I like your gzip configuration settings, but I'm not too sure they're good to include in this particular guide. My intention of these changes was to just point the user in the right direction to get them to use gzip with nginx, not necessarily to fully configure the file.

That said, there's real merit to at least defining some of these terms (particularly the gzip_disable, since the special value msie6 is not intuitive). I wouldn't object if you added in that full gzip configuration, though I think it might be slightly against the spirit of the guide, which I thought was to ramp people up on production react-server with minimal distractions. If I'm wrong about that though, then go for it!

drewpc:
Sounds good, thanks @krisxw!

PR #866, opened 2017-02-01
Protect multiple RSA requesters from each other

gigabo:
If two ReactServerAgent requests are made to a given endpoint only one upstream http request is actually issued, and the result is provided to both requesters.  Previously this result was passed by reference, so mutations by one requester interfered with the data for others.

This patch provides a fresh deep copy to each requester.

This has the unfortunate side effect of introducing a deep copy in the browser where we previously thought we could get away without one.  It's a minor perf hit, but it's important for data integrity.

### DIFF
--- a/packages/react-server/core/ReactServerAgent/Cache.js
+++ b/packages/react-server/core/ReactServerAgent/Cache.js
@@ -177,22 +177,16 @@ class CacheEntry {
 		this.res    = res;
 		this.loaded = true;
 
-		if (SERVER_SIDE){
-
-			// Deep copy.
-			//
-			// Leave ourselves with a clean copy of the original
-			// response regardless of what mutation might happen
-			// once stores get ahold of it.
-			//
-			// This is important to ensure that we provide the same
-			// data from the cache when we wake up in the browser
-			// as we initially provide on the server.
-			//
-			res = JSON.parse(JSON.stringify(res));
-		}
-
-		this.dfd.resolve(res);
+		// Resolve with a serialized copy.  We'll unserialize for each
+		// requester.  This way we provide a fresh copy each time so mutations
+		// don't leak.
+		//
+		// This also leaves _us_ with a clean copy of the original response.
+		// This is important to ensure that we provide the same data from the
+		// cache when we wake up in the browser as we initially provide on the
+		// server.
+		//
+		this.dfd.resolve(JSON.stringify(res));
 	}
 
 	setError (err) {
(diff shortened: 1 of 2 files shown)


packages/react-server/core/ReactServerAgent/Cache.js
```diff
@@ -177,22 +177,16 @@ class CacheEntry {
...
+		// Resolve with a serialized copy.  We'll unserialize for each
+		// requester.  This way we provide a fresh copy each time so mutations
+		// don't leak.
+		//
+		// This also leaves _us_ with a clean copy of the original response.
+		// This is important to ensure that we provide the same data from the
+		// cache when we wake up in the browser as we initially provide on the
+		// server.
+		//
+		this.dfd.resolve(JSON.stringify(res));
```
roblg: I apparently can't leave comments willy-nilly in this file, but should we do the same thing for error responses (line ~209)

gigabo: Yeah, it's annoying that you can click to _see_ those lines but can't comment on them. 👿

gigabo: Nice catch, didn't realize those were handled separately. 👍

doug-wade:
Rather than pass a different mutable copy to each requester, why not pass a single immutable reference to each requester?

gigabo:
@doug-wade How would that work?

doug-wade:
If we were willing to add a [library](https://facebook.github.io/immutable-js/) for immutability, it'd be really simple

```
const Immutable = import("immutable");
...
this.dfd.resolve(Immutable.map(res));
```

If instead we wanted to roll our own, we could do a recursive freeze:

```
this.dfd.resolve(deepFreeze(res));

function deepFreeze (o) {
  Object.freeze(o);

  Object.keys(o).forEach((prop) => {
    if (typeof o[prop] === "object") {
      deepFreeze(o[prop]);
    }
  });
  
  return o;
}
```

gigabo:
@doug-wade I see.  That would be a pretty big breaking change.

Worth discussing, but I'd still like to get a fix for the bug out in the meanwhile. 🐛

packages/react-server/core/ReactServerAgent/Cache.js
```diff
@@ -223,7 +217,7 @@ class CacheEntry {
 			// server-side, we increment the number of requesters
 			// we expect to retrieve the data on the frontend
 			this.requesters += 1;
-			return this.dfd.promise;
+			return this.dfd.promise.then(val => JSON.parse(val));
```
drewpc: don't we need to parse the JSON in both the `if` and `else` blocks?

gigabo: Oh, probably. 😁 

Nice catch @drewpc!

drewpc:
Seems like there's more work and it's failing the tests.  Should I close the PR for now?

gigabo:
It's not dead yet!  It could pull through!

gigabo:
> don't we need to parse the JSON in both the if and else blocks?

Good call.  Done.

> should we do the same thing for error responses

I don't think so, actually.  Error responses are less likely to be mutated, and their handling is out of band.

> adding some tests

The coverage here was already pretty good (thanks @roblg), so I just modified existing tests to verify that values are deeply equal, but not references.

So... how about it? 😁

PR #867, opened 2017-02-02
Adding large data request site to provide additional testing for 866

krisxw:
When originally writing the react-server test sites, I did not account for the possibility that larger data requests, as opposed to the quantity of data requests, may be a bottleneck or performance regression. This change adds an additional test site that makes 100 larger data requests locally. 

In local testing, the page load times for this page were 6084ms, 5318ms, 5902ms, 5339ms, and 5416ms, for an average load time of 5612ms, in case this information is relevant after #866 .


gigabo:
Thanks @krisxw! 🤘 

FYI - If you prefix issue/PR numbers with a `#` GitHub will link them. Like: #866.

Issue #868, opened 2017-02-05
Option to Bypass Compiling Client Code on Startup


drewpc:
Add an option to tell `react-server` that the client code has already been compiled and to bypass compiling when starting the server.  This is especially useful when baking the application on to an image that will be used to start a number of servers.  The `compile` command can be run during the baking process and then, at run time, servers bypass compiling and can start much more quickly.

PR #869, opened 2017-02-05
Add compileOnStartup Option

### LINKED ISSUE #801: Remove compileOnly option
Following up #794 

---

I haven't found documentation on this, but in the code there's still a `compileOnly` option. Given that there is a `compile` command, I don't think this option makes sense anymore. Correct me if I'm wrong.

drewpc:
Addresses #868.  Tests and documentation have been updated as well.  Also, fixes #801 by removing references to the old `compileOnly` option from source code and documentation.  Lastly, fixes a minor issue with the Promise chain within the cli script.

### DIFF
--- a/docs/guides/react-server-cli.md
+++ b/docs/guides/react-server-cli.md
@@ -383,7 +383,6 @@ run({
     jsPort          : 3001,
     hot             : true,
     minify          : false,
-    compileOnly     : false,
     jsUrl           : null,
     longTermCaching : true,
 });
--- a/docs/guides/react-server-cli.md
+++ b/docs/guides/react-server-cli.md
@@ -157,7 +157,8 @@ export default (server, reactServerMiddleware) => {
 	server.use(compression());
 	server.use(bodyParser.urlencoded({ extended: false }));
 	server.use(bodyParser.json());
-	server.use(session(options));
+	server.use(helmet());
+	server.use(session(options));  // Custom express middleware being added
 	reactServerMiddleware(); // Must be called once or server will not start
 }
 ```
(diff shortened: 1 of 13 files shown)


packages/react-server-cli/src/commands/start.js
```diff
@@ -238,15 +242,41 @@ export default function start(options){
...
+			});
+
+			// mock the compiler object used by the various JS servers so that the compiler.run function always succeeds.
+			// This will allow the JS servers to work properly, thinking that the compiler actually ran.
+			compiler = {};
+			compiler.run = (cb) => {
+				cb(null, null);
+			};
+		} else {
+			({serverRoutes, compiler} = compileClient(options));
```
roblg: nit: I find the parentheses around this expression make it harder to parse visually.

drewpc: Me too.  Here's the issue I have: I defined `serverRoutes` and `compiler` previously using `let`.  ES6 destructuring only works like this:
```
let {a, b} = stuff;
const {c, d} = stuff;
let e, f;
({e, f} = stuff);
```
This is a syntax error: `{a, b} = stuff;`.  I could put a comment above that line so that it's clear?

gigabo: Huh, TIL.

docs/guides/react-server-cli.md
```diff
@@ -157,7 +157,8 @@ export default (server, reactServerMiddleware) => {
 	server.use(compression());
 	server.use(bodyParser.urlencoded({ extended: false }));
 	server.use(bodyParser.json());
-	server.use(session(options));
+	server.use(helmet());
```
gigabo: Why is this adding helmet here now?

drewpc: I read through the documentation to update it with `compileOnStartup` stuff and noticed that the custom express middleware example wasn't updated previously.  I can remove it if it's an issue.

gigabo: Oh, somehow I thought we were adding it automatically now.  Makes sense. 👍

drewpc: well...we are because of #812, but we missed updating this example to depict the new default express middleware.

gigabo: Oh, I see, and users are responsible for adding it themselves if they override the default with custom middleware.

Nice catch.

Issue #840, opened 2017-01-19
`UnhandledPromiseRejectionWarning` in react-server-website


davidalber:
### Reproduction
1. Follow the instructions in the [react-server-website README](https://github.com/redfin/react-server/blob/master/packages/react-server-website/README.md):
```
$ git clone git@github.com:redfin/react-server.git
$ cd react-server
$ npm run bootstrap
$ cd packages/react-server-website
$ npm run build-assets
$ npm start
```
2. Navigate to http://localhost:3010 in a browser.

### Bug
The following appears in the service log:
```
2017-01-19T07:47:45.564Z - error: [react-server-data-bundle-cache.main]  name=react-server-data-bundle-cache, message=RLS() access
 outside of request!, stack=Error: RLS() access outside of request!
    at react-server/node_modules/request-local-storage/lib/server/index.js:26:35
    at getCurrentRequestContext (context/RequestContext.js:19:10)
    at exports.default (react-server/packages/react-server-data-bundle-cache/lib/listen.js:10:12)
    at before (react-server/packages/react-server-data-bundle-cache/lib/before.js:25:2)
    at react-server/node_modules/continuation-local-storage/context.js:84:17
    at _fulfilled (react-server/node_modules/q/q.js:834:54)
    at self.promiseDispatch.done (react-server/node_modules/q/q.js:863:30)
    at Promise.promise.promiseDispatch (react-server/node_modules/q/q.js:796:13)
    at react-server/node_modules/q/q.js:857:14
    at runSingle (react-server/node_modules/q/q.js:137:13)
2017-01-19T07:47:45.569Z - fast: [react-server-data-bundle-cache.main] stats.error.returned ms=14
(node:69432) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: RLS() access outside of request!
```

So far as I can see, the site is loading fine. There's nothing concerning in the JavaScript console.

Produced using Node v6.9.4, npm v3.10.10.

gigabo:
Thanks for reporting @davidalber.

davidalber:
How do we feel about deploying a new version of the website while this bug exists? I couldn't detect any usability issues, but I didn't look very thoroughly.

drewpc:
[...]

PR #870, opened 2017-02-07
Change title for home page

emecell:
To get more SEO traffic, we should adjust the title for what people
search for. Change recommended by Joe Lei.

### DIFF
--- a/packages/react-server-website/pages/homepage.js
+++ b/packages/react-server-website/pages/homepage.js
@@ -5,7 +5,7 @@ import './homepage.less';
 export default class Homepage {
 
 	getTitle() {
-		return "React Server";
+		return "React Server - Fast Server and Client Side Rendering";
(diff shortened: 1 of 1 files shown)


emecell:
Right now we're ranked #20 for when you search for "react server side rendering". Let's try to at least get onto the first page and hopefully top 5. We probably also want to explore adding some more content to our home page like companies using React Server and community stats.

gigabo:
Nice, good idea!

gigabo:
@davidalber This will require a re-deploy of the website. 👹

davidalber:
Yeah, there are a couple other changes that are waiting for a deployment. I've been meaning to ask a question in #840 about how we feel about that, which I just did [here](https://github.com/redfin/react-server/issues/840#issuecomment-278827940).

PR #871, opened 2017-02-07
Get dynamic config from process.env.REACT_SERVER_CONFIGS

mocheng:
There is a hidden feature in [`packages/react-server/core/config.js`](https://github.com/redfin/react-server/blob/master/packages/react-server/core/config.js), it read `config.json` file from folder specified by `process.env.REACT_SERVER_CONFIGS` and client side gets same config by rehydration. However, since it is json file, there is no way to config dynamically. For example, different API endpoint for dev env and production env.

This PR make `react-server` try to load `config.js` file in `process.env.REACT_SERVER_CONFIGS` folder first. If `config.js` doesn't exist, then `config.json` file is used.

An exmaple of  `.config.js` file might be like below.
```
if (process.env.NODE_ENV==='production') {
  module.exports = {
     api: 'http://some-domain/'
  };
} else {
  module.exports = {
     api: 'http://dev.some-domain/'
  };
]
```
This might be helpful for a lot of application.

### DIFF
--- a/packages/react-server/core/config.js
+++ b/packages/react-server/core/config.js
@@ -19,7 +19,7 @@ if (SERVER_SIDE) {
 				// Node.js tries to load `config.js` file first. If `config.js` doesn't exist, Node.js
 				// then try to load `config.json`.
 				configFilePath = path.join(process.cwd(), configFilePath + "/config");
-				config = Object.freeze(configFilePath);
+				config = Object.freeze(require(configFilePath));
 			} else {
 				config = Object.freeze({});
 			}
--- a/packages/react-server/core/config.js
+++ b/packages/react-server/core/config.js
@@ -1,4 +1,3 @@
-
 /**
  * Thin wrapper around the environment-specific configuration file
  */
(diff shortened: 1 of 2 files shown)


packages/react-server/core/config.js
```diff
@@ -10,12 +10,18 @@ if (SERVER_SIDE) {
...
+				//eslint-disable-next-line no-process-env
+				config = require(path.join(process.cwd(), process.env.REACT_SERVER_CONFIG_SOURCE));
+			}
+			//eslint-disable-next-line no-process-env
+			else if (process.env.REACT_SERVER_CONFIGS) {
 				var fs = require("fs");
-				/*eslint-disable no-process-env */
+				//eslint-disable-next-line no-process-env
 				var configFile = fs.readFileSync(process.env.REACT_SERVER_CONFIGS + "/config.json");
-				/*eslint-disable no-process-env */
```
mocheng: `eslint-disable` in original code actually disable eslint check for all below code. So, I just update it with `eslint-disable-next-line`.

karlhorky:
The integration tests I introduced in #828 are failing. These are starting an actual `react-server` via the CLI - not sure if this is an issue.

drewpc:
The guide for [Running a Production Server](https://react-server.io/docs/guides/production) describe how to change these configs per-environment.  Do you think that accomplishes what you're trying to do here?

mocheng:
@drewpc It is feasible to have different `REACT_SERVER_CONFIGS` in npm script. But, in case there are a lot common config values in different config files, e.g. 100 config values is same for both `development` and `beta` env, it would be a burden to keep two config json file same. In this case, yaml file would be a better choice which need to be run by js not json.

drewpc:
Closing & reopening PR to force a Travis rebuild.

mocheng:
@drewpc Nice suggestion.

mocheng:
Update code per @drewpc 's suggestion.

mocheng:
I have rebased this branch to master, but the travis unit test still failed. 
Interestingly, I beheld some PR passed the travis check. Anybody has any clue to make it work?

packages/react-server/core/config.js
```diff
@@ -10,13 +9,17 @@ if (SERVER_SIDE) {
...
-				config = Object.freeze(JSON.parse(configFile));
+				var path = require('path');
+				//eslint-disable-next-line no-process-env
+				var configFilePath = process.env.REACT_SERVER_CONFIGS;
+
+				// Node.js tries to load `config.js` file first. If `config.js` doesn't exist, Node.js
+				// then try to load `config.json`.
+				configFilePath = path.join(process.cwd(), configFilePath + "/config");
+				config = Object.freeze(configFilePath);
```
roblg: should this be `Object.freeze(require("configFilePath"))`?

drewpc: Whoops.  I had to add the Object.freeze and removed the require() part.  Fixed!

gigabo:
@roblg Have your concerns been addressed here?

roblg:
@gigabo yeah, seems reasonable to me

gigabo:
Thanks @mocheng. 👍

PR #876, opened 2017-02-10
Fix react-server-cli AVA test issues

drewpc:
Addresses issues discussed [here](https://react-server.slack.com/archives/discussion/p1485902484000972) and, possibly, reasons why #871 is failing build.

* Uses custom ports for the CLI tests
* Tests additional edge case for stdout/stderr in case there are other errors in stderr that need to be displayed.


drewpc:
@gigabo Because the build on `master` is failing right now, I'm going to approve & merge this PR right now to see what's happening.  If there's public outcry about it, let me know and I'll reverse it.

PR #877, opened 2017-02-10
Migrate back to Lerna :tada:

gigabo:
Lerna supports hoisting now, so I think we're all set to migrate back.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ Not sure where to start? Join us [on slack](https://slack.react-server.io/) and 
 
 ## Getting started
 
-React Server uses a tool called [asini](https://www.npmjs.com/package/asini) to
+React Server uses a tool called [Lerna](https://www.npmjs.com/package/lerna) to
 manage the entire React Server ecosystem in a single repository.  This allows
 maintainers to know that everything works together well, without having to know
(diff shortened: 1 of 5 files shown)


drewpc:
Sweet!  Looks like there's still some [asini references](https://github.com/redfin/react-server/search?p=1&q=asini&utf8=✓) in the codebase: `.gitignore` and some `react-server-website` references.  Do those need to be changed as well?

gigabo:
Oh, nice catch on the `.gitignore`.  Didn't check hidden files.  Seems like no harm in leaving it?

The website references are actually for the Asini slackin site, which we're keeping up for now.

gigabo:
Any objections here?  Users are [running into issues](https://github.com/asini/asini/issues/82) that [have already been fixed](https://github.com/lerna/lerna/pull/586) on Lerna.

drewpc:
Conceptually, I'm all for it.  Practically, however, I obstain on this one because I don't know enough about either lerna or asini to provide any insight.

gigabo:
@drewpc That's fair.  Asini's dead, though, so it's a matter of getting back onto a tool that's maintained.

Gonna merge this.  You'll need to `npm i` again when you pull it, but other than that it should be a seamless transition. 🤞 

Let me know if you run into any trouble!

Issue #775, opened 2016-12-01
Drop frameback


gigabo:
Nobody's using frameback, right?

gigabo:
Closed by #880. 🎉

PR #880, opened 2017-02-13
Issue #775 Drop Frameback

szhou8813:
No one uses frameback, so we're dropping it.


packages/react-server-test-pages/pages/navigation/playground.js
```diff
@@ -131,8 +105,6 @@ export default class NavigationPlaygroundPage {
 					<li>CT: Client Transition</li>
 					<li>RD: Reuse DOM</li>
 					<li>BD: Bundle Data</li>
-					<li>FB: Frameback</li>
-					<li><span className='not-available'>FB</span>: Frameback disabled (already in a frame)</li>
```
gigabo: There should be some styling (strikethrough, I think) associated with the `not-available` class that can be removed from the CSS, too.

gigabo:
This is awesome @szhou8813!  It makes the history navigation stack management _so much simpler!_

I think there's probably some documentation that will need to be tweaked, too.

packages/react-server-test-pages/pages/navigation/playground.js
szhou8813: Sure, I ll check that out!

szhou8813:
Oh I guess we ll need to update tutorials/docs on react-server.io too. How do I do that?

davidalber:
The content on react-server.io is queried from GitHub. If you update the appropriate docs in the repository it will be reflected on the site within one minute of merging to master.

gigabo:
Awesome!  Thanks for dealing with this @szhou8813!

roblg:
Added `breaking-change`. I believe it's probably unused, but you never know ;)

PR #881, opened 2017-02-13
Add an option to split bundle load from execution

gigabo:
Experiments at Redfin with the `jsBelowTheFold` option have turned up some improvement in time to display content at the expense of degradation of time to make the page _interactive_.  This new option is an attempt to retain the benefit while regaining some of the lost ground in terms of interactivity.

The _load_ of JS bundles is started immediately, but the _parse_ of the bundles is delayed until `<TheFold/>` is reached.

Pages can opt into this using the response from `handleRoute()`:

```js
handleRoute() {
  ...
  return {code: 200, splitJsLoad: true};
}
```

There is also a `?_debug_split_js_load=1` option that may be used for testing.

### DIFF
--- a/packages/flab/LAB.src.js
+++ b/packages/flab/LAB.src.js
@@ -8,6 +8,7 @@
 
 		// constants for the valid keys of the options object
 		_UseLocalXHR = "UseLocalXHR",
+		_UseCORSXHR = "UseCORSXHR",
 		_AlwaysPreserveOrder = "AlwaysPreserveOrder",
 		_AllowDuplicates = "AllowDuplicates",
 		_CacheBust = "CacheBust",
@@ -149,8 +150,14 @@
 					script.src = src;
 					// NOTE: no append to DOM yet, appending will happen when ready to execute
 				}
-				// same-domain and XHR allowed? use XHR preloading
-				else if (preload_this_script && src.indexOf(root_domain) == 0 && chain_opts[_UseLocalXHR]) {
+
+				// This is the ultimate fallback in React Server.  The
+				// "cache-preloading" option in stock LABjs doesn't work in
+				// modern Chrome, so... this is our last best hope.  If you're
+				// configured for splitJsLoadFromExecution then you'd better
+				// have xhr access to your scripts!  They need to either be on
+				// the same domain or have CORS headers.
+				else if (chain_opts[_UseCORSXHR] || (src.indexOf(root_domain) == 0 && chain_opts[_UseLocalXHR])) {
 					xhr = new XMLHttpRequest(); // note: IE never uses XHR (it supports true preloading), so no more need for ActiveXObject fallback for IE <= 7
 					if (chain_opts[_Debug]) log_msg("start script preload (xhr): "+src);
 					xhr.onreadystatechange = function() {
@@ -204,6 +211,7 @@
 
 		// global defaults
 		global_defaults[_UseLocalXHR] = true;
+		global_defaults[_UseCORSXHR] = false;
 		global_defaults[_AlwaysPreserveOrder] = false;
 		global_defaults[_AllowDuplicates] = false;
 		global_defaults[_CacheBust] = false;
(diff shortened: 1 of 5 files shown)


gigabo:
> 139 specs, 85 failures

😳 

That must be some sort of record...

gigabo:
Ugh, looks like the XHR loading doesn't work in the jasmine tests.  I guess they were falling through to cache-preloading?  That doesn't work in modern Chrome.

How does React Server even work? 🤦‍♂️

gigabo:
Okay, tests all passing. 👍

drewpc:
This may not have anything to do with anything, but when I was working on the server-side webpack stuff I ran into a similar issue with the tests failing after combining the HTML + JS servers (thus removing the separate CORS domains).  I placed the following comment in my file when I made the change:

```
		if (process.env.NODE_ENV === "test") { // eslint-disable-line no-process-env
			// ZombieJS does not handle loading async scripts in the proper order and ensuring that .wait() waits for
			// execution.  Thus, we need to force LABjs to not use local XHRs in order for all tests to work properly.
			// This used to work prior to combining the HTML and JS servers into a single server because the JS files
			// were considered part of a different domain when running tests, so it never used XHR preloading, it used
			// "cache preloading".
			// https://github.com/getify/LABjs/blob/master/LAB.src.js#L152-L175
			res.write("$LAB.setGlobalDefaults({UseLocalXHR:false,AlwaysPreserveOrder:true});");
		} else {
			// We always want scripts to be executed in order.
			res.write("$LAB.setGlobalDefaults({AlwaysPreserveOrder:true});");
		}
```

gigabo:
> I ran into a similar issue with the tests failing after combining the HTML + JS servers

Yeah, ZombieJS is just weird enough to stumble on things that browsers don't have trouble with.  In this case I was able to address it with a check that wasn't literally `if (test)`, which I prefer when possible.  Thanks for the context.

Any objections to this?  Pretty excited to try it out!

packages/flab/LAB.src.js
```diff
@@ -149,8 +150,14 @@
...
-				// same-domain and XHR allowed? use XHR preloading
-				else if (preload_this_script && src.indexOf(root_domain) == 0 && chain_opts[_UseLocalXHR]) {
+
+				// This is the ultimate fallback in React Server.  The
+				// "cache-preloading" option in stock LABjs doesn't work in
+				// modern Chrome, so... this is our last best hope.  If you're
+				// configured for splitJsLoadFromExecution then you'd better
+				// have xhr access to your scripts!  They need to either be on
+				// the same domain or have CORS headers.
+				else if (chain_opts[_UseCORSXHR] || (src.indexOf(root_domain) == 0 && chain_opts[_UseLocalXHR])) {
```
krisxw: I only have a cursory familiarity with LABjs, but I wonder if this comment is discoverable enough, since I'm not sure it's obvious that you need xhr access in order to use this (though I would imagine access would generally be granted unless it's cross-domain or some such). Will this actually pop up for people in the real world?

packages/flab/LAB.src.js
```diff
@@ -8,6 +8,7 @@
 
 		// constants for the valid keys of the options object
```
krisxw: It's tough for me to tell how the options object is set for LABjs. Looks like we pass just a couple of parameters into the LABjs `globaldefaults` and let our typical defaults dictate all of the other options. This is a nit pick, but I find that flow a bit confusing. It's certainly the most line-efficient way to go about it, though, and if it doesn't matter most of the time then feel free to ignore this.

packages/react-server/core/util/PageUtil.js
```diff
@@ -191,6 +191,8 @@ var PAGE_CHAIN_PROTOTYPE = {
 	setHasDocument     : makeSetter('hasDocument'),
 	getJsBelowTheFold  : makeGetter('jsBelowTheFold'),
 	setJsBelowTheFold  : makeSetter('jsBelowTheFold'),
+	getSplitJsLoad     : makeGetter('jsBelowTheFold'),
```
sresant: should this be splitJsLoad?

gigabo: Nice catch!

PR #882, opened 2017-02-15
Updated delete method

cashhh:
The delete method now takes another argument for 'data' like its sibling methods do and uses the 'send' prototype to pass the data.


gigabo:
Thanks @cashhh.  Nice to have this consistent with the other methods. 👍

cashhh:
Awesome. Glad I could help.

Issue #885, opened 2017-03-01
Custom 404 Page


feychou:
Is there a way to easily customise the 404 response - for instance with a custom page?

I also tried to add my catch-all route in the route entry file and associate it with a not-found-page: however, this results in a 200. Is there a way to define the 404 status code in the page lifecycle?

gigabo:
> Is there a way to define the 404 status code in the page lifecycle?

Yep, it's not well documented, but you should be able to do this by adding `hasDocument` to your response from `handleRoute`:
```js
class NotFoundPage {
    handleRoute() {
        return {code: 404, hasDocument: true};
    }

    getElements() {
        return <Awesome404 />;
    }
}
```

feychou:
@gigabo That works, thank you. I could open a PR for documenting this somewhere in the docs (maybe where I would have expected to find it). Maybe one of the examples could also have a custom 404 page?

gigabo:
@feychou Yeah, that would be awesome!

doug-wade:
This was fixed in #893; closing

PR #893, opened 2017-03-10
Document hasDocument option

feychou:
I also added a custom `404` example in bike-share.
This is the related issue #885


gigabo:
Awesome!  Thanks @feychou!

gigabo:
Looks like something weird happened with Travis.  It's stuck "waiting for status to be reported", and it's not even giving me a link to the build so I can re-start it. 😕

feychou:
Odd, I was just wondering where the link is. Anyway I wrote something else in the docs, hope this triggers the build.

gigabo:
Yep, that did the trick.  Thanks @feychou!

PR #895, opened 2017-03-13
Upgrade meteor to use new ReduxAdapter

ptomasroos:
meteor-site is not using the new ReduxAdapter and since the new ReduxAdapter was introduced I guess the goal should be use it, because otherwise it makes no sense to have it in the framework.

So here it goes.

Also upgraded to latest react-server 0.6.0 version
And exposed the redux store to devtools for easier debugging on server-side transfered state on @@init


sresant:
LGTM, thanks for doing this @ptomasroos!

PR #899, opened 2017-03-16
Updating generator-react-server to use ^v0.6.0

roblg:
`generator-react-server` got missed when 0.6.0 was released, resulting in confusing issues for folks trying to use features that are new in 0.6.0.


gigabo:
I wonder if we could trick Lerna into treating the generator template directory as a package so it would manage the versions for us... 🤔

PR #902, opened 2017-03-21
Gitignore string/ in flab

doug-wade:
Ran bootstrap and found I had untracked files -- looks like they should be gitignored.


gigabo:
Oh, yep.  Thanks @doug-wade.

PR #905, opened 2017-03-23
Async flag to allow scripts to be rendered async or not

zlegein:
Add an async flag to allow fine grained control over how scripts should be rendered. Helpful when you need to include some third party libraries.


zlegein:
I was able to achieve what was needed by using *getBodyStartContent* and passing in scripts there.
If you are looking to place a script synchronously in the HEAD then you can assign a type="application/javascript" and force the loading of all scripts synchronously, not very elegant, but works, just note any browser incompatibilities. 

This PR may is no longer needed, closing.

Issue #879, opened 2017-02-12
How to make dynamic routes?


mtrabelsi:
I've been reading through the documentation and I can't find any way to dynamically create routes.
Or at least there is no clear statement about this, looks like react-server deliberately make us free to use static files specified in **routes.json** only.

However, I found this in the doc - we can use dynamic routes, by using routes.**js** intead of .**json** : 

```
const fs = require('fs');
const path = require('path');

const routes = fs.readdirSync('./pages').reduce((acc, curr) => {
      const baseName = path.basename(curr, 'js');
      return {
            ...acc,
            [`${baseName}Page`]: {
                  path: [`/{baseName}`],
                        page: `pages/${baseName}`
            }
        };
}, {})
module.exports = {
    middleware: fs.readdirSync('./middleware'),
    routes
}
```

That's cool, but still not sufficient, its still generating routes from static files.
We should have another way for making route ( like in expressJs routes using wildcards )

Is react-sever aware of this?

drewpc:
It depends on how far you want to go with 'dynamic routes'.  Take a look at the `react-server-website` package `routes.js` file.  It uses [regex](https://github.com/redfin/react-server/blob/master/packages/react-server-website/routes.js#L19-L23) to identify [path name variables](https://github.com/redfin/react-server/blob/master/packages/react-server-website/pages/docs.js#L15-L18)

`react-server` uses [`yahoo/routr`](https://github.com/yahoo/routr/blob/master/docs/routr.md) for pattern matching.  That documentation is not so great.  It complies with [`pillarjs/path-to-regexp`](https://github.com/pillarjs/path-to-regexp)

Our documentation could be improved greatly.  Interested in helping with that?

mtrabelsi:
Nice! that's exactly what I'm looking for. Thanks for your comment.
I wonder how to contribute to the doc? Should I create a regular PR in this repo?

drewpc:
Yup, just a regular PR in this repo.  The [website](https://react-server.io/docs) pulls documentation in the Markdown format from the [docs directory](https://github.com/redfin/react-server/tree/master/docs).

mtrabelsi:
@drewpc : 

I cloned the repo: 

`$ git clone https://github.com/redfin/react-server.git`

then ran : 

`$ npm run bootstrap`

After that, I made some changes to the file **docs/guides/react-server-cli.md** then compiled the files : 

```
$ cd packages/react-server-website
$ npm run build-assets
```
Finally started the website : 

[...]

PR #909, opened 2017-03-24
Document LOCAL_DOCS for website

doug-wade:
Per @gigabo's comment here https://github.com/redfin/react-server/issues/879#issuecomment-288243997


gigabo:
Thanks @doug-wade!

PR #918, opened 2017-04-06
Allow group selection in `logging.addRewriter`

gigabo:
This makes it possible to add a rewriter only for the main logger, for example.

This also makes the interface consistent with `addTransport` and `setLevel`.

This is a breaking change.


gigabo:
Looks like the CLA integration had a hiccup.  Bypassing.

PR #928, opened 2017-04-11
Front load and cache some page chain class prep

gigabo:
This takes some work that used to happen for each request and caches it.

This is mostly groundwork for fine-grained instrumentation of page lifecycle methods.  It incidentally also slightly reduces the overhead of building up the page chain object during each request.

[`core/__bench__/handlePage.js`](https://github.com/redfin/react-server/pull/926) before:
```
1       0.000481308279135711
10      0.0009405891195623999
100     0.005067864883424908
1000    0.04754033069230767
```

[`core/__bench__/handlePage.js`](https://github.com/redfin/react-server/pull/926) after:
```
1       0.0004272630582241284
10      0.0006927164659505115
100     0.0034683026867557273
1000    0.031324801985714294
```

Basically reduces middleware overhead to from ~50µs to ~30µs.  Not a huge deal, but the instrumentation is likely to add a little overhead, so it's nice to pay some of that down ahead of time. :grin:

### DIFF
--- a/packages/react-server-integration-tests/src/specRuntime/testHelper.js
+++ b/packages/react-server-integration-tests/src/specRuntime/testHelper.js
@@ -298,7 +298,7 @@ var testSetupFn = function (specFile, routes, httpsOptions) {
 			// Since we're not using hot-reloading, we need to explicitly delete all of the require.cache to ensure
 			// that the latest Webpack compiled server code is being used.
 			Object.keys(require.cache)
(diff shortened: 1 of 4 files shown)


gigabo:
Ugh, no idea why those zombie tests are failing.  Everything looks good browser-side in `react-server-test-pages` and `react-server-website`. :hurtrealbad:

Issue #910, opened 2017-03-25
withCredentials is not implemented in ReactServerAgent


matthewdwong:
Request.js:38 Uncaught (in promise) Error: withCredentials() from superagent's API isn't implemented yet

gigabo:
Thanks for reporting @matthewdwong.  Shouldn't be too hard to get it patched through.

PR #935, opened 2017-06-15
Implement superagent's req.withCredentials() method

mlolson:
https://github.com/redfin/react-server/issues/910

The [superagent implementation](https://github.com/visionmedia/superagent/blob/f5e39b7746268901c591dde08cd68a5e3944ced1/lib/request-base.js#L396), for reference.

### DIFF
--- a/packages/react-server/core/ReactServerAgent/Request.js
+++ b/packages/react-server/core/ReactServerAgent/Request.js
@@ -22,7 +22,8 @@ function Request(method, urlPath, cache) {
 	this._headers = {};
 	this._timeout = null;
 	this._type = "json"; // superagent's default, only for POST/PUT/PATCH methods
-
+	// CORS off by default
+	this._withCredentials = false;
 	// public field
 	this.aborted = undefined; //default to undefined
 }
@@ -122,6 +123,15 @@ Request.prototype.toJSON = function(){
 		urlPath: this._urlPath,
 	};
 }
+
+// Enable CORS
+Request.prototype.withCredentials = function (on) {
+	if (on === undefined) {
+		on = true;
+	}
+	this._withCredentials = on;
+	return this;
+};
 
 /**
  * Wrap superagent's end() method to create an entry in a request-local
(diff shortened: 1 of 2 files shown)


packages/react-server/core/ReactServerAgent/Request.js
```diff
@@ -123,6 +124,15 @@ Request.prototype.toJSON = function(){
 	};
 }
 
+// Enable CORS
+Request.prototype.withCredentials = function (on) {
+	if (on === undefined) {
```
decoy31: Change this to: `if (typeof on === 'undefined')`.

decoy31 (changes requested):

packages/react-server/core/ReactServerAgent/Request.js
```diff
@@ -123,6 +124,15 @@ Request.prototype.toJSON = function(){
...
+		on = true;
+	}
+	this._withCredentials = on;
```
decoy31: If `this._withCredentials` is expected to be a boolean, I'd recommend changing this to: `this._withCredentials = !!on;`

packages/react-server/core/ReactServerAgent/Request.js
```diff
@@ -22,7 +22,8 @@ function Request(method, urlPath, cache) {
 	this._headers = {};
 	this._timeout = null;
 	this._type = "json"; // superagent's default, only for POST/PUT/PATCH methods
-
+	// CORS off by default
+	this._withCredentials = false;
```
decoy31: To me, the property name `_withCredentials` is vague. One of/variant of the following would be more self-documenting:
- `_enableCORS`
- `_isCORSEnabled`
- `_corsEnabled`
- `_withCORSEnabled`
- `_withCrossOriginHTTPRequestsEnabled`

mlolson: I think it's best to stick with  the same syntax that [superagent uses](https://github.com/visionmedia/superagent/blob/f5e39b7746268901c591dde08cd68a5e3944ced1/lib/request-base.js#L396)

decoy31: Oh right. I didn't notice we were modeling this after superagent. That's unfortunate.

Issue #936, opened 2017-06-18
"npm test" fails in two places


mlolson:
I'm seeing tests fail in two places.  

1. The react-server-cli stdout tests in `commands.js` fail because React v15.6.1 throws a deprecation warning regarding proptypes. See: https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

2. react-server-integration tests in InternalServerErrorSpec fail because the tests expect a text response: '[object Object]', but the end point actually returns html that looks like this:

```
<!DOCTYPE html>
  <html lang="en">
    <head>
    <meta charset="utf-8">
    <title>Error</title>
  </head>
  <body>
    <pre>[object Object]</pre>
  </body>
</html>
```

Repro:

1. Clone react-server, 

2. Run  `npm install`

3. `npm test`

gigabo:
@mlolson Was this fixed by #937?

mlolson:
Yes I believe so.

PR #937, opened 2017-06-18
Fix unit and integration tests

mlolson:
* The `react-server-cli` tests in commands.js fail with the lastest version of React, due to new [deprecation warnings emitted by the latest version of React](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html)

    * Solution: Ignore warnings in stdout.

* react-server-integration tests in InternalServerErrorSpec fail because the tests expect a text response: '[object Object]', but the end point actually returns html. 

    * Solution: Update tests to expect an html response.

* Ava was automatically running `packages/generator-react-server/generators/app/templates/test.js`. This file should be run explicitly by `packages/generator-react-server/test/app.js`, otherwise it fails.

    * Solution: add packages/generator-react-server/generators/app/templates/test.js to list of ignored files in ava config so that it is not run automatically.

* Ava overwrites `require.cache` with an object lacking a prototype. This causes the server hot reload cache to throw an error when it attempts to access `require.cache.hasOwnProperty`

    * Solution: Check for `require.cache.hasOwnProperty` before enabling hot reloading cache.

https://github.com/redfin/react-server/issues/936

### DIFF
--- a/packages/generator-react-server/generators/app/templates/test.js
+++ b/packages/generator-react-server/generators/app/templates/test.js
@@ -7,7 +7,7 @@ let rs;
 test.before('start the server', async () => {
 	rs = cp.spawn('npm', ['start']);
 	rs.stderr.on('data', data => console.error(data.toString()));
-	await sleep(10000);
+	await sleep(15000);
 });
 
 test('server is running', async t => {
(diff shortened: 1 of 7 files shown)


gigabo:
Wow, this is awesome.  Thanks for getting this stuff all fixed up @mlolson!

mlolson:
@gigabo For sure! I think you need to hit merge FYI. I don't have authorization

Issue #938, opened 2017-06-20
Client side getQuery() includes fragment identifers in parsed query parameters


edsonlo:
Let's say you submit a url like `http://www.domain.com/page?param1=xyz&param2=123#anchor`. Calling `getQuery()` results in parameters `{param1: "xyz", param2: "123#anchor"}` when `{param1: "xyz", param2: 123}` is expected.

gigabo:
Nice find @edsonlo!

Looks like this is browser-side only.  Likely here (in `ClientRequest.js`):

```js
    getQuery() {
        var indexOfQuestion = this._url.indexOf("?");
        if (-1 === indexOfQuestion) {
            return {};
        }
        return decode(this._url.substring(indexOfQuestion + 1));
    }
```

PR #939, opened 2017-06-20
Exclude hash fragments when extracting query string

mlolson:
https://github.com/redfin/react-server/issues/938


gigabo:
Awesome!  Thanks @mlolson!

Issue #853, opened 2017-01-27
Clarify 500 management


doug-wade:
Currently, when there is an error in the server, we render a page containing `[object Object]`, which is unhelpful.  Maybe in dev, render the stack trace, and allow for providing a 500 page in prod?

PR #940, opened 2017-06-21
Pretty print errors for the client in dev environments

mlolson:
* Add `HttpError` class that extends the native error. This will handle 4xx, 3xx, etc.

* Promisify the navigator sequence, so that errors are passed down through the promise chain and handled in one place, rather than being handled haphazardly or swallowed.

https://github.com/redfin/react-server/issues/853

The react-server middleware is last in the middleware stack, so that a user can insert their own error handling middleware ahead of the default handler.

Example error messages:
error_404

internal_server_error

In production:
error_404_prod

### DIFF
--- a/packages/react-server-cli/src/commands/start.js
+++ b/packages/react-server-cli/src/commands/start.js
@@ -69,6 +69,7 @@ const startHtmlServer = (serverRoutes, port, bindIp, httpsOptions, customMiddlew
 		server.use(bodyParser.json());
 		server.use(helmet());
 		rsMiddleware();
+		server.use(reactServer.errorHandlingMiddleWare);
 	};
 
 	return {
@@ -88,10 +89,9 @@ const startHtmlServer = (serverRoutes, port, bindIp, httpsOptions, customMiddlew
 				// sets the namespace that data will be exposed into client-side
 				// TODO: express-state doesn't do much for us until we're using a templating library
 				server.set('state namespace', '__reactServerState');
+				server.set("reactServerRoutes", require(serverRoutes));
+				server.use(reactServer.renderMiddleware);
 
-				server.use((req, res, next) => {
-					reactServer.middleware(req, res, next, require(serverRoutes));
-				});
 			};
 
 			if (customMiddlewarePath) {
--- a/packages/react-server-cli/src/commands/start.js
+++ b/packages/react-server-cli/src/commands/start.js
@@ -69,7 +69,7 @@ const startHtmlServer = (serverRoutes, port, bindIp, httpsOptions, customMiddlew
 		server.use(bodyParser.json());
 		server.use(helmet());
 		rsMiddleware();
-		server.use(reactServer.errorHandlingMiddleWare);
+		server.use(reactServer.errorHandlingMiddleware);
 	};
 
 	return {
--- a/packages/react-server/core/context/Navigator.js
+++ b/packages/react-server/core/context/Navigator.js
@@ -1,4 +1,3 @@
-
 var EventEmitter = require('events').EventEmitter,
 	logger = require('../logging').getLogger(__LOGGER__),
 	Router = require('routr'),
@@ -7,6 +6,7 @@ var EventEmitter = require('events').EventEmitter,
 	ReactServerAgent = require("../ReactServerAgent"),
 	PageUtil = require("../util/PageUtil"),
 	DebugUtil = require("../util/DebugUtil"),
+	HttpError = require('../errors/HttpError'),
 	{setResponseLoggerPage} = SERVER_SIDE ? require('../logging/response') : { setResponseLoggerPage: () => {} };
 
 var _ = {
@@ -52,7 +52,8 @@ class Navigator extends EventEmitter {
 		if (route) {
 			logger.debug(`Mapped ${request.getUrl()} to route ${route.name}`);
 		} else {
-			this.emit('navigateDone', { status: 404, message: "No Route!" }, null, request.getUrl(), type);
+			let url = request.getUrl();
+			this.emit('navigateDone', new HttpError("Route not found " + url, {code: 404, url: url}), null, request.getUrl(), type);
 			return;
 		}
 
@@ -203,7 +204,11 @@ class Navigator extends EventEmitter {
 			// TODO: I think that 3xx/4xx/5xx shouldn't be considered "errors" in navigateDone, but that's
 			// how the code is structured right now, and I'm changing too many things at once at the moment. -sra.
 			if (handleRouteResult.code && ((handleRouteResult.code / 100)|0) !== 2) {
-				this.emit("navigateDone", {status: handleRouteResult.code, redirectUrl: handleRouteResult.location}, page, request.getUrl(), type);
+				let redirectError = new HttpError("Redirect");
+				redirectError.code = handleRouteResult.code;
+				redirectError.redirectUrl = handleRouteResult.location;
+
+				this.emit("navigateDone", redirectError, page, request.getUrl(), type);
 				return;
 			}
 			if (handleRouteResult.page) {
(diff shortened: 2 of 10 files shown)


packages/react-server/core/renderMiddleware.js
```diff
@@ -67,85 +69,43 @@ module.exports = function(req, res, next, routes) {
...
-			}
-
-			// Success.
-			navigateDfd.resolve();
-
-
-			if (err) {
-				// The page can elect to proceed to render
-				// even with a non-2xx response.  If it
-				// _doesn't_ do so then we're done.
```
mlolson: Is this functionality still needed?

packages/react-server-cli/src/commands/start.js
```diff
@@ -88,10 +89,9 @@ const startHtmlServer = (serverRoutes, port, bindIp, httpsOptions, customMiddlew
 				// sets the namespace that data will be exposed into client-side
 				// TODO: express-state doesn't do much for us until we're using a templating library
 				server.set('state namespace', '__reactServerState');
+				server.set("reactServerRoutes", require(serverRoutes));
+				server.use(reactServer.renderMiddleware);
```
gigabo: Whoa.  This is way nicer!

packages/react-server/core/context/RequestContext.js
```diff
@@ -51,6 +52,26 @@ class RequestContext {
 		this.navigator.on('navigateDone', callback);
 	}
 
+	onNavigatePromise () {
+		let defer = Q.defer();
+		this.navigator.on('navigateDone', (err, page) => {
```
gigabo: This should be a `once`, right?

packages/react-server/core/renderMiddleware.js
```diff
@@ -67,85 +69,41 @@ module.exports = function(req, res, next, routes) {
...
+		const timeout = setTimeout(() => {
+			let metaData = {
+				page: context.page,
 				path: req.path,
-			});
-			handleResponseComplete(req, res, context, start, context.page);
-			next({status: 500});
+			};
+			timeoutDefer.reject(new HttpError(`Navigation timed out after ${FAILSAFE_ROUTER_TIMEOUT}ms`, metaData));
+		}, FAILSAFE_ROUTER_TIMEOUT);
```
gigabo: Should `unref()` this timeout.

packages/react-server/core/renderMiddleware.js
```diff
@@ -213,69 +177,47 @@ function renderPage(req, res, context, start, page) {
...
-	});
-
-	// TODO: we probably want a "we're not waiting any longer for this"
-	// timeout as well, and cancel the waiting deferreds
+		}), Q())
 }
 
 function rawResponseLifecycle () {
 	return [
-		Q(), // NOOP lead-in to prime the reduction
```
gigabo: Nice, been meaning to get rid of these forever. :)

packages/react-server/core/renderMiddleware.js
```diff
@@ -67,85 +69,41 @@ module.exports = function(req, res, next, routes) {
...
+			context.onNavigatePromise(),
+			timeoutDefer.promise,
+		])
+		.tap(() => clearTimeout(timeout)) //Clear timeout if promise succeeds - not strictly neccessary
+		.then((page) => renderPage(req, res, context, page))
+		.then(() => endResponse(req, res))
+		.catch(next)
+		.finally(() => {
+			clearTimeout(timeout);
+			handleResponseComplete(req, res, context);
```
gigabo: This looks like it switches the order of execution between `endResponse` and `handleResponseComplete`.

Trying to think if that could cause any problems... 🤔

packages/react-server/core/renderMiddleware.js
```diff
@@ -1042,7 +984,7 @@ function logRequestStats(req, res, context, start){
 	logger.gauge("bytesRead", stash.bytesR, {hi: 1<<12});
 	logger.gauge("bytesWritten", stash.bytesW, {hi: 1<<18});
 
-	var time = new Date - start;
+	var time = new Date - RLS().startTime;
```
gigabo: Thank you.  These little bits of cleanup are great.

packages/react-server/core/errorHandlingMiddleware.js
```diff
@@ -0,0 +1,53 @@
...
+
+	//Some rough bounds checking here
+	if (typeof code !== "number" || code <= 0 || code >= 1000) {
+		code = 500;
+	}
+
+	res.status(code);
+
+	if (process.env.NODE_ENV === "production") { // eslint-disable-line no-process-env
+		res.send(`<pre>${err.message || err}</pre>`); //Just send error message in production
```
gigabo: This looks like an XSS vector.

Need to escape `err.message`, I think, to avoid malicious values in the requested URL for 404s (for example).

packages/react-server/core/context/Navigator.js
```diff
@@ -215,11 +220,7 @@ class Navigator extends EventEmitter {
 			}
 
 			this.emit('navigateDone', null, page, request.getUrl(), type);
-		}).catch(err => {
-			logger.error("Error while handling route", err);
-
-			this.emit('navigateDone', {status: 500}, page, request.getUrl(), type);
-		});
+		}).catch(err => this.emit('navigateDone', err, page, request.getUrl(), type));
```
gigabo: Does `err` need to be converted to an `HttpError` here?  With code 500?  Or is that already done and I'm just missing it?

packages/react-server/core/context/Navigator.js
```diff
@@ -203,7 +204,11 @@ class Navigator extends EventEmitter {
 			// TODO: I think that 3xx/4xx/5xx shouldn't be considered "errors" in navigateDone, but that's
 			// how the code is structured right now, and I'm changing too many things at once at the moment. -sra.
 			if (handleRouteResult.code && ((handleRouteResult.code / 100)|0) !== 2) {
-				this.emit("navigateDone", {status: handleRouteResult.code, redirectUrl: handleRouteResult.location}, page, request.getUrl(), type);
+
+				let error = new HttpError("Page returned code " + handleRouteResult.code, {code: handleRouteResult.code});
+				error.redirectUrl = handleRouteResult.location;
+				page.setRedirectUrl(handleRouteResult.location);
+				this.emit("navigateDone", error, page, request.getUrl(), type);
```
gigabo: The client controller will also need to be updated to handle this new interface, right?

gigabo:
@mlolson This is awesome!  There's a lot in here.  Way more than just a stack trace in dev. :)

I left some comments and questions.  Thanks for working on this!

packages/react-server/core/context/RequestContext.js
mlolson: Yeah, good call

packages/react-server/core/errorHandlingMiddleware.js
mlolson: Good call!

packages/react-server/core/context/Navigator.js
mlolson: That's a good question- I think one could make the argument either way. Let's discuss further this afternoon.

mlolson:
@gigabo Yeah no problem! Hope it's useful. It was mostly a learning exercise for my own benefit, so no worries if you want to push back on some parts of it. I can break it out into smaller PR's if that's more convenient.

packages/react-server/core/renderMiddleware.js
mlolson: Ah yeah, I can switch those back if need be.

packages/react-server/core/renderMiddleware.js
mlolson: Oh good. Wasn't sure if that was going to work

packages/react-server/core/context/Navigator.js
mlolson: `HttpError` should have members `status` and `redirectUrl` same as the object that was being returned before, but this should be tested further.

PR #944, opened 2017-07-21
Add english as the default document language

monicatie:
This helps screen readers use the correct pronunciation rules and to help automatic translation. I'd like to add support for specifying the document language so it isn't just english, but starting with English so there's something.


gigabo:
Good idea. Thanks @monicatie. Please sign the CLA when you get a chance so we can merge.

monicatie:
Thanks @gigabo , CLA signed!

PR #946, opened 2017-08-07
	fix no onclick bug on Link Component

hbucius:
fix no onclick bug on Link Component
--


hbucius:
who can help me review it ?

hbucius:
Currently, Link do not have any callbacks. It is very inconvenient. For <a>, besides hrefs, It can also add callbacks. We need to add this for Link.

gigabo:
Nice fix. Thanks @hbucius!

Issue #942, opened 2017-06-30
TypeError: require.cache.hasOwnProperty is not a function


EthanOrange:
platform: win10 *64
node: v 8.1.2
npm : v 5.0.4

EthanOrange:
npm run start error

rutger:
+1

AlecBoutin:
Seems to be a problem with Node 8.x. The require.cache object no longer has a hasOwnProperty method. Change is reproducible by openning node REPL and running `require.cache.hasOwnProperty`.

Full stack trace
```C:\Users\Alec\Code\react-server-test\node_modules\react-server-cli\target\serverSideHotModuleReload.js:37
                if (/node_modules/.test(filename) === false && require.cache.hasOwnProperty(filename)) {
                                                                             ^

TypeError: require.cache.hasOwnProperty is not a function
    at C:\Users\Alec\Code\react-server-test\node_modules\react-server-cli\target\serverSideHotModuleReload.js:37:64
    at Array.map (native)
    at serverSideHotModuleReload (C:\Users\Alec\Code\react-server-test\node_modules\react-server-cli\target\serverSideHotModuleReload.js:36:37)
    at Compiler.<anonymous> (C:\Users\Alec\Code\react-server-test\node_modules\react-server-cli\target\compileClient.js:213:45)
    at Compiler.applyPlugins (C:\Users\Alec\Code\react-server-test\node_modules\tapable\lib\Tapable.js:26:37)
    at Watching._done (C:\Users\Alec\Code\react-server-test\node_modules\webpack\lib\Compiler.js:78:17)
    at Watching.<anonymous> (C:\Users\Alec\Code\react-server-test\node_modules\webpack\lib\Compiler.js:61:18)
    at Compiler.emitRecords (C:\Users\Alec\Code\react-server-test\node_modules\webpack\lib\Compiler.js:282:37)
    at Watching.<anonymous> (C:\Users\Alec\Code\react-server-test\node_modules\webpack\lib\Compiler.js:58:19)
    at C:\Users\Alec\Code\react-server-test\node_modules\webpack\lib\Compiler.js:275:11```

gestone:
Running into the same issue. Any fixes besides downgrading the version of node?

hugotox:
Same issue. Mac 10.12.6, node 8.3.0

dfadev:
See PR #947

jdt1204:
[...]

PR #947, opened 2017-08-15
fix for #942, require.cache.hasOwnProperty call for Node 8

dfadev:


gigabo:
Awesome! Thanks @dfadev!

Issue #883, opened 2017-02-22
Consider getting rid of mobile-detect as a dependency on the client.


aickin:
I was curious how big the basic bundle is, so I ran an webpack-bundle-size-analyzer on `Homepage.js` from the basic react-server `yo` scaffold. I posted the results [here](http://react-server-analysis.surge.sh). The file sizes shown are minified, *un*gzipped.

There are a few interesting things, but the thing that most sticks out to me is the amount of space `mobile-detect` takes: 36.2kB out of 307kB, or 11.8%. 

As far as I can tell, mobile-detect is only used to see if the user is mobile/phone/tablet. If that's right, it'd probably be better to figure out a way to run `mobile-detect` on the server and then just send down the string "mobile"/"phone"/"tablet", while excluding `mobile-detect` from the client build. Thoughts?

gigabo:
> run `mobile-detect` on the server and then just send down the string "mobile"/"phone"/"tablet",

@aickin Ooh!  Good idea!  I think we're currently exposing a full `mobile-detect` instance, so that would probably be a breaking change.  Or maybe... we could send down a thin dummy that implements the interface with values hard-coded by the server? 🤔

ptomasroos:
I think this is a great idea. Every page starts with a response form the server, moving as many bits as possible and letting the output be a bool just nails it.

gigabo:
This is addressed by #950 which landed in `react-server@0.7.0`.

PR #950, opened 2017-09-08
Simplify device type routing

gigabo:
The mobile-detect package is large. We shouldn't send it to the browser.

This change uses mobile-detect on the server only. It associates a device type with the request and sends only that pre-determined value to the browser.

This eliminates the "mobile" route target, which was previously the union of "phone" and "tablet". These two must be specified individually going forward (though they may route to the same page class). It also adds a "desktop" route target.

The `getMobileDetect()` method on the request context is eliminated, and replaced with a `getDeviceType()` method that returns a string.

This is a breaking change.

### DIFF
--- a/packages/react-server/core/context/Navigator.js
+++ b/packages/react-server/core/context/Navigator.js
@@ -83,45 +83,20 @@ class Navigator extends EventEmitter {
 
 			var loaders = route.config.page;
 
-			// Normalize.
-			// The page object may either directly be a loader or
-			// it may be an object whose values are loaders.
-			if (_.isFunction(loaders)){
-				loaders = {'default': loaders};
-			}
-
-			var mobileDetect = this.context.getMobileDetect();
-
 			// Our route may have multiple page implementations if
 			// there are device-specific variations.
 			//
 			// We'll take one of those if the request device
 			// matches, otherwise we'll use the default.
 			//
-			// Note that 'mobile' is the _union_ of 'phone' and
-			// 'tablet'.  If you _really_ want an iPad and an
-			// iPhone to get the _same_ non-desktop experience,
-			// use that.
-			//
-			var loadPage = [
-				'phone',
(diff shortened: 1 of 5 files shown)


gigabo:
This addresses #883.

packages/react-server/core/context/Navigator.js
```diff
@@ -83,45 +79,20 @@ class Navigator extends EventEmitter {
...
-				'phone',
-				'tablet',
-				'mobile',
-			].reduce((loader, format) => {
-
-				// We'll take the _first_ format that matches.
-				if (!loader && loaders[format] && mobileDetect[format]()){
-
-					// Need to disambiguate for bundleNameUtil.
-					route.name += '-'+format;
```
roblg: Do we not need this route.name override thing anymore for some reason?

roblg: I guess maybe it's a product of eliminating `mobile`?

gigabo: Oh, nice catch. We use it for stat partitioning. I'll reintroduce that.

packages/react-server/core/renderMiddleware.js
```diff
@@ -1070,6 +1071,17 @@ function getNonInternalConfigs() {
 	return nonInternal;
 }
 
+function getDeviceType(req) {
+	var md = new MobileDetect(req.get('user-agent'));
+	var types = [ 'phone', 'tablet', 'mobile' ];
```
roblg: This still includes `mobile`, but the PR comments say we're eliminating that. Can it be removed here?

gigabo: Thank goodness you're reviewing this carefully. 😳

roblg: For once ;)

PR #954, opened 2017-09-24
Update HomeWhySection.md

JennHaggerty:
Update stale documentation to point to correct link.


gigabo:
Thanks @JennHaggerty!

gigabo:
Oh, whoops, didn't notice that this was pointing at `gh-pages` instead of `master`.

We actually don't serve react-server.io on github pages, since the site itself is run on React Server. :)

PR #955, opened 2017-09-29
Classify "unknown" mobile devices as phones

gigabo:
This deals with an edge case of mobile-detect. See commentary.


gigabo:
Gonna cowboy this. https://github.com/redfin/react-server/pull/956 is taking too long. :hurtrealbad:

PR #956, opened 2017-09-29
Take a carrot-range dep on compression in cli

gigabo:
This is an attempt to deal with an nsp failure that's breaking travis.


drewpc:
Unfortunately, this PR is so old that NSP is no longer in service.  Need to convert things to NPM audit (see #998).

PR #959, opened 2017-10-10
Added nsp exceptions to generator-react-server

davidk107:
This PR adds some exceptions to `nsp` by specifying a `.nsprc` file described under https://github.com/nodesecurity/nsp#exceptions. This means the our build won't fail for the two nsp errors that we are currently seeing. 

The suppressed errors are https://nodesecurity.io/advisories/479 which seems to be a result of us using `superagent` 1.8.4. That in turn causes https://nodesecurity.io/advisories/535 since it imports `mime` 1.3.4.


gigabo:
Awesome @davidk107! Thank you!

PR #962, opened 2017-10-11
Client transitions still need to render in order

### LINKED ISSUE #961: Navigation playground is broken
Looks like I broke it in https://github.com/redfin/react-server/pull/948.

This likely means client transitions are totally borked right now.

gigabo:
During client transitions the root structure is built up using a state machine that requires an in-order render. First render can still go out of order because the server has already laid out the root structure.

This fixes #961.

### DIFF
--- a/packages/react-server/core/ClientController.js
+++ b/packages/react-server/core/ClientController.js
@@ -601,26 +601,35 @@ class ClientController extends EventEmitter {
 			}
 		};
 
-		// As elements become ready, prime them to render as soon as
-		// their mount point is available.
-		//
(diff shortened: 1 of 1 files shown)


packages/react-server/core/ClientController.js
```diff
@@ -601,26 +601,35 @@ class ClientController extends EventEmitter {
 			}
```
gigabo: May be easier to look at with [whitespace changes hidden](https://github.com/redfin/react-server/pull/962/files?w=1).

Issue #884, opened 2017-02-23
RootElement not updated with new props on listen


salehhamadeh:
I am trying to re-render a page after a change is emitted to `listen`. `RootElement` is getting the changes from the emitter, and the callback in `componentDidMount` is called with the correct `childProps`. `this.props` is set to use the new `childProps` correctly when we set it using `_.assign`. However, once we call `this.forceUpdate()`, the props are reset to what they were before the change. Since this happens with every update, the page is not responsive on the client-side. Here is the code for reference.
```
this.props.subscribe(childProps => {
        ...
        // Okay, now we've complained about it
        // sufficiently, let's go ahead and update.
        this.props = _.assign({}, this.props, {childProps});
        this.forceUpdate();
});
```
I am using React 15.4.1

gigabo:
Thanks for reporting @salehhamadeh.

I wonder if this broke with the React 15 upgrade... 🤔 

Any idea what we could do here to get it working?

salehhamadeh:
Can we use `setState` instead of updating the props and forcing an update?

gigabo:
React Server RootElements set props on first render.  Seems weird to change that to state later?

PR #963, opened 2017-10-13
Add code splitting support to RootElement

davidk107:
This introduces a `componentLoader` and `childProps` prop to `RootElement`. `componentLoader` takes in an asynchronous function that resolves to a component. `RootElement` will pause on rendering until `componentLoader` has been resolved. Using this with dynamic code loaders such as `require.ensure` or `import()` can result in having server side rendered sections without having to download the client side Javascript until later in the page loading process. 

`childProps` will be additional props that can be passed to the resolved component. `childProps` are mainly used for sending down synchronous props. It's usage is not restricted to `componentLoader` and can be used with out it. 

Examples and more detailed descriptions can be found under https://github.com/davidk107/react-server/tree/add-componentLoader-support/packages/react-server-examples/code-splitting


packages/react-server/core/components/RootElement.js
```diff
@@ -29,8 +36,10 @@ class RootElement extends React.Component {
 
 				// Okay, now we've complained about it
 				// sufficiently, let's go ahead and update.
-				this.props = _.assign({}, this.props, {childProps});
```
davidk107: Took a look into React internals and it appears that setting `this.props = ...` has no effect whatsoever. Thus the subsequent call to `this.forceUpdate()` re-renders the component without any `prop` updates. By using `setState` this will also fix the issue described in https://github.com/redfin/react-server/issues/884

packages/react-server/core/components/RootElement.js
```diff
@@ -157,26 +166,36 @@ RootElement.installListener = function(element, listen) {
...
+		.then(results => {
+			var [whenResult, loadedComponent] = results;
+			if (whenResult.value || loadedComponent.value || childProps) {
+				// merge in child props from listen, when, and childProps
+				const clonedChildProps = _.assign({}, rendered.props.childProps, whenResult.value, childProps);
+
+				// if we have a component loader specified, copy the resolved component
+				// and render that with the current child as a child of that component
+				const currentChild = rendered.props.children;
+				const childToRender = componentLoaderDeferred ? React.createElement(loadedComponent.value, null, currentChild) : currentChild;
```
gigabo: It looks like this will always create a new element from the `componentLoaderDeferred` if there was one.

Should this only use `loadedComponent.value` on the first pass?

packages/react-server/core/components/RootElement.js
```diff
@@ -157,26 +166,36 @@ RootElement.installListener = function(element, listen) {
...
-			// to prevent cloneElement shallow merge from clobbering "listen" childProps
-			?React.cloneElement(rendered, {childProps: _.assign({}, rendered.props.childProps || {}, childProps)})
-			:rendered
-		)
+		.then(() => Q.allSettled([when, componentLoaderDeferred]))
+		.then(results => {
+			var [whenResult, loadedComponent] = results;
+			if (whenResult.value || loadedComponent.value || childProps) {
+				// merge in child props from listen, when, and childProps
+				const clonedChildProps = _.assign({}, rendered.props.childProps, whenResult.value, childProps);
```
gigabo: Does this need to also include `rendered.state.childProps` now?

davidk107: I tested it out and it seems that at this point the `constructor` for `RootElement` has not been called (which is to be expected) Thus `rendered.state.childProps` will be `undefined`. However even if the constructor is called, both `rendered.state.childProps` and `rendered.props.childProps` will be the same object and so we are fine with just relying on using `rendered.props.childProps` here. 

I can say that I am confident that `rendered`'s `componentDidMount` will not have been called. That is what hooks up the listener to `listen` which might cause changes to `rendered.state.childProps`.

packages/react-server/core/components/RootElement.js
davidk107: Not sure I understand. What do we mean by on the first pass?

packages/react-server-examples/code-splitting/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "private": true,
+  "version": "0.0.1",
+  "description": "",
+  "main": "HelloWorld.js",
+  "scripts": {
+    "start": "IS_SERVER=true react-server start",
+    "clean": "rm -rf build __clientTemp",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Sasha Aickin",
```
gigabo: This should be you. :)

packages/react-server/core/components/RootElement.js
```diff
@@ -39,6 +48,16 @@ class RootElement extends React.Component {
 		if (this.props.unsubscribe) this.props.unsubscribe();
 	}
 
+	componentWillReceiveProps(nextProps) {
+		// Incase we receive new props such as during client transitions
+		// We will want to update our state's childProp with any new childProps
+		// that may have been passed in. This still respects props as the ultimate source of truth
+		const newChildProps = _.assign({}, this.state.childProps, nextProps.childProps);
```
davidk107: One downside to doing this is that if `nextProps.childProps` is empty, then `newChildProps` will still retain old keys/values from `this.state.childProps`. The workaround to this would be to pass in `null` as the new value for those keys if that behavior is desired.

PR #965, opened 2017-10-25
Upgrade React to v16.0.0

salehhamadeh:
This PR only upgrades React to v16.0.0. It does not use any new features, aside from changing `React.render` to `React.hydrate` in `ClientController`. I thought about using `React.renderToNodeStream` instead of `React.renderToString` in `renderMiddleware.getElements`, but I was not sure what is the best way to do it. I was thinking of having a stream for every `RootElement` that are all read and poured into a higher-level stream. I do not know if this will work. If you have ideas let me know, and I can either include it in this PR or in a separate one.

I used react-codemod (https://github.com/reactjs/react-codemod) to refactor PropTypes and React.createClass. I tested all react-server-test-pages. The only test page that behaves differently is the Error Logging Tests Page. Instead of the buttons logging errors and staying on the page, they are completely removed. This is the expected behavior since the introduction of error boundaries in React 16 https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html


packages/react-server-cli/package.json
```diff
@@ -43,7 +43,7 @@
     "pem": "^1.8.1",
     "q": "1.4.1",
     "raw-loader": "^0.5.1",
-    "react-hot-loader": "~1.3.0",
+    "react-hot-loader": "^3.1.1",
```
gigabo: Is this version compatible with the older versions of React that are supported below?

packages/react-server/core/components/FragmentDataCache.jsx
```diff
@@ -1,6 +1,8 @@
 
 var ReactServerAgent = require("../ReactServerAgent"),
-	React = require("react");
+	React = require('react'),
+	PropTypes = require('prop-types'),
+	createReactClass = require("create-react-class");
```
gigabo: Could we avoid this new dependency by making `FragmentDataCache` an es6 class?

salehhamadeh: Sure I'll do that.

salehhamadeh:
Bo and I talked offline. We decided to update react-hot-loader to v3.1.1. Server-side hot reloading will continue to work, so making a change and refreshing the page will show the changes. However, client-side hot reloading is broken because React 16 does not have ReactMount, which react-hot-loader v1 depended on. I will open an issue for the broken client-side hot module replacement.

packages/react-server-cli/package.json
salehhamadeh: I added a comment below. Yes it is compatible, but we will lose client-side HMR because react-hot-loader now needs us to use `module.hot.accept` and rendering the component tree in our entrypoints rather than using the no longer supported `ReactMount`.
See step 2 here http://gaearon.github.io/react-hot-loader/getstarted/

packages/react-server-redux/package.json
```diff
@@ -26,7 +26,7 @@
     "redux": "^3.6.0"
   },
   "peerDependencies": {
-    "react": "~0.14.2 || ^15.1.0"
+    "react": "~0.14.9 || ^15.3.0 || ^16.0.0"
```
gigabo: Why are the older version numbers changing here?

Lerna will only hoist if version strings match exactly.

salehhamadeh: `prop-types` is only compatible with these versions. In previous versions, React will log console warnings if we use `prop-types` with previous versions https://www.npmjs.com/package/prop-types

PR #972, opened 2017-11-30
Make ClientRequest chop off fragment identifiers from urls

j-beatzz:
This PR tries to make `ClientRequest` and `ExpressServerRequest` consistent by ensuring the that fragment identifiers are chopped of from urls in `ClientRequest` and that they are not included when `getUrl` is called.

### DIFF
--- a/packages/react-server/core/ClientRequest.js
+++ b/packages/react-server/core/ClientRequest.js
@@ -10,10 +10,18 @@ class ClientRequest {
 		bundleData,
 		reuseDom,
 	}={}) {
-		this._url = url;
 		this._opts = {
 			bundleData,
 			reuseDom,
+		}
+
+		// Chop off the fragment identifier from the url i.e everything from the # to the end of the url
+		// if it exists to make this consistent with ExpressServerRequest.
+		var match = url.match(/([^#]*)/);
+		if (match === null || !match[1]) {
+			this._url = url
+		} else {
+			this._url = match[1];
 		}
 	}
 
@@ -38,8 +46,8 @@ class ClientRequest {
 	}
 
 	getQuery() {
-		//Grab fragment between first "?" and first "#" or end of string
-		var match = this._url.match(/\?([^#]*)/);
+		// Grab fragment after first "?"
+		var match = this._url.match(/\?(.*)/);
 
 		if (match === null || !match[1]) {
 			return {};
--- a/packages/react-server/core/__tests__/ClientRequestSpec.js
+++ b/packages/react-server/core/__tests__/ClientRequestSpec.js
@@ -7,6 +7,26 @@ describe("ClientRequest", () => {
 		clientRequest = new ClientRequest("/");
 	});
 
+	it("removes fragment identifiers from the url", (done) => {
+		clientRequest = new ClientRequest("/");
+		expect(clientRequest.getUrl()).toEqual("/");
+		clientRequest = new ClientRequest("/react-server/foo#bar");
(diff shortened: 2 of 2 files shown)


packages/react-server/core/ClientRequest.js
```diff
@@ -15,6 +15,15 @@ class ClientRequest {
...
 		}
+
+		// Chop off the fragment identifier from the url i.e everything from the # to the end of the url
+		// if it exists to make this consistent with ExpressServerRequest.
+		var match = url.match(/([^#]*)/);
+		if (match === null || !match[1]) {
+			console.error("An Invalid url was provided");
+		} else {
+			this._url = match[1];
+		}
```
roblg: Couple requests for this section:

* I think for the purposes of this PR, I'd prefer not to log anything in the case an "invalid" URL was provided. We could/should validate the URL that we're given, but if we were doing that I'd probably want to `throw` rather than just log, and that's considerably more involved than stripping the hash off the end, so we can put that aside for now.
* I'd like to keep the two assignments to `this._url` in constructor closer together. Feel free to move the first one down into your first if-case.


This doesn't apply to this specific PR, since I'd like to remove the new logging anyway, but for future react-server submissions, we've got a logger set up in react-server that ensures that we get log messages to the appropriate places w/ the appropriate severity levels. In general, rather than `console.error`, we should use that setup (if you search for "logger" in the code, it should be easy to find examples).

packages/react-server/core/__tests__/ClientRequestSpec.js
```diff
@@ -7,6 +7,24 @@ describe("ClientRequest", () => {
...
+		clientRequest = new ClientRequest("/react-server/foo/?#bar");
+		expect(clientRequest.getUrl()).toEqual("/react-server/foo/?");
+		clientRequest = new ClientRequest("/react-server/foo/?foo=bar&baz=123");
+		expect(clientRequest.getUrl()).toEqual("/react-server/foo/?foo=bar&baz=123");
+		clientRequest = new ClientRequest("/react-server/foo/?foo=bar&baz=123#");
+		expect(clientRequest.getUrl()).toEqual("/react-server/foo/?foo=bar&baz=123");
+		clientRequest = new ClientRequest("/react-server/foo/?foo=bar&baz=123&zed=abc?#some-fragment?#");
+		expect(clientRequest.getUrl()).toEqual("/react-server/foo/?foo=bar&baz=123&zed=abc?");
+		done();
+	});
```
roblg: Thanks for adding tests!

One additional test I'd like to see here is a test that has a URL with a `?` following the `#`. e.g. `/react-server/foo#?bar=3&foo=7`. I feel like I've seen libraries that store data in the hash in that format, and because we're looking for both `#` and `?` in regexes in `ClientRequest`, it seems like a place we could have regressions in the future.

j-beatzz: In this case, what would the expected behavior be? Based on this:

https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

It seems like it will be wrong to store the part after the `?` as as query params in this case since it does not comply with the uri syntax.

roblg: Right, I'd expect that everything after the `#` is stripped -- it's still part of the hash, and not the query string. The concern would be that something in the code changes later, and all of a sudden we start matching the `?` before the `#` or something, and it _accidentally_ gets included as the query string. A test case should help prevent that kind of regression.

gigabo:
This looks good to me with the latest changes. Nice fix @j-beatzz.

PR #988, opened 2018-04-12
Fixed code indentation, code format and modify docs

Syed-Saad-Qamar:


roblg:
Hi @Syed-Saad-Qamar. Thanks for the submission, but unfortunately we can't accept it in its current form. The PR is much too large to review carefully, and at a minimum it should be broken up change type. Even then, the JS changes would likely make a giant PR and should be split.

I skimmed the changes in the first few files, and I have a few concerns with the content of the change as well from what I saw there:

* I don't think I agree with the changes to the documentation. Ending the headings with a `:` looks reasonable in the markdown, but I don't think it makes sense when the docs are deployed to react-server.io
* It looks like the changes were generated by a tool of some sort? In addition to nice-to-have changes (consolidating on one quote style), it also picked up some spacing/alignment issues that were intentional and we would prefer not to change. It also looks like at least some of the changes violate our lint checks, since the build failed.

PR #1001, opened 2019-04-23
Upgrade lerna

drewpc:
Addresses all of the issues in #997 and #934

Issue #999, opened 2019-04-23
Do Lint Better


drewpc:
ESLint should be used in all of the projects, all of the source code, and incorporate React best practices as well.  Streamline the use of the top-level eslint tool (not a mixture of eslint and gulp-eslint) and update the rules + code to make life better.  Blocked by #997 .

Issue #997, opened 2019-04-23
Improve Lerna + NPM Integration


drewpc:
Lerna has been incorporated but most of the packages and documentation need to be updated so it is clear how to build and work with packages within the monorepo.  Additionally, NSP has been deprecated in favor of `npm audit`.  This issue should accomplish the following items:

- [x] Upgrade Lerna to a recent major version
- [x] Upgrade requirements for NPM to be >= 6.4
- [x] Streamline the individual package NPM scripts to support lint, build, clean, and test scripts as appropriate.
- [x] Use Lerna to hoist the devDependencies to the top level of the repo and link individual packages together.
- [x] Address #934 
- [x] Ensure Travis CI is testing with NPM >= 6.4 and for all versions of Node we support (6, 8, 9, 10, and 11).

PR #1002, opened 2019-04-23
Do Lint Better

drewpc:
This addresses all of #999 and incorporates/relies on changes in #997 and #1001 to be approved/integrated.  Things shouldn't be linked together, I know, but it's hard to not do that with the number of changes coming.


drewpc:
Since this is a minor infrastructure change and all tests are passing, I'll merge after one approval.

Issue #1000, opened 2019-04-23
Upgrade to Webpack 4 + Babel 7


drewpc:
It's time, folks.  This project is stuck on Webpack 1 and Babel 6 which limits a lot of what can be done, in addition to just being old.  Webpack 4 and Babel 7 will poise us for the future and allow us to better integrate the Rollup changes proposed in #985 . 

- [x] Upgrade packages to use Webpack 4 and Babel 7
- [x] Ensure sane default values for the react-server-cli generated Webpack configs
- [x] Incorporate the use of `webpack-dev-middleware` instead of `webpack-dev-server` to simplify things (addresses proposal #807 and PR #808 (that was rejected)...gonna have to push this through due to unnecessary complexity here).  This also removes the need for #932 and #774

drewpc:
Addressed in #1003

Issue #237, opened 2016-05-31
Use tree shaking


doug-wade:
This could in theory reduce the size of the initial bundle. Webpack 2 is not out of beta, yet, though, so we'd have to migrate to [rollup](https://github.com/rollup/rollup).

Migrated from #50

doug-wade:
From @kcjonson 

>>> Tree shaking is awesome but good linting goes a long way for blocking unused imports, especially if everything is on ES6 syntax 😉
Seems like we should do this before we look at RootElements having their own CSS. Highly theoretical but we could use the new System.import syntax to write CSS dependencies in as RootElements are encountered, or even resolved.
If we're planning on upgrading to webpack 2, seems like we should do it first since there were "major changes" to the resolver which might effect other work. However, its in beta so are we ok with the following items being held up by it?

doug-wade:
More evidence in favor of rollup instead of webpack https://nolanlawson.com/2016/08/15/the-cost-of-small-modules/

drewpc:
Webpack 2 is out of beta now...maybe worth taking a look at migrating from Webpack 1 to Webpack 2 and using the tree shaking stuff there: https://medium.com/modus-create-front-end-development/webpack-2-tree-shaking-configuration-9f1de90f3233

roblg:
FYI, I started this upgrade for Redfin's code internally, and one issue that I ran into that I haven't figured out yet is that the ExtractTextPlugin is still busted in watch mode, and the little hack we use in `react-server-cli` to "subclass" ETWP and make in non-cacheable didn't seem to work to fix it this time. :(

karlhorky:
webpack 2 has released a final release candidate before shipping the final version 2: https://twitter.com/TheLarkInn/status/819215334590783489

doug-wade:
> Maybe even today 😏

Woah.  Let's land #809 so that @drewpc doesn't have to entirely rewrite that _super awesome_ pr, and then jump on this.

drewpc:
@doug-wade thanks for the shout out!  Unfortunately....that super awesome PR was kinda thrown out.  I'll need to rewrite it anyway.  There's a number of architecture decisions that need to be made first, including when/if to move to Webpack 2.

roblg:
[...]

PR #1003, opened 2019-04-24
Upgrade to Webpack 4 + Babel 7

### LINKED ISSUE #970: Client-side hot module replacement does not work
If you modify a file while react-server is running with `hot: true`, the server updates the file and the client receives the updated file (can be seen in the browser console), but the new component is not used. This broke down when we upgraded react-hot-loader to v3. We had to do this because react-hot-loader v1 depended on `ReactMount`, which is no longer supported in React 16. The new version of react-hot-loader requires us to tell a component how to render itself. This is shown in "Step 2: Using HMR to replace the root component" on react-hot-loader's Getting Started page http://gaearon.github.io/react-hot-loader/getstarted/

davidgolden:
Upgrades webpack to 4.x and babel to 7.x. This addresses #1000, and fixes #970, #807, and #237.


drewpc:
This builds on #1001 and #1002, so once those PRs are merged, this will seem a lot less daunting.  Please approve #1001 and #1002 first.

Note that this includes some breaking changes that we can address in the release documentation/change log.

drewpc:
Merging per conversation with @roblg.

